PR #8, opened 2015-04-14
Acoustics imex

danielru:
Provides an implicit-explicit SDC for the 1D acoustic-advection problem. The explicit advective part is provided by sharpclaw, the implicit acoustic part by a simple finite difference matrix.

While the result appears to be correct, the SDC residual for some reason remains fixed at about 1.0


danielru:
Something is still fishy here... I suggest not to merge right now, but wait until I push corrections into this branch.

danielru:
One-dimensional acoustic-advection with IMEX splitting is now working fine: Can reproduce analytical solution up to high accuracy (tried ~5e-9).

@pancetta Ready to merge!

PR #18, opened 2015-05-07
Example/acoustic 2d implicit

danielru:
Adds another example solving a 2D acoustic problem (so far no advection) which corresponds to the first order system version of the 2D wave equation.

Caveat: Probably because of the treatment of the boundary conditions, residuals are not decaying although error is (in simple quasi-1D test case). The problem is the before solving the linear system in solve_system, the right hand side is modified according to the boundary conditions. This can't be done in fimpl, so there is some inconsistency here that is likely preventing residuals from decreasing.

I am not sure at the moment how this can be remedied...


danielru:
All screwed up - starting over. There is more clever way to incorporate the BC into the FD matrix.

PR #24, opened 2015-07-07
added counter for required GMRES iterations in boussinesq example

danielru:
- adds a logging and callback object definition to the ProblemClass of the boussinesq example
- the callback object is used in SciPy GMRES to track the number of required iterations and the residual
- the logging object tracks the total number of iterations over the whole integration

[Review on Reviewable](https://reviewable.io/reviews/parallel-in-time/pysdc/24)


pancetta:
This will work and is in the spirit of the ProblemClass, but you also could use the stats object as container for the result, see e.g. 
https://github.com/Parallel-in-Time/pySDC/blob/master/examples/penningtrap/HookClass.py#L64 
for an example.

PR #39, opened 2015-10-14
minor tweaks and parameter changes

danielru:
Minor changes and fixed parameters. Now runs PFASST with 16 processors - alas, ultimately the last process needs 16 iterations to converge.

### DIFF
--- a/examples/boussinesq_2d_imex/playground.py
+++ b/examples/boussinesq_2d_imex/playground.py
@@ -24,11 +24,11 @@ if __name__ == "__main__":
     # set global logger (remove this if you do not want the output at all)
     logger = Log.setup_custom_logger('root')
 
-    num_procs = 8
+    num_procs = 16
 
     # This comes as read-in for the level class
     lparams = {}
-    lparams['restol'] = 1E-8
+    lparams['restol'] = 5E-6
     
     swparams = {}
     swparams['collocation_class'] = collclass.CollGaussLobatto
@@ -36,12 +36,12 @@ if __name__ == "__main__":
     swparams['do_LU'] = True
 
     sparams = {}
-    sparams['maxiter'] = 12
+    sparams['maxiter'] = 16
 
     # setup parameters "in time"
     t0     = 0
-    Tend   = 3000
-    Nsteps =  500
+    Tend   = 384
+    Nsteps =  128
     #Tend   = 30
     #Nsteps =  5
     dt = Tend/float(Nsteps)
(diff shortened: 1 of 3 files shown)


examples/boussinesq_2d_imex/playground.py
```diff
@@ -36,12 +36,12 @@
     swparams['do_LU'] = True
 
     sparams = {}
-    sparams['maxiter'] = 12
+    sparams['maxiter'] = 16
```
pancetta: Do you need this fixed at 16? The thing is that for FT we need to keep this free, otherwise we cannot estimate the overhead..

examples/boussinesq_2d_imex/playground.py
```diff
@@ -36,12 +36,12 @@
...
 
     # setup parameters "in time"
     t0     = 0
-    Tend   = 3000
-    Nsteps =  500
+    Tend   = 384
+    Nsteps =  128
```
pancetta: The GS example takes 640 steps.. can you extend this a little bit? We have randomized fault injections, so we need many steps to get good statistics.

danielru: You can just increase Tend and Nsteps by constant factors, so e.g. Tend = 768 and Nsteps = 1000 or so. Only increasing Nsteps would also work, a finer resolution might help PFASST converge better.

examples/boussinesq_2d_imex/playground.py
danielru: No, you can set this to whatever value you want. The only problem is that with too few iterations, PFASST can become unstable over the course of the simulation.

pancetta: OK, so PFASST will reasonably converge with the residual criterion?

danielru: In general, yes. You might have to tweak the tolerance a bit, PFASST takes quite long to reach very fine residuals. I tried to go for something between 1e-5 and 1e-6. The residual after the first iteration is already 1e-4, so larger values do not make much sense.

pancetta: What would be a reasonable tolerance to get approx. 10-20 iterations per block, then? Does this change during the simulation or is it more or less constant?

danielru: Try 1e-6 or 1e-7... for 16 processors, this might already give you quite a lot of iterations. It seems not to change too much over the simulation, but I did not take a systematic look at it.

PR #49, opened 2015-12-03
Example with multiscale initial data for fwsw-sdc

danielru:
Adds a script that generates the figure showing the final result of using different integrators (trapezoidal, bdf-2, SDC(4), DIRK(4)) for an example with multi-scale initial data.

Generates camera ready figure.


pancetta:
Damn it, just clicked on "merge", but didn't see if there was Jenkins
feedback..

danielru:
`nosetests` are all passed in the merged master, so we should be good.

PR #110, opened 2021-04-02
Small performance boost for QMatrix computation

tlunet:
Hi Robert,

the original implementation of the QMatrix computation was suffering from 2 main downsides
1. each time _get_weights was called, it need to re-compute the same Barycentric interpolation polynomials that were already computed for the previous node => redundancy
2.  the use of quad is very expensive, and probably not needed since the integrand function is a polynomial of order M.

So I put the computation of the Barycentric polynomials directly in the _gen_QMatrix function, and changed the quad interpolation with an order M interpolation using Legendre nodes and weights. This does not change the dependencies of pySDC (nodes and weights come from scipy, quadrature use vectorized numpy operations). It improves a little bit the computation for large number of nodes (for M=100, requires only 6 sec. on my laptop, compared to 300 sec. for the previous version). From what I've compared, results are similar to machine precision (maybe if you have some regression test I could additionally check it).
Here is a comparison with the pythag package in terms of similarity, up to 150 nodes :

image

I did not remove the _get_weights method, since I'm not sure it is not used in an other function (did not modify it either ...).
So in case you're interested, feel free to take this merge request ;)

Best,
Thibaut

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,6 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+
+### Spyder ###
+.spyproject
--- a/pySDC/core/Collocation.py
+++ b/pySDC/core/Collocation.py
@@ -111,15 +111,13 @@ class CollBase(object):
     @property
     def _gen_Qmatrix(self):
         """
-        Compute tleft-to-node integration matrix for later use in collocation
-        formulation
+        Compute tleft-to-node integration matrix for later use in collocation formulation
 
         Returns:
             numpy.ndarray: matrix containing the weights for tleft to node
         """
         if self.nodes is None:
-            raise CollocationError(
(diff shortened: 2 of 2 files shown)


pancetta:
Nice, thank you for the contribution. I'll get rid of the CI errors (which come from updated Python packages) and then merge it ASAP!

pancetta:
OK, turns out I need the `_get_weights` function after all. Could you implement it with this function?

pancetta:
You can run `nosetests -v pySDC/tests` to see what's going on. Takes about 30-40 minutes to complete.

tlunet:
Done with #111 ... the nosetests gave me 3 errors and 2 fails due to packages that were not installed on my laptop (pestc4py, ...), but all the other went well. So I assume that the weight computation is not broken then ... ?

PR #114, opened 2021-09-16
Changed parameter set.

MichaelFlec:
The reason for the changes to the program run_temp_forcing_reference.py are:
 1. The program now runs significantly longer with the same number of images: long enough so that the initial nucleus can completely dissolve
 2. The thermal diffusion is now 100 times faster to achieve faster relaxiation
 3. The parameter for the driving force strength (interaction parameter) is selected to be significantly smaller in this case (dw = 1.0) so that the nucleus disappears as quickly as possible under the pressure of its surface
 4. the initial radius of the nucleus has been increased to radius = 0.35682. This corresponds to an initial solid phase volume fraction of 0.4
 5. the initial temperature field was set homogeneously to 1.0, which corresponds to the melting temperature.
 6. The width of the diffuse solid-liquid interface was chosen to be somewhat smaller: eps = 0.03

After the nucleus has completely dissolved and the system has completely balanced itself out, the dimensionless temperature in the system must be as close as possible at the homogeneous value 0.6  in order to preserve the total energy. This is simply the starting temperature minus the solid phase content.
With a changed sign in the source term of the diffusion equation (lines 125 and 141 in the AllenCahn_Temp_MPIFFT.py file) this seems to come out very nicely.
I'm excited!

For the benchmark: run_temp_forcing_benchmark.py I would recommend a slightly different set of parameters:
1. The program should cover the longest possible simulation time. After an initial phase in which all the nuclei shrink at the same time and thereby slightly lower the temperature to a temperature of local equilibrium, the system changes into the process of "ripening", which is interesting from a physical and material science perspective. In this case, large nuclei grow due to the fact that the temperature in their environment is slightly lower because small nuclei have disappeared elsewhere and have cooled the system there slightly. The ripening process is diffusion limited. The rate of diffusion is crucial for the rate of ripening.
2. the width of the diffuse solid-liquid interface was not changed. (I would have liked to have taken 0.03 in this case as well, but did not work because the convergence was too slow)
3. The thermal diffusion is now only 10 times faster: D = 1.0 (D = 10.0 is of course also possible and would also greatly accelerate the ripening process. With D = 1, however, the output is a little more interesting because the temperature field has more structure)
5. The parameter for the driving force strength (interaction parameter) is selected to be significantly larger in this case (dw = 300.0) so that the nuclei stabilize at the highest possible temperature (as close below the melting temperature as possible).

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -17,6 +17,7 @@ jobs:
         
       - name: Add packages
         run: |
+          sudo apt-get update
           sudo apt-get install texlive-latex-recommended texlive-fonts-recommended texlive-latex-extra cm-super dvipng
           
       - name: Cache conda
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 /doc/latex
 /doc/html
 *out.txt
+*out_1x1.txt
+*out_1x2.txt
+*out_2x2.txt
 *sdc.txt
 *detail.txt
 data/*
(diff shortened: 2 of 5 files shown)


pancetta:
Thanks! Working on fixing the CI pipeline, hang on.

Issue #116, opened 2022-03-01
Gitlab mirror test failing on forks


tlunet:
Just updated my fork repository (https://github.com/tlunet/pySDC) to the latest version, but then it runs some test to check a "gitlab mirror", and it does not pass (see error log in https://github.com/tlunet/pySDC/runs/5378850003?check_suite_focus=true).
Seems to be some authentication issue with the gitlab mirror repository from Helmotz Codebase repo ... 

Should we try to fix it or just ignore ?

pancetta:
Ingore or, better, delete the `.github/workflows/mirror_to_gitlavb.yml` file, i.e. [this](https://github.com/Parallel-in-Time/pySDC/blob/master/.github/workflows/mirror_to_gitlab.yml) one.

pancetta:
I think I should rather restrict the execution to this repository. This way forks will simply ignore this workflow.. I’ll have a look at it asap.

pancetta:
Look at the recent commit. I added [this line](https://github.com/Parallel-in-Time/pySDC/blob/83ac8b0c41395658cb5e204a5d67f0960261098e/.github/workflows/mirror_to_gitlab.yml#L9) which should stop the job from running on forks. Let me know if that does the trick.

tlunet:
That's weird ... now it's the pytest before that fails : https://github.com/tlunet/pySDC/runs/5389320157?check_suite_focus=true

pancetta:
That's a different issue... with the new quadrature, the error is much higher in the tests than they used to be. Since the tests take random values, sometimes the bounds are too strict, still. I have no idea what the new quadrature does this (need to set the bounds from ~`1E-14` to ~`1E-10`, it seems).

tlunet:
Can't you just set a fixed random seed for the tests (at least, one that work). And btw, what do you mean by new quadrature ? Is it the optimized implementation I provided some time ago ?

tlunet:
I'll try to look into it. But in the meantime, the mirror test passed on my development branch, so I guess this issue can be closed :wink:

PR #117, opened 2022-03-02
One collocation class to rule them all

tlunet:
Implementation of a generic collocation class ``Collocation`` in [pySDC/implementations/collocation_classes/generic.py](https://github.com/tlunet/pySDC/blob/sauron/pySDC/implementations/collocation_classes/generic.py) allowing to produce many kind of quadrature nodes from various distributions. It is based on the two main parameters that define the nodes :

- node_type : the node distribution used for the collocation method
- quad_type : the type of quadrature used (inclusion of not of boundary)

Current implementation provides the following available parameter values for node_type :

- EQUID : equidistant node distribution
- LEGENDRE : distribution from Legendre polynomials
- CHEBY-{1,2,3,4} : distribution from Chebyshev polynomials of a given kind

The type of quadrature can be GAUSS (only inner nodes), RADAU-LEFT (inclusion of the left boundary), RADAU-RIGHT (inclusion of the right boundary) and LOBATTO (inclusion of left and right boundary).

Furthermore, the ``useSpline`` option can be activated to eventually use spline interpolation when computing the weights.

Here is the equivalency table with the original classes implemented in pySDC :

| Original Class          | node_type | quad_type   | useSpline |
|-------------------------|-----------|-------------|-----------|
| Equidistant             | EQUID     | LOBATTO     | False     |
| EquidistantInner        | EQUID     | GAUSS       | False     |
| EquidistantNoLeft       | EQUID     | RADAU-RIGHT | False     |
| EquidistantSpline_Right | EQUID     | RADAU-RIGHT | True      |
| CollGaussLegendre       | LEGENDRE  | GAUSS       | False     |
| CollGaussLobatto        | LEGENDRE  | LOBATTO     | False     |
| CollGaussRadau_Left     | LEGENDRE  | RADAU-LEFT  | False     |
| CollGaussRadau_Right    | LEGENDRE  | RADAU-RIGHT | False     |

Those equivalencies are checked during regression test in [test_collocation.py](https://github.com/tlunet/pySDC/blob/sauron/pySDC/tests/test_collocation.py) with the new function ``testEquivalencies()``, that compares each attribute of the equivalent classes.

Also, it includes modifications to the ``EquidistantSpline_Right`` class (basically, reverting to the original computation of the quadrature matrix with the modified ``_getWeights`` method) in order to avoid random errors in regression tests for collocation (see [this issues](https://github.com/Parallel-in-Time/pySDC/issues/116#issuecomment-1056798555))


tlunet:
Ok, this has been corrected with [the last commit](https://github.com/Parallel-in-Time/pySDC/pull/117/commits/9f97cab8ae024bf9a93f3d91e447a04d08da8ff0). Now both new implementation and the corrected ``EquidistantSpline_Right`` class are coherent (btw. not so found of the use of ``@property`` for the ``_getNodes`` and ``_gen_Qmatrix``methods, while it's not used for ``_getWeights`` : this impose a weird manipulation in __init__ method for the Collocation class in order to take into account the spline interpolation, and furthermore, it's strange to have an attribute starting with "get" as a property (of course, just my humble opinion :wink:).

But concerning the original classes, yes you're totally right ! I could propose the following stages :
1. moving the generic class into a collocation.py module directly inside the implementations submodule
2. replacement of the original classes by an interface to the new generic one, using ``partial`` from the ``functools`` library, e.g `EquidistantSpline_Right = partial(Collocation, node_type='EQUID', quad_type='RADAU-RIGHT, useSpline=True)`
4. Add in the original class module a depreciation warning indicating that those class and imports will be removed in future versions
5. Adapt the playground and tutorials to use only the generic class
6. After some time, remove the original classes

What do you think ? And also, which part should be included in master, and which part should stay in the development branch ?

pancetta:
Thanks for the detailed explanation. I'm totally fine with getting rid of the original classes, even right now, even in the master branch. Let's be honest: we don't have 100s of users who will start crying.. the cleaner, the better. This includes removing weird usages of the `@property`..

I did hard/quick breaking API changes before and I'll do it again, if there is a clear benefit (and as long as we don't have 100s of users). Question is: are you going to do it in this PR? Do we make another PR? Should I, do you?

tlunet:
For now, I could suggest to do 1 -> 3 in this PR, since it does not involve a lot of changes from the current version, and it won't break the API. Points 4 \& 5 are however a little bit more heavy, since it implies to modify all the "old" classes in many scripts, plus an update of the regression test. I would suggest a second PR that would come later, maybe with other improvement likes a thorough cleaning of the `@property` when not necessary, of redundant implementations in some classes (like the compute_end_point that is shared by a lot of implemented Sweeper classes, and only change in a few classes => this could be written in the base class, and overwritten when there is a need of a different implementation), ...

But isn't some feature in Github where we could write and discuss all the potential improvements for pySDC (that are not necessarily issues ...) ?

pancetta:
> PS : cleaning the Github CI could be nice too, here the "Deploy" stage is actually failing 😄

Done in the master branch.

pancetta:
I second your proposal. If you don't mind, would you do steps 1-3 in this PR, then?

tlunet:
> I second your proposal. If you don't mind, would you do steps 1-3 in this PR, then?

Ok, doing that asap

tlunet:
> I second your proposal. If you don't mind, would you do steps 1-3 in this PR, then?

Done :rocket:

pancetta:
OK, I have no clue why the deploy stage is still executed.. this should only be done on `master`, but it looks like this gets ignored. Anyway.. the tests go through now, ready to merge?

pySDC/projects/AllenCahn_Bayreuth/run_simple_forcing_benchmark.py
```diff
@@ -1,7 +1,9 @@
 from argparse import ArgumentParser
+import warnings
 import numpy as np
 from mpi4py import MPI
 
+warnings.filterwarnings("ignore")
```
pancetta, line 6: What's that about?

pancetta:
And: can we go back to much more strict tolerances in `pySDC/tests/test_collocation.py`? It was originally something like `1E-14`.

pancetta:
> OK, I have no clue why the deploy stage is still executed.. this should only be done on `master`, but it looks like this gets ignored. Anyway.. the tests go through now, ready to merge?

Please merge the `upstream/master` again, I think I got the CI `if`-condition now working..

tlunet:
> > OK, I have no clue why the deploy stage is still executed.. this should only be done on `master`, but it looks like this gets ignored. Anyway.. the tests go through now, ready to merge?
> 
> Please merge the `upstream/master` again, I think I got the CI `if`-condition now working..

I'll do that, and try to set tolerances back to `1E-14` in collocation tests

tlunet:
> And: can we go back to much more strict tolerances in `pySDC/tests/test_collocation.py`? It was originally something like `1E-14`.

This currently does not work for the (new) `EquidistantInner` and `EquidistantNoLeft` classes, which have en error a little bit more that `1E-14` for M=8, little bit more `1E-13` for M=9, etc ... not sure if this is related to implementation, or simply to the fact that quadrature with equidistant nodes is not well posed

pancetta:
One tolerance to rule them all. If you can take e.g. `5E-13` for all of them, do it. Or is it much, much higher for equidistant nodes?

tlunet:
Equidistant node actually require a tolerance around `1E-11` to go up to M=12. But I'm not so sure about the degree of the polynomial that is tested, currently checking ...

tlunet:
So, after some tests, seems to me that the order of the tested polynomials are coherent. But when a tolerance of `1E-14` is set, then systematically `EquidistantInner` and `EquidistantNoLeft` fail, while `Equidistant` some times fail. The issues is then probably coming from the interpolation using Legendre quadrature on the barycentric interpolant polynomial (scipy implementation). I've tried to change the number of nodes for Legendre quadrature, but it seems to have no impact. I still wonder why the previous implementation (using the quadpack routine instead of Legendre quadrature) did not have this issue ...

pancetta:
Linter vs. @tlunet: 3-0

pancetta:
Wanna run the linter locally before getting upset...?

pySDC/projects/AllenCahn_Bayreuth/run_simple_forcing_benchmark.py
tlunet: If not, the warning message of deprecation for the "old" classes is printed in the error of the mpi process, which fails the test

pancetta:
> > Wanna run the linter locally before getting upset...?
> 
> I do, just sometimes I forget. Btw. I really don't understand why you impose this.

https://blog.devgenius.io/why-do-we-lint-97e6f134953f

I have the [black formatter](https://black.readthedocs.io/en/stable/) on my todo list, this should help to ease the pain.

pancetta:
YOU DID IT!

PR #120, opened 2022-05-12
Adaptivity

brownbaerchen:
Major question: I trigger a restart by setting Step.status.stage='RESTART', which means I have to add the abort function to the switcher in the pfasst function. Is that in line with the style of the code? Alternatively, I can introduce a variable like Step.status.restart=True or whatever. Please comment on if you think this is legible and so on.

### DIFF
--- a/pySDC/implementations/controller_classes/controller_MPI.py
+++ b/pySDC/implementations/controller_classes/controller_MPI.py
@@ -413,9 +413,9 @@ class controller_MPI(controller):
                 'IT_UP': self.it_up
             }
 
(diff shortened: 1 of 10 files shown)


pancetta:
I'd prefer to do this via a variable and not via a new stage. The `abort` function would only used for adaptivity and is thus not generic (yet).

pancetta:
In the spirit of parallelism, the MPI variant will have only one variable per step which will then communicated. For the non-MPI version the approach you now took is fine.

PR #121, opened 2022-05-16
Hotrod

brownbaerchen:
Hot Rod and Adaptivity only time-serial


pySDC/core/Level.py
```diff
@@ -22,6 +22,8 @@ def __init__(self):
         self.updated = False
         self.time = None
         self.sweep = None
+        self.e_embedded = None
+        self.e_extrapolated = None
         # freeze class, no further attributes allowed from this point
```
pancetta, line 27: Please be more expressive in your variable names, e.g. `error_est_embedded` or so.

pancetta: Better: be consistent with what you did in the controller, e.g. `error_embedded_estimate`

pySDC/implementations/controller_classes/controller_nonMPI.py
```diff
@@ -73,6 +74,22 @@ def __init__(self, num_procs, controller_params, description):
...
+        self.store_uold = self.params.use_iteration_estimator or self.params.use_embedded_estimate
+        self.restart = [False] * num_procs
+        if self.params.use_adaptivity:
+            if 'e_tol' not in description['level_params'].keys():
+                raise ParameterError('Please supply "e_tol" in the level parameters')
+            if 'restol' in description['level_params'].keys():
+                if description['level_params']['restol'] > 0:
+                    description['level_params']['restol'] = 0
+                    print(f'I want to do always maxiter={description["step_params"]["maxiter"]} iterations to have a consta\
+nt order in time for adaptivity. Setting restol=0')
```
pancetta, line 90: Use the logger here, no `print`

pySDC/implementations/controller_classes/controller_nonMPI.py
```diff
@@ -490,6 +517,13 @@ def it_check(self, local_MS_running):
         if self.params.use_iteration_estimator:
             self.check_iteration_estimator(local_MS_running)
 
+        self.error_estimator.estimate(local_MS_running)
```
pancetta, line 520: Can you think of a reason why this would need to be deactivated? E.g., data type stuff? If so, please add a flag.

pySDC/implementations/controller_classes/controller_nonMPI.py
```diff
@@ -712,3 +746,61 @@ def default(self, local_MS_running):
...
+        for i in range(len(local_MS_running)):
+            S = local_MS_running[i]
+            if S.status.iter == S.params.maxiter:
+                for l in S.levels:
+                    # throw away the final sweep to match the error estimates
+                    l.u[:] = l.uold[:]
+
+                    # check if a fault is detected
+                    if None not in [l.status.e_extrapolated, l.status.e_embedded]:
+                        diff = l.status.e_extrapolated - l.status.e_embedded
```
pancetta, line 773: Can you/do you make sure these values are set before using them? And set to something meaningful?

pySDC/implementations/controller_classes/controller_nonMPI.py
```diff
@@ -712,3 +746,61 @@ def default(self, local_MS_running):
...
+
+            # check if we performed the desired amount of sweeps
+            if S.status.iter < S.params.maxiter:
+                continue
+
+            L = S.levels[0]
+
+            # compute next step size
+            order = S.status.iter  # embedded error estimate is same order as time marching
+            h_opt = L.params.dt * 0.9 * (L.params.e_tol / L.status.e_embedded)**(1. / order)
```
pancetta, line 794: Same question here as in the hotrod

pySDC/playgrounds/HotRod/accuracy_check.py
```diff
@@ -0,0 +1,188 @@
...
+    ks = [5, 4, 3, 2]
+    fig, ax = plt.subplots(1, 1)
+    for i in range(len(ks)):
+        k = ks[i]
+        mulitple_runs(k=k, ax=ax)
+    ax.plot([None, None], color='black', label=r'$\epsilon_\mathrm{embedded}$', ls='-')
+    ax.plot([None, None], color='black', label=r'$\epsilon_\mathrm{extrapolated}$', ls='--')
+    ax.legend(frameon=False)
+    fig.savefig('error_order.pdf', bbox_inches='tight')
+
```
pancetta, line 185: You could start adding `assert`-statements to the different functions to help you testing the results.

pancetta (changes requested):

pySDC/implementations/controller_classes/controller_nonMPI.py
brownbaerchen: No, that's all "on the lowest level". You run into problems with data types atm only when storing values for Hot Rod, so that's where the error is raised. This estimate function has exactly the purpose to figure out if something should be estimated right now based on parameters as well as iterations, so I don't think there needs to be a further flag.

pySDC/implementations/controller_classes/controller_nonMPI.py
brownbaerchen: I added an assertion in case the error was not estimated before calling adaptivity, which might happen when we start doing PinT adaptivity, but that is really something to be changed in the error_estimation module, which I don't think should be changed by the user later on. So it's really just for me. I actually raise a parameter error when e_tol is not specified since that's equivalent to no supplying a step size in a fixed step scheme and I don't think there is a universally useful choice. I can ofc just put in like 1e-5 but I think forcing the user to make a choice themselves is better.

pancetta:
LGTM! How do you want to continue? Should I merge, should I wait for further additions, to you want to start a new PR?

brownbaerchen:
I think you can merge. The next step for me is make this stuff parallel, which will take quite some time and a whole new branch.

pancetta (approved):

PR #123, opened 2022-05-18
Put all the things in a project

brownbaerchen:
Added a project including tests and documentation for error estimation and what to do with it, i.e. Hot Rod and adaptivity. Also, the plot_helper can now restore the defaults if an additional flag is passed to setup_mpl

### DIFF
--- a/docs/source/pySDC/core.rst
+++ b/docs/source/pySDC/core.rst
@@ -52,6 +52,14 @@ core.Level module
    :undoc-members:
    :show-inheritance:
 
+core.Nodes module
+-----------------
+
+.. automodule:: core.Nodes
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
 core.Problem module
 -------------------
 
--- a/pySDC/helpers/plot_helper.py
+++ b/pySDC/helpers/plot_helper.py
@@ -2,6 +2,8 @@ import matplotlib as mpl
 import matplotlib.pyplot as plt
 from distutils.spawn import find_executable
 
+default_mpl_params = mpl.rcParams.copy()
+
 
 def figsize(textwidth, scale, ratio):
     fig_width_pt = textwidth                            # Get this from LaTeX using \the\textwidth
@@ -12,7 +14,10 @@ def figsize(textwidth, scale, ratio):
     return fig_size
 
 
-def setup_mpl(font_size=8):
+def setup_mpl(font_size=8, reset=False):
+    if reset:
+        mpl.rcParams.update(default_mpl_params)
+
     # Set up plotting parameters
     style_options = {  # setup matplotlib to use latex for output
         "font.family": "serif",
--- a/pySDC/projects/error_estimation/README.rst
+++ b/pySDC/projects/error_estimation/README.rst
@@ -1,0 +1,75 @@
+Error Estimation in SDC
+=======================
+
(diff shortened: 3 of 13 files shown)


docs/source/pySDC/core.rst
```diff
@@ -52,6 +52,14 @@ core.Level module
    :undoc-members:
    :show-inheritance:
 
+core.Nodes module
```
brownbaerchen, line 55: I guess sphinx created all these things automatically. I can remove them from the commit, but I didn't know what to do with them

pySDC/helpers/plot_helper.py
```diff
@@ -2,6 +2,8 @@
 import matplotlib.pyplot as plt
 from distutils.spawn import find_executable
 
+default_mpl_params = mpl.rcParams.copy()
```
brownbaerchen, line 5: This solves the problem where plots look different when generated by the CI pipeline, since the module is imported only once, so the first time plot_helper is imported, whatever are the settings then remain here until pytest is done. Ofc, if you change them and then import the helper, I guess this would mess things up, but this is the best I can do right now.

pySDC/projects/error_estimation/README.rst
```diff
@@ -0,0 +1,75 @@
+Error Estimation in SDC
```
brownbaerchen, line 1: Please give some pointers. Should I be more formal, less formal? More elaborate, more technical, more images,...?

pySDC/projects/error_estimation/piline_hotrod_adaptivity.py
```diff
@@ -0,0 +1,169 @@
...
+    v1 = np.array(sort_stats(filter_stats(stats, type='v1'), sortby='time'))[:, 1]
+    v2 = np.array(sort_stats(filter_stats(stats, type='v2'), sortby='time'))[:, 1]
+    p3 = np.array(sort_stats(filter_stats(stats, type='p3'), sortby='time'))[:, 1]
+    t = np.array(sort_stats(filter_stats(stats, type='p3'), sortby='time'))[:, 0]
+    dt = np.array(sort_stats(filter_stats(stats, type='dt'), sortby='time'))[:, 1]
+    e_em = np.array(sort_stats(filter_stats(stats, type='e_embedded'), sortby='time'))[:, 1]
+    e_ex = np.array(sort_stats(filter_stats(stats, type='e_extrapolated'), sortby='time'))[:, 1]
+    ready = np.logical_and(e_ex != np.array(None), e_em != np.array(None))
+
+    assert np.allclose([v1[-1], v2[-2], p3[-1]], [83.88431516810506, 80.62596592922169, 16.134334413301595], rtol=1),\
```
brownbaerchen, line 117: I don't love this, but this is the way to do it, right?

pySDC/projects/error_estimation/README.rst
pancetta: I have no complaints! I don't care for formal or no formal. In any case, this project will evolve over time anyway, so for now this is great!

pySDC/projects/error_estimation/piline_hotrod_adaptivity.py
pancetta: Yes..

PR #124, opened 2022-05-20
Added a function to the hooks class that counts restarts

brownbaerchen:
This is otherwise difficult because there is no function that does not overwrite entries in the stats object. This function can also be called inside a hook.


pancetta:
Ahaa, that's clever!

pancetta:
No/yes, I can add to your PR.

brownbaerchen:
Ok, that's good to know!

PR #127, opened 2022-06-01
Block Gauss-Seidel Adaptivity and Hot Rod

brownbaerchen:
This is not complete yet, but for non MPI block Gauss-Seidel this can do some form of Hot Rod and adaptivity for either one process or enough processes to avoid memory overhead due to the extrapolation based estimate.

I messed around with the stats module a bit because of the following problem: After recomputing a block with smaller step sizes, the stats object contains both the solutions before and after the restart. If we plot the quantities, they will wobble about the place because they alternate between accurate and restarted values.
To get around this, the steps now know if they are restarted and the post_step hook records whether the step was in fact restarted. Here comes the mess, though: The filter_stats function filters attributes that are set by the user when storing the value, but the restart is only recorded in the post_step hook. The implementation here is the only way I could think of where everything that gets recorded before the post_step hook is also filtered out. If you have a better idea, I would love to hear it!


pySDC/implementations/controller_classes/controller_nonMPI.py
```diff
@@ -799,13 +820,33 @@ def adaptivity(self, MS):
...
+            L.status.dt_new = h_opt
 
             # check whether to move on or restart
             if L.status.error_embedded_estimate >= L.params.e_tol:
                 self.restart[i] = True
-            else:
-                self.restart[i] = False
+                S.status.restart = True
+
+    def adaptivity_update_step_sizes(self, active_slots):
```
pancetta, line 830: Do we agree that having all those function as part of the controller makes the controller even worse in terms of readability and maintainability?

pancetta:
Please get rid of the typo in the function `resilence`.

pySDC/helpers/stats_helper.py
```diff
@@ -13,16 +15,28 @@ def filter_stats(stats, process=None, time=None, level=None, iter=None, type=Non
     Returns:
         dict: dictionary containing only the entries corresponding to the filter
     """
-
     result = {}
+
+    # check which steps have been recomputed
+    if recomputed is not None:
+        restarts = np.array(sort_stats(filter_stats(stats, process=None, time=None, iter=None, type='recomputed',
+                            recomputed=None), sortby='time'))
```
pancetta, line 23: Want to use your new convenience function here?

pySDC/helpers/stats_helper.py
```diff
@@ -13,16 +15,28 @@ def filter_stats(stats, process=None, time=None, level=None, iter=None, type=Non
...
     for k, v in stats.items():
         # get data if key matches the filter (if specified)
         if (k.time == time or time is None) and \
                 (k.process == process or process is None) and \
                 (k.level == level or level is None) and \
                 (k.iter == iter or iter is None) and \
                 (k.type == type or type is None):
-            result[k] = v
+
+            if k.time in restarts[:, 0]:
```
pancetta, line 35: Please add a comment explaining why and how

pySDC/implementations/controller_classes/controller_nonMPI.py
```diff
@@ -91,7 +91,7 @@ def __init__(self, num_procs, controller_params, description):
         if self.params.use_HotRod and self.params.HotRod_tol == np.inf:
             self.logger.warning('Hot Rod needs a detection threshold, which is now set to infinity, such that a restart\
  is never triggered!')
-        self.error_estimator = ErrorEstimator_nonMPI(self)
+        self.error_estimator = ErrorEstimator_nonMPI().get_estimator(self)
```
pancetta, line 94: So, the rest of adaptivity and resilience is part of the controller.. why not this one, too? Or, better, how can we get the other ones out of the controller?

pancetta (changes requested):

pySDC/implementations/controller_classes/controller_nonMPI.py
brownbaerchen: Naming is indeed a very difficult task. I can also make a separate resilience module. Adaptivity as a control of the step size makes the most sense inside the controller, but as we discussed earlier, this perhaps makes it difficult to give a custom adaptivity function. I will leave that here for now and make a resilience module, and then we can shift adaptivity to a better position later. I agree with all other comments.

pancetta: So, I wonder if we need more building blocks which make up "a" controller.. this way we could have just a controller, an adaptive controller, a resilient controller, both, etc.. This would be like the opposite we have now (one to rule them all)...

brownbaerchen: I would prefer individual modules that the controllers interface, like I am doing here with the error estimator. Such that, for instance, you do controller.resilience() and if you have no resilience specified, nothing happens. I think this makes it easier to mix and match different thinks and most importantly avoids copy pasting a whole bunch of code.
What we could do is something like the hooks, but for whatever, rather than just recording data. Then you can put that in the switcher and this way I think we can get about the same, but with inheritance rather than modules.
So pick you poison, because there is going to be some mess either way: Do you prefer modules or inheritance.
I started writing this comment absolutely wanting modules, but now I think maybe inheritance is the way to go. I am pretty torn honestly...

PR #132, opened 2022-06-10
Created the project PinTSimE

lisawim:
I've created the project folder for PinTSimE, which includes files for simulation of piline, buck converter and the battery drain model. Furthermore, I added documentation regarding the project, and tests to check the files.

The playground still exists so that somebody can play with it.

### DIFF
--- a/pySDC/playgrounds/EnergyGrids/playground_battery.py
+++ b/pySDC/playgrounds/EnergyGrids/playground_battery.py
@@ -1,7 +1,7 @@
 import numpy as np
 import dill
 
-from pySDC.helpers.stats_helper import filter_stats, sort_stats
+from pySDC.helpers.stats_helper import filter_stats, sort_stats, get_sorted
 #from pySDC.helpers.visualization_tools import show_residual_across_simulation
 from pySDC.implementations.collocations import Collocation
 from pySDC.implementations.controller_classes.controller_nonMPI import controller_nonMPI
@@ -19,9 +19,9 @@ def main():
     
     # initialize level parameters
     level_params = dict()
-    level_params['restol'] = 1E-10
+    level_params['restol'] = -1E-10
     level_params['e_tol'] = 1e-5
-    level_params['dt'] = 1e-3
+    level_params['dt'] = 2e-2
     
     # initialize sweeper parameters
     sweeper_params = dict()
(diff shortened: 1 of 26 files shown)


pancetta:
So, this does not yet contain the switch handling in the controller, right? You could use a local version of the controller (in the project) where this is implemented and where you can start playing around. You just need to inherit from the original one and replace/add what you need (then you don't have to copy all the other functions from the original one, just replace `it_fine` and add your switch manager).

pancetta:
You can just push to your master branch and this PR will update itself. No need to create a new one.

pySDC/projects/PinTSimE/battery_model.py
```diff
@@ -1,11 +1,11 @@
 import numpy as np
 import dill
 
-from pySDC.helpers.stats_helper import filter_stats, sort_stats
+from pySDC.helpers.stats_helper import sort_stats, filter_stats, get_sorted
 from pySDC.implementations.collocations import Collocation
```
pancetta, line 5: Please make consistent use of `get_sorted` and remove the two others from the import (and below)

pySDC/projects/PinTSimE/buck_model.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 import dill
 
-from pySDC.helpers.stats_helper import filter_stats, sort_stats
+from pySDC.helpers.stats_helper import filter_stats, sort_stats, get_sorted
 from pySDC.implementations.collocations import Collocation
```
pancetta, line 5: Same here

pySDC/projects/PinTSimE/piline_model.py
```diff
@@ -4,7 +4,7 @@
 
 mpl.use('Agg')
 
-from pySDC.helpers.stats_helper import filter_stats, sort_stats
+from pySDC.helpers.stats_helper import filter_stats, sort_stats, get_sorted
 from pySDC.implementations.collocations import Collocation
```
pancetta, line 8: Same here

pancetta (changes requested):

pySDC/playgrounds/EnergyGrids/playground_battery.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 import dill
 
-from pySDC.helpers.stats_helper import filter_stats, sort_stats
+from pySDC.helpers.stats_helper import filter_stats, sort_stats, get_sorted
```
pancetta, line 4: And one more

PR #134, opened 2022-06-20
Controller restructure

brownbaerchen:
A restructure of the controller to incorporate various methods of controlling convergence. I also adapted the switch estimator to the new style, however it does not work. (Just like in the master branch) A switch is never detected, which I think is due to the choice of tolerance in `np.isclose(t_switch - L.time, L.dt, atol=1e-2)`. I think I reproduced everything that should happen here. The CI pipeline might complain, but I thought I give you the chance to complain also in the mean time ;)

Lisa, you are welcome to checkout this branch on your local computer and make changes to it. Just do the following:
Go to my GitHub page and lookup the address of my fork. Then type `git remote add Thomas <address>`, which will lead to my fork behaving just like this main repository, which we called "upstream", on you computer. You can do `git fetch Thomas` and then create a copy of this branch in your fork with `git checkout -b <your-name> Thomas/controller_restructure`. Now you can do whatever you want to this branch and then push it to your fork, after which you can create a pull request onto my fork, which I can merge and then your changes will appear in this pull request onto the main version.


pySDC/core/Controller.py
```diff
@@ -37,7 +39,7 @@ class controller(object):
     Base abstract controller class
     """
 
-    def __init__(self, controller_params):
+    def __init__(self, controller_params, description):
```
pancetta, line 42: Hm.. why not have this as part of the `controller_params`? Or even dedicated `convergence_params`?

pancetta: I guess my problem with this is that the description goes into the steps and gets unwrapped there. It should not define the controller, esp. given that we have `controller_params`.

pySDC/core/Controller.py
```diff
@@ -226,3 +230,33 @@ def hooks(self):
...
+        if self.params.use_adaptivity:
+            convergence_controller_classes.append(Adaptivity)
+
+        if self.params.use_HotRod:
+            convergence_controller_classes.append(HotRod)
+
+        if 'convergence_controllers' in description.keys():
+            for convergence_controller_class in description['convergence_controllers']:
+                convergence_controller_classes.append(convergence_controller_class)
+
```
pancetta, line 247: Why the difference here? You have one default one (`CheckConvergence`) and then custom ones. `Adaptivity` and `HotRod` are currently.. neither/both?

pySDC/core/ConvergenceController.py
```diff
@@ -0,0 +1,70 @@
...
+    Base abstract class for convergence controller, which is plugged into the controller to determine the iteration
+    count and time step size.
+    """
+
+    def __init__(self, controller, description):
+        self.params = _Pars(description.get('convergence_control_params', {}))
+        params_ok, msg = self.check_parameters(controller, description)
+        assert params_ok, msg
+        self.reset_global_variables(controller)
+
```
pancetta, line 26: Do people HAVE to implement any of the following functions? If so, use `raise NotImplementedError` instead of `pass`.

pancetta (changes requested):
Neat! I like what I see so far.

PR #135, opened 2022-06-28
Addition of a standalone LagrangeApproximation class

tlunet:
Following a request from @lisawim, here is the first step for the implementation of a generic class for Lagrange barycentric polynomial, that would allow on middle-long term to uniform all procedures for interpolation/integration using polynomial approximation on a given set of points.

This shouldn't normally not modify the core implementation of pySDC, but let me know @pancetta if there is a need to add/correct a few aspects of this PR ...


pancetta:
I wonder why the `Lagrange` class should be in the `core` or in the `helper` part of pySDC.. what's the plan with this?

pySDC/core/Lagrange.py
```diff
@@ -0,0 +1,161 @@
...
+        raise ValueError(
+            f'Max imaginary value to important for ifft: {max(w.imag)}')
+    # -- Store weights
+    weights[:] = w.real
+
+    return nodes, weights
+
+
+class LagrangeApproximation(object):
+
```
pancetta, line 48: If this should be part of the `core` part of pySDC, I'd require a lot more documentation here and everywhere. In the current state I would prefer to let this be part of the playground.

tlunet:
The `Lagrange` should be used for the `Collocation` class and any transfer operator (interpolation, restriction). That's why I supposed it would be in the `core` subpackage, but I can totally change it to the `helper` subpackage if you want.

pySDC/core/__init__.py
```diff
@@ -1 +1,5 @@
+from pySDC.core.Nodes import NodesGenerator
+from pySDC.core.Lagrange import LagrangeApproximation
+
 __author__ = 'robert'
+__all__ = ['NodesGenerator', 'LagrangeApproximation']
```
pancetta, line 5: Do you really need this somewhere? Importing `all` is bad practice anyway.

pySDC/tests/test_projects/test_AC/test_simple_forcing.py
```diff
@@ -15,15 +15,15 @@ def test_main_parallel():
     import mpi4py
 
     nprocs = 2
-    cmd = f"export PYTHONPATH=$PYTHONPATH:$(pwd); mpirun -np {nprocs} python pySDC/projects/AllenCahn_Bayreuth/run_simple_forcing_benchmark.py -n {nprocs}"
+    cmd = f"export PYTHONPATH=$PYTHONPATH:$(pwd); export HWLOC_HIDE_ERRORS=2; mpirun -np {nprocs} python pySDC/projects/AllenCahn_Bayreuth/run_simple_forcing_benchmark.py -n {nprocs}"
```
pancetta, line 18: What's the matter here?

pancetta (changes requested):

pySDC/tests/test_projects/test_AC/test_simple_forcing.py
tlunet: So the tests won't fail on OpenSuse systems (the ones we have to deal with at TUHH ...)

pancetta: Ah, ok, thanks. Didn't know that!

pySDC/core/__init__.py
tlunet: I'm not so sure why this came up ... I'll remove it

tlunet: Oh, actually I know why : this is because of the first two lines before :

```python
from pySDC.core.Nodes import NodesGenerator
from pySDC.core.Lagrange import LagrangeApproximation
```

Those allow more convenient import statements

```
from pySDC.core import LagrangeApproximation, NodesGenerator
```
but then we have to add the `__all__` variable to avoid this kind of flake8 errors :
```
pySDC/core/__init__.py:1:1: F401 'pySDC.core.Nodes.NodesGenerator' imported but unused
from pySDC.core.Nodes import NodesGenerator
^
pySDC/core/__init__.py:2:1: F401 'pySDC.core.Lagrange.LagrangeApproximation' imported but unused
from pySDC.core.Lagrange import LagrangeApproximation
^
2     F401 'pySDC.core.Nodes.NodesGenerator' imported but unused
2
```

pancetta: OK, good.

pancetta:
> The `Lagrange` should be used (in a close future) for the `Collocation` class and any transfer operator (interpolation, restriction). That's why I supposed it would be in the `core` subpackage, but I can totally change it to the `helper` subpackage if you want.

Absolutely make sense and I appreciate this approach. I'd then require more documentation in order to keep it there. If you don't want to/can do this now, move it to the `playground` for now so that it's at least there.

tlunet:
I added some documentation in the Lagrange.py module, hope that it's enough.

Also, I also removed the deprecation warning for the "old-style" collocation class, since there is currently not alternative for the use with the controller classes.

And finally (not mentioned in the PR description), there is a small playground with a Jupyter notebook allowing to compare the accuracy of the different implementation for the QMatrix computation (original method, the current one, and the eventually future one using this LagrangeApproximation class)

pySDC/core/Lagrange.py
tlunet: Done

pancetta (approved):

pancetta:
Well done, well done. Awaiting CI results and will merge then. Note that I changed your role for the pySDC repo (did some cleanup and your account had some weird mixture of outside and inside collaborator). This should not change your access permissions. If something does not work, let me know.

Issue #118, opened 2022-03-04
Better node-to-node transfer


pancetta:
I'm pretty sure they way interpolation and restriction in time (i.e. between collocation nodes) is currently done is not a great idea, see [here](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/core/BaseTransfer.py#L67). @tlunet: Would you mind having a look at it? Can your interpolation routine make this better?

tlunet:
I could try something using a common object for interpolation and quadrature ... PR incoming. But to be clear, the get_transfer_matrix_Q method define interpolation using the Lagrange interpolating polynomial, right @pancetta ?

pancetta:
Exactly. Maybe this is all good and well, but if you have a better idea, shoot! This code is taken from `libpfasst`, I assume.

tlunet:
So, what I imagine yet would be to centralize every Lagrange-based functionalities (quadrature matrix, interpolation and transfers operators) into one main `LagrangeInterpolator` class (which would be a homemade extended `BarycentricInterpolator` class), and it would be instantiated once the nodes of the collocation class are generated. Then, the other classes would simply use its functionalities to generate weights, QMatrix, transfers matrix, interpolation matrix, etc ... (see [here](https://gitlab.com/tlunet/pythag/-/blob/master/scripts/standalone/lagrange.py) for a reference class).

If you agree with that, then should I continue working with my forked repo, or create a `dev` branch in this repo and work on it ?

pancetta:
I do STRONGLY agree with that. This sounds like a very clean structure, which would help this code. I'd prefer PRs from your fork.

tlunet:
So with #135 there is now the possibility to base the Collocation class on the LagrangeApproximation class (rather than using a modified barycentric interpolator from scipy, see https://github.com/Parallel-in-Time/pySDC/blob/8562c7c25a5819ece7deef551e3f288138caafdf/pySDC/core/Collocation.py#L10-L30

This can be one quick modification, but @pancetta you'll have to ok with the new tolerance threshold that would be used for the tests, see [pySDC/playgrounds/lagrange/quad_error.ipynb](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/playgrounds/lagrange/quad_error.ipynb)

[...]

PR #137, opened 2022-07-19
Base collocation class on a more efficient/reliable barycentric interpolation

tlunet:
There you go, following #118 ...


tlunet:
PS : a possible optimization would be to instantiate the LagrangeApproximation only once, once the node distribution on one level is fixed (and then, use the same for every block with the same node distribution, eventually with some scaling to take into account different intervals). Also, the CollBase class in core and the generic Collocation class could probably be merged, but that will be something for a more extensive rework of pySDC later, right ... ?

pancetta:
Well done, well done! Thanks for your contribution, again!

Concerning optimization: (1) I'm not sure this is needed. In terms of performance, I doubt it. In terms of style, maybe yes, but hey.. (2) Yes, merging `Collbase` and `Collocation` absolutely makes sense, also because then the Python file is gone from `implementations`. Why wait for an extensive rework?

tlunet:
Hum ... busyness and fear of the current structure of pySDC that requires plain collocation class for controllers arguments ? 

But I guess I can try to do something during my SBahn travel tomorrow ...

tlunet:
Actually, I may have to correct some error first ...

tlunet:
Guess you're good to go @pancetta :wink:

pancetta:
Thanks a lot, this is really great work!

PR #138, opened 2022-07-20
Debugging function

brownbaerchen:
Added function for debugging that displays the convergence controllers and their order since otherwise it is easy to lose track of which is the last one to change the step size for instance


pancetta:
Why don’t you use the logger for this?

brownbaerchen:
> Why don’t you use the logger for this?

Well the way I pictured the use of this function is when you develop a convergence controller you use it once to figure out which `control_order` to give it and then don't call it again the next time you run. As far as I can tell the logger at debug level just floods the console with information so it would be hard to see this. But I haven't tried the debug level in the logger so maybe I didn't understand this correctly

PR #140, opened 2022-07-21
Notebooks are now on website

brownbaerchen:


tlunet:
So proud of you :star_struck:

brownbaerchen:
> So proud of you 🤩

Thanks, I hope the notebooks make some sense :D

tlunet:
You should probably give a like to the master version, rather than to a specific commit tho ... in case the Notebook is updated, you don't want to update also the documentation ...

eg : instead of https://github.com/Parallel-in-Time/pySDC/blob/c0debd18d1e6f5a933bcbe2c90b74c297ddf7935/pySDC/projects/Resilience/Adaptivity.ipynb
put
https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/projects/Resilience/Adaptivity.ipynb

PR #141, opened 2022-07-21
Verified the error estimates with scipy

brownbaerchen:
What do you think about the "exact" solution in the Pi-line problem class? Compared to the run-time of pySDC the SciPy solution is available virtually instantaneously.


tlunet:
But is `solve_ivp` not using a numerical scheme to approximate the exact solution too ? What are actually the full equations for the Pi-Line problem (not familiar with it yet ...)

brownbaerchen:
> But is `solve_ivp` not using a numerical scheme to approximate the exact solution too ? What are actually the full equations for the Pi-Line problem (not familiar with it yet ...)

Yes, `solve_ivp` only gives me a numerical solution, but I set the tolerances to `10*np.finfo(float).eps` to get a somewhat accurate solution, which I want to use as a reference solution. I can't get an exact solution to the Pi-line problem.
I don't have any more information about it than you can find in the problem class either. It's a linear system of odes, with the rhs here: https://github.com/Parallel-in-Time/pySDC/blob/478f9a7d50ca62957582f17e4e5e5366889c34ef/pySDC/implementations/problem_classes/Piline.py#L41 @lisawim knows more about the problem.

tlunet:
Then I guess this could work, but there should be a small mesh convergence test, that checks if a more and more fine solution converges to the "analytical" solution computed with `solve_ivp`

brownbaerchen:
Well I check the "true" local error computed with this SciPy reference solution in the notebook and it has the expected order of number of sweeps plus one. That is exactly what you are asking for, right?

tlunet:
> Well I check the "true" local error computed with this SciPy reference solution in the notebook and it has the expected order of number of sweeps plus one. That is exactly what you are asking for, right?

Which notebook ?

brownbaerchen:
> > Well I check the "true" local error computed with this SciPy reference solution in the notebook and it has the expected order of number of sweeps plus one. That is exactly what you are asking for, right?
> 
> Which notebook ?

It's at the bottom above the How to use part. This is actually the entire point of this PR :D
https://github.com/Parallel-in-Time/pySDC/blob/deb8f8ebd69c6259f646915a1a18176fcc45c2c2/pySDC/projects/Resilience/HotRod.ipynb

tlunet:
Yes, that's exactly what i meant :smile: guess it's a good approach, the `solve_ivp` appears to be quite optimized. However, you should be aware that you get a 5^th order error, since it uses by default the RK54 method. I wonder if using matrix exponential would not actually provide even more accuracy (to go to k=6 or higher), while still being efficient ...

brownbaerchen:
Well I don't really care about the order of the error of the SciPy solution since I don't tell it which step size to use. So I guess the accuracy would be the same. Do you think the matrix exponential thing would be more efficient?

tlunet:
No idea ... could be something we could look at in Michigan. So, should I review this PR so it gets accepted ?

brownbaerchen:
> No idea ... could be something we could look at in Michigan. So, should I review this PR so it gets accepted ?

I would appreciate that! The functionality we want is there, so I think we can merge and, if we want to, come back and optimise it at some point.

tlunet (approved):

tlunet:
Yeah ... thing is, I don't have write access :sweat_smile: So only @pancetta could ultimately validate it

brownbaerchen:
Yeah.. I requested a review from you because I wanted some feedback to this, but I guess we need Robert or Ruth to actually merge something ;) Thanks anyways!

brownbaerchen:
That was quick!

tlunet:
> And the almighty pancetta did.

:melon:

PR #142, opened 2022-07-29
Hamiltonian

ikrom96git:
Changes in harmonic oscillator including damped case.


pancetta:
Hi @ikrom96git, thanks for the PR. Please make sure you follow [PEP8](https://peps.python.org/pep-0008/) code style conventions. The PR will fail if you don't. The easiest way to do that is to use [black](https://github.com/psf/black) right before you commit. I'm on my way porting the whole CI system to black (see branch `v5`), but this will take some time.

pancetta (changes requested):
Please do not check in figures unless you need them (e.g. if they are expensive to generate). It is better to let the CI generate the figures so that they will always be up-to-date.

If you still want to have figure in your PR, please put them into the `playground`.

pancetta:
You still have the figures in the PR.

PR #156, opened 2022-08-18
Made a convergence controller out of `check_iteration_estimator`

brownbaerchen:
The convergence controllers are only implemented in the non MPI controller, so I didn't remove it from the MPI controller. Also, I think it could do with some more documentation on what actually happens. I changed it step8/C in the tutorial worked, but I am not always sure what's going on and if this will also work with multiple steps..

### DIFF
--- a/pySDC/implementations/convergence_controller_classes/estimate_embedded_error.py
+++ b/pySDC/implementations/convergence_controller_classes/estimate_embedded_error.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-from pySDC.core.ConvergenceController import ConvergenceController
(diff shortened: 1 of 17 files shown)


pySDC/implementations/convergence_controller_classes/estimate_embedded_error.py
```diff
@@ -1,6 +1,6 @@
 import numpy as np
 
-from pySDC.core.ConvergenceController import ConvergenceController
+from pySDC.core.ConvergenceController import ConvergenceController, _Pars
```
pancetta, line 3: If you have to import an underscore method, better not call it an internal one (which is what the underscore means).

pancetta (changes requested):

PR #161, opened 2022-09-02
Bugfixes & improvements for Switch Estimator

lisawim:
The bugs are fixed in the SE (hopefully). Furthermore, I added an own method for finding the root needed to detect the switch. In estimation_check.py, for different step sizes the difference between the voltage and the reference value will plotted before the switch, after the switch (for SE=False) and exactly at the time point of the switch (SE=True). Second plot: The time evolution of the difference will be plotted for each time step.


pancetta:
Looks good so far, but why don't you use the convergence controller @brownbaerchen did? I'd hope that this way you can get away with not having to mess with the original controller whatsoever. Or are you still missing functionality?

lisawim:
The switch_controller_nonMPI is not part of the PR, so I deleted it. But the switch_estimator uses the convergence controller. Or what exactly do you mean?

PR #163, opened 2022-09-13
RK sweeper

brownbaerchen:
New sweeper that takes a lower triangular Butcher tableau as input and does Runge-Kutta. I tried to eliminate unnecessary computations by forgoing the preconditioner, rather than setting preconditioner equal to the quadrature rule.

In the jupyter notebook there is a plot that confirms the order of a few RK schemes, but I have not made any further analysis and also implemented only few RK schemes.


pancetta (changes requested):
Could you move the `test_order` into the `tests`? Then it'll be picked up during CI testing.

pySDC/playgrounds/Runge_Kutta/test_order.py
```diff
@@ -0,0 +1,53 @@
...
+        RK1: 'blue',
+        MidpointMethod: 'red',
+        RK4: 'orange',
+    }
+    ax.get_lines()[-1].set_color(colors.get(sweeper, 'black'))
+
+    label = f'{sweeper.__name__} - {ax.get_lines()[-1].get_label()[5:]}'
+    ax.get_lines()[-1].set_label(label)
+    ax.legend(frameon=False)
+
```
pancetta, line 34: Please add tests here, making sure that the expected order is reached, see e.g. [here](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/tutorial/step_1/B_spatial_accuracy_check.py#L47).

pySDC/projects/Resilience/accuracy_check.py
```diff
@@ -12,6 +12,10 @@
 from pySDC.projects.Resilience.piline import run_piline
```
pancetta, line 12: Looks like changes made here are not part of this PR? If so, please remove.

pancetta (changes requested):

pySDC/projects/Resilience/accuracy_check.py
brownbaerchen: This makes the script much more general for checking the order. I actually took it from another PR, that is true. But I need the changes to supply a custom description, which is where I set the sweeper and I use this script from the test script.

brownbaerchen:
Please have a look at the stability plots at the bottom of https://github.com/Parallel-in-Time/pySDC/blob/8201e6e4e46d1866b506b6081065456b8d313aa2/pySDC/playgrounds/Runge_Kutta/Runge-Kutta-Methods.ipynb

Are they what you would expect? I have no idea about RK methods and stability. In particular: Is implicit midpoint A-stable or did I make a mistake?

pySDC/tests/test_projects/test_resilience/test_rk.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+from pySDC.playgrounds.Runge_Kutta.test_order import test_vdp, test_advection
+
```
pancetta, line 4: So, I'd rather avoid importing from the playgrounds.. it's wild west out there. Rather put the files into your project, make a new project or add the test directly to the `test` directory (where stuff like the IMEX sweeper are tested, too).

pySDC/implementations/sweeper_classes/Runge_Kutta.py
```diff
@@ -58,18 +58,18 @@ def __init__(self, weights, nodes, matrix):
         self.delta_m[0] = self.nodes[0] - self.tleft
 
         # check if the RK scheme is implicit
-        self.implicit = any([matrix[i, i] != 0 for i in range(self.num_nodes)])
+        self.implicit = any([matrix[i, i] != 0 for i in range(self.num_nodes - 1)])
```
pancetta, line 63: We need to make sure people do not specify nodes when using this. See e.g. the Dahlquist example.. you specify the collocation class, but this will be overwritten later. This is confusing.

pancetta (changes requested):

pancetta:
> Please have a look at the stability plots at the bottom of https://github.com/Parallel-in-Time/pySDC/blob/8201e6e4e46d1866b506b6081065456b8d313aa2/pySDC/playgrounds/Runge_Kutta/Runge-Kutta-Methods.ipynb
> 
> Are they what you would expect? I have no idea about RK methods and stability. In particular: Is implicit midpoint A-stable or did I make a mistake?

LGTM. The plot for the explicit methods is somewhat hard to read, though. Could you highlight the region of stability?

pancetta:
And yes, implicit midpoint is A-stable, being a Gauss-Legendre method.

brownbaerchen:
Explicit Euler is unusually stable (entire left half plane). That's weird.. Any ideas why?

pancetta:
Where do you see this? Smells like you did implicit Euler by accident.

brownbaerchen:
> Where do you see this? Smells like you did implicit Euler by accident.

Again in the last plot in https://github.com/Parallel-in-Time/pySDC/blob/a1bfc50fe1e15799d260e0b51362cb949dd78b75/pySDC/playgrounds/Runge_Kutta/Runge-Kutta-Methods.ipynb

The stability region looks different than that of implicit Euler, though..

pancetta (approved):

PR #178, opened 2022-10-06
Added utility function for listing all available variables in stats

brownbaerchen:
Added a function that prints all available values for `type` when filtering stats produced during a pySDC run.
This is useful when using a hook that somebody else wrote or if you forgot what you did yourself.


pancetta:
Looks good, but please add a test showing how to use this.

brownbaerchen:
As it turns out, there is already a function with the same functionality only a few lines above the function called `get_list_of_types`. My bad, sorry!

PR #180, opened 2022-10-07
Added alternative text for logos that can't be dispayed on the website

brownbaerchen:
By writing `:alt: <text>` in an image block in an rst file, you can add an alternative text that will be displayed when the image can't be rendered.
Since the acknowledgements cannot be rendered on the website, I thought we can do this to pretend to be slightly more professional. 

Before:
Screenshot 2022-10-07 at 11 31 39

After:
Screenshot 2022-10-07 at 11 31 58

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,9 +99,12 @@ This project also received funding from the `German Federal Ministry of Educatio
 
 .. image:: docs/source/logo/Time-X.png
    :width: 20%
+   :alt: Time-X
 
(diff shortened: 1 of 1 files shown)


pancetta:
I originally intended to get rid of the logos with `v5`. I'm not sure what's better.

brownbaerchen:
You decide, I don't really care. I was just looking into rst files a bit.

PR #183, opened 2022-10-11
Extended switch estimator with new model(s)

lisawim:
The switch estimator is now extended for a battery drain model with n condensators. The pull request also comes with a battery drain model using two capacitors. There is a frontend to plot the differences v_C1-V_ref1 and vC2 - V_ref2 (estimation_check_extended.py).

I also added a new model - the state-space inverter. But for the first time, it is a playground, because the problem class needs to be adapted sometime in the future.

Hope that everything works. I used black and flakeheaven to check my code. Please let me know, when there is anything to improve/adapt.

Do not wonder of the big pull request.. something went wrong..


pancetta:
Hmm.. looks like there are still linting errors. Also, some files have changed that should not be part of this PR (or touched at all). 

Ideally, check out a clean version of pySDC and move the relevant files over. Then run black and flakeheaven. Maybe that helps. Welcome to git hell..

pancetta:
> More than 1k commits?!? You have been busy :D

So, it looks like this PR contains the history

PR #184, opened 2022-10-12
Extended switch estimator + new battery model

lisawim:
Now, this pull request only contains the files to be needed with a moderate number of commits.. There are an extended switch estimator, which can be used now for battery models using n condensators to predict the switches. There is also a new battery model, it consists of two condensators. Two frontends for the new problem class were added as well.

Black and flakeheaven should give no errors, hopefully..


pancetta:
Well done! Could you also add tests for the new things in your projects? If so, please add `@pytest.mark.base` as decorator (as you did in the previous PR), because otherwise the CI will not pick them up.

lisawim:
Done! The last pull request had a test for it. Now, forgot to implement it. But now, there is also a test. I only have one test added. Is that enough or should I add more tests?

pancetta:
> Done! The last pull request had a test for it. Now, forgot to implement it. But now, there is also a test. I only have one test added. Is that enough or should I add more tests?

As long as this test covers all things new to the project, I'm fine. The goal is to have 100% test coverage for the project files.

lisawim:
I will add some more tests.

pancetta:
You can run the tests locally, too, especially the ones that failed. Just use
> pytest pySDC/tests/test_projects/test_pintsime

pancetta:
One of the new tests (`test_estimation_check_extended.py`) is way too long (~30 minutes!). Please change, ~5 min. per test max.

pancetta:
Well done, thanks!

PR #188, opened 2022-10-18
CuPy-Datatype

timo2705:
Changed the CuPy Datatype for CuPy Version 11.1 with the __new__ function


pancetta:
Please run
> black pySDC  --check --diff --color

and 
> flakeheaven lint pySDC

before submitting.

PR #191, opened 2022-10-25
Added function to add variables to frozen classes

brownbaerchen:
This allows the convergence controllers to store variables in a central place, i.e. the steps of levels without having to change the step and level classes.
Since the frozen classes remain frozen, this does not impact their intent in my opinion. Accessing variables that are not added during initialisation of the step/level instance or that are added with this function which works via `level.status.__dict__["new_variable"] = "WOW"` will still raise an error.

If you are fine with this approach I will include it in the other convergence controllers while implementing their MPI versions.

Also part of this PR is some cleanup in `check_convergence.py` which does not change the functionality.

### DIFF
--- a/pySDC/core/ConvergenceController.py
+++ b/pySDC/core/ConvergenceController.py
@@ -147,15 +147,19 @@ class ConvergenceController(object):
         """
         pass
 
-    def setup_status_variables(self, controller, **kwargs):
+    def setup_status_variables(self, controller, reset=False, **kwargs):
         """
         Setup status variables.
         This is not done at the time of instatiation, since the controller is not fully instantiated at that time and
         hence not all information are available. Instead, this function is called after the controller has been fully
         instantiated.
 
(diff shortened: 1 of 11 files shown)


pancetta:
Hmm.. why can't these status variables be part of the convergence controllers? Those you can modify at will and this is where variables like these should belong, no?

brownbaerchen:
The major reason to it this way is to interface the values from other convergence controllers and from the hooks.

If I want to do adaptivity for instance, I need the error estimate in the adaptivity convergence controller. To do this, I would need to go to the controller and then look which of its convergence controllers carries the error estimates. 
The hooks in particular don't even have access to the controller, they only see the steps. 

Because of the dependencies in convergence controllers which are loaded automatically, there is also no risk of trying to access a variable that does not exist. Unless of course you made a typo, in which case the frozen class will raise an error.

PR #193, opened 2022-10-28
removed embedded error estimate variable from Level

brownbaerchen:
Your daily pull request fix

### DIFF
--- a/pySDC/core/ConvergenceController.py
+++ b/pySDC/core/ConvergenceController.py
@@ -174,22 +174,6 @@ class ConvergenceController(object):
         """
         return None
 
-    def reset_buffers(self, controller, **kwargs):
-        """
-        Buffers refer to variables used across multiple steps that are stored in the convergence controller classes to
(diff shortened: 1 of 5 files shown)


pySDC/core/ConvergenceController.py
```diff
@@ -169,29 +168,12 @@ def setup_status_variables(self, controller, **kwargs):
...
-        Args:
-            controller (pySDC.Controller): The controller
-
-        Returns:
-            None
-        """
-        pass
-
     def reset_buffers_nonMPI(self, controller, **kwargs):
         """
```
pancetta, line 178: Do you need this function?

brownbaerchen: The non MPI one I would like to keep. It doesn't really add much functionality, but it's meant to prevent bugs. By reinitialising the buffers for simulating sending between each step you make sure you don't accidentally use data sent in the last step. Also, this is called during initialisation to setup these buffers before the run.

The MPI one, however is not needed because there is only real communication.

PR #196, opened 2022-11-07
Moved variable for extrapolated error estimate from level to convergence controller

brownbaerchen:
The hooks become slightly more convoluted because I use the same hook class for different configurations of convergence controllers, such that the values I want to record don't necessarily exist.

I could implement a `get` function in the frozen class, which wraps `__dict__.get` in order to streamline this a bit. What do you think?


pancetta:
Yes please. These `__dict__.get` are ugly..

PR #199, opened 2022-11-16
Clean up + implicit battery model

lisawim:
I performed a clean up for more clarity in `estimation_check.py`. Furthermore it can plot beautiful plots regarding the usage of the switch estimator. For example the ones that I sent to you.

Then, there is the battery model in a fully implicit version with a file `battery_implicit.py` for simulate it. `estimation_check_implicit.py` can do the same as `estimation_check.py`, but for the `generic_implicit` sweeper. 

This time I was so smart to test everything using black, flakeheaven and pytest before the PR. Hopefully, all tests should be passed. Hopefully..


pySDC/projects/PinTSimE/estimation_check_implicit.py
```diff
@@ -0,0 +1,318 @@
+import numpy as np
```
pancetta, line 1: So, this file looks a lot like the original `estimation_check`. Can't you combine them and re-use what you already have? If this only changes the treatement of the ODE system, but not parameters, output, plot etc, you can make this more abstract.

pySDC/projects/PinTSimE/battery_implicit.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # filter statistics by number of iterations
+    iter_counts = get_sorted(stats, type='niter', recomputed=False, sortby='time')
+
+    # compute and print statistics
+    min_iter = 20
+    max_iter = 0
+
+    f = open(
+        'data/battery_implicit_out_dt{}_USE{}_{}.txt'.format(dt, use_switch_estimator, description['sweeper_class']),
```
pancetta, line 103: Yeah, that's not a good filename.  It breaks the CI because if contains `<` and `>`. Did you mean `.__name__`, maybe?

pancetta (changes requested):

pySDC/projects/PinTSimE/battery_implicit.py
```diff
@@ -0,0 +1,162 @@
+import numpy as np
```
pancetta, line 1: Same here. This looks a lot like your battery model, so you could simply merge those two, can't you? This way to you can "guarantee" that both scripts run the same setup, just using a different sweeper (and an adapted problem class).

lisawim: You're right. I can. This would save lot of (unnessecary) files..

pancetta:
Excellent! Welcome to the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), well done.

pancetta (approved):

PR #202, opened 2022-12-07
Update for stats_helper.py

lisawim:
This is a small PR for an update of stats_helper.py. I add the possibility to get stats back, but unsorted. This is needed, because there are some cases in which the switch estimator find more than one switch (if the first estimation is not accurate yet) and the last founded switch is needed in the end. Shortly: We need the switch which is lastly founded, not the switch which is the last (or the first) in the order sorted by time.

### DIFF
--- a/pySDC/helpers/stats_helper.py
+++ b/pySDC/helpers/stats_helper.py
@@ -34,7 +34,7 @@ def filter_stats(stats, process=None, time=None, level=None, iter=None, type=Non
     return result
 
 
-def sort_stats(stats, sortby, sorting, comm=None):
+def sort_stats(stats, sortby, comm=None):
     """
     Helper function to transform stats dictionary to sorted list of tuples
 
@@ -58,10 +58,7 @@ def sort_stats(stats, sortby, sorting, comm=None):
         result = [item for sub_result in comm.allgather(result) for item in sub_result]
 
     # sort by first element of the tuple (which is the sortby key) and return
-    if sorting:
-        sorted_data = sorted(result, key=lambda tup: tup[0])
-    else:
-        sorted_data = result
+    sorted_data = sorted(result, key=lambda tup: tup[0])
 
     return sorted_data
 
@@ -111,7 +108,7 @@ def get_list_of_types(stats):
     return type_list
 
 
-def get_sorted(stats, sortby='time', sorting=True, comm=None, **kwargs):
+def get_sorted(stats, sortby='time', comm=None, **kwargs):
     """
     Utility for filtering and sorting stats in a single call. Pass a communicatior if using MPI.
     Keyword arguments are passed to `filter_stats` for filtering.
@@ -127,6 +124,5 @@ def get_sorted(stats, sortby='time', sorting=True, comm=None, **kwargs):
     return sort_stats(
         filter_stats(stats, **kwargs),
         sortby=sortby,
-        sorting=sorting,
         comm=comm,
     )
--- a/pySDC/helpers/stats_helper.py
+++ b/pySDC/helpers/stats_helper.py
@@ -34,7 +34,7 @@ def filter_stats(stats, process=None, time=None, level=None, iter=None, type=Non
     return result
 
 
-def sort_stats(stats, sortby, comm=None):
+def sort_stats(stats, sortby, sorting, comm=None):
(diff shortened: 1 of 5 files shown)


brownbaerchen:
If I understand the situation correctly, you detect a switch and then restart with a smaller step size. Then you detect the switch more accurately and restart again with an even smaller step size. Afterwards the correct switch is the one that you added to the stats last, but which is associated with an earlier time than the more coarse switch.

Is that right? In that case, I think you don't need to modify the stats class.
Instead, it should suffice to store the switch at the time of beginning of the the step rather than the time the switch occurs here: https://github.com/Parallel-in-Time/pySDC/blob/f86bfa582c863f9128db5266eb5e02b225ba67b6/pySDC/projects/PinTSimE/switch_estimator.py#L117 .
That way you can easily match the switch to the step in which it happened after the fact and the `recomputed` stuff should be able to filter out multiple switches. Although you might run into some problems because the switch variables are added with different names in your implementation.

If that is too much hassle to change at this point, I would suggest to use instead the `filter_stats` function and implement the functionality you put into `sort_stats` somewhere inside your project. I think this is an issue really specific to your setup and putting in an argument for whether or not to sort something into a function that sorts stuff is a bit confusing in my opinion.

Then again, I added a whole bunch of stuff that is maybe not needed for everyone and confusing too, so maybe I am not the right person to make this complaint. And maybe the problem goes more deep than I thought.

If you want to add this in the `stats_helper` I would suggest adding a default argument `sorting=True` to the `sort_stats` function just like you did in the `get_sorted` function.

lisawim:
@brownbaerchen You are completely right. I see, the hooks for the switches are chosen in a bit confusing way. Therefore, I changed everything back to the origin. There is only a small update for `switch_estimator.py` in which the hooks are stored with time at the begin of a step, when a switch is found. So, it should be fixed now. Such a stupid PR..

lisawim:
Of course, it is cumbersome with the different names. The first two battery models contains a countable number of switches. But this is only the first version of the switch estimator. And this is the first idea I had to handle the problem with one than more switch. In the (near) future we get another battery model and then you do not know how many switches we have. This would be the next issue to address.

pancetta:
How about making a list of switches, then? Wouldn't that make sense?

lisawim:
What exactly do you mean by making a list of switches? Do you mean we create a list, after predicting one switch, the switch is added to list, then after predicting the second switch we add it to the list and then add the list to the stats? But in any case this probably makes more sense than giving each switch its own name..

brownbaerchen (approved):

pancetta:
Yeah, I think this is what I meant. Want to do that within this PR?

pancetta:
@brownbaerchen just mentions (offline) that you need to make sure you are able to handle multiple switches NOT detected in the correct order. Think of this scenario: you have large time-steps and detect a switch. You reduce the time-step and then detect another switch, earlier. I'm not sure how realistic that is, but if it is, this could cause problems. Maybe we should construct such a test case..

brownbaerchen:
I think you don't even need a list. Because you store in the entry class the number of restarts of the step here: https://github.com/Parallel-in-Time/pySDC/blob/f86bfa582c863f9128db5266eb5e02b225ba67b6/pySDC/core/Hooks.py#L80

That means you should be able to access different switches in the same step without doing anything extra.

Also, note that the maximum number of restarts is limited usually. You can change that in the description like here:
https://github.com/Parallel-in-Time/pySDC/blob/f86bfa582c863f9128db5266eb5e02b225ba67b6/pySDC/implementations/convergence_controller_classes/basic_restarting.py#L70

If you set that to a huge value, you can get ever more switches maybe.

lisawim:
Now for this PR: What would be the best strategy? I found a testcase for multiple switches, but this includes the usage of both, the SE and Adaptivity. In pySDC, I have not implemented Adaptivity yet. And this is not the topic of this PR. I am thinking of either closing this PR and starting a new PR (with the correct topic) or pushing the stuff so that it can be seen here under a incorrect topic of "Update for stats_helper.py", which is not changed at the end..

The third option: The PR will be accepted. What do you think?

pancetta:
I guess I prefer to close this PR and start a new one with all the neat stuff!

PR #203, opened 2022-12-14
TL: added current scripts for diagonal SDC optimization

tlunet:
Hi @pancetta 

following you email, here are some first commit of the scripts used for optimization of diagonal SDC, with the main files being

- `spectralRadius.py` : main script for multi-type optimization
- `optim.py` the home-made optimization routine used to find optimal candidates

I've set the parameters so that you get the low spectral radius values I showed during the workshop, but don't hesitate to play a bit with it to see how it changes the simulation results.
By default, all optimized coefficients are stored in the `optimDiagCoeffs.md` file, and each optimization run re-use eventual coefficients that where already found in previous script run (and only replace those if better where found).

When going to larger number of nodes, you may have to play with the `alphaFunc` coefficient (put it to 0.5 for instance), the `localOptimTol` (to be less strict with Nelder-Mead optimization) and the `threshold` to accept candidate with higher spectral radius values. Also, the more samples you use, the more chance you have to find some optimum candidate ...

Additionally, I've put the IMEX-SDC code I used for Dahlquist (`dahlquist.py`) and the associated script for convergence order analysis (`convergenceOrder.py`), FWSW stability (`stabilityFWSW.py`) and Neuman contour stability for base SDC (no IMEX, `stabilityNeumann.py`).

Enjoy, and don't hesitate to ask if you need any indication (maybe in a Github discussion :smile: ...),

Best,
Thibaut


tlunet:
Can't we just remove the playgrounds from black checks ... ? :disappointed:

pancetta:
Happy now?

pySDC/playgrounds/Diagonal/optim.py
```diff
@@ -0,0 +1,65 @@
...
+            if funcEval < res[xOrig]:
+                res.pop(xOrig)
+                res[tuple(x0)] = funcEval
+                print('     -- found better local minimum')
+        else:
+            if funcEval < threshold:
+                print('     -- found new local minimum')
+                res[tuple(x0)] = funcEval
+
+    return res, xStarts
```
brownbaerchen, line 65: I like your optimisation script a lot! I should try a similar approach with my adaptivity because it just depends so critically on the initial conditions.

pyproject.toml
```diff
@@ -84,6 +84,7 @@ flake8-comprehensions = ["+*", '-C408', '-C417']
 [tool.black]
 line-length = 120
 skip-string-normalization = true
+force-exclude = ['playgrounds']
```
brownbaerchen, line 87: 👍👍👍

pySDC/playgrounds/Diagonal/optimDiagCoeffs.md
```diff
@@ -0,0 +1,57 @@
+|   M | quadType    | distr    | optim   | coeffs                                             |         rho |
+|----:|:------------|:---------|:--------|:---------------------------------------------------|------------:|
+|   2 | GAUSS       | LEGENDRE | QmQd    | (0.105662, 0.394338)                               | 4.1514e-10  |
+|   2 | GAUSS       | LEGENDRE | QmQd    | (0.394338, 0.105662)                               | 3.06546e-10 |
+|   2 | GAUSS       | LEGENDRE | Speck   | (0.166667, 0.5)                                    | 1.44955e-10 |
+|   2 | GAUSS       | LEGENDRE | Speck   | (0.5, 0.166667)                                    | 2.51252e-10 |
```
brownbaerchen, line 6: Is there an intuitive explanation for why you can swap the values on the diagonal and get about the same spectral radius? The full Q matrix is probably not really symmetric in a similar way?

pySDC/playgrounds/Diagonal/optimDiagCoeffs.md
```diff
@@ -0,0 +1,57 @@
...
+|   2 | LOBATTO     | LEGENDRE | QmQd    | (0.0, 0.5)                                         | 2.47826e-11 |
+|   2 | LOBATTO     | LEGENDRE | Speck   | (0.0, 0.5)                                         | 9.8952e-12  |
+|   2 | RADAU-LEFT  | LEGENDRE | QmQd    | (0.0, 0.333333)                                    | 2.25197e-11 |
+|   2 | RADAU-LEFT  | LEGENDRE | Speck   | (0.0, 0.333333)                                    | 1.11866e-12 |
```
brownbaerchen, line 10: Why would the stiff limit and the non-stiff limit yield the same optimisation results? Is it maybe independent of lambda? Maybe it would be interesting to plot the spectral radius of the spectral radius of the iteration matrix with this preconditioner for different values of lambda?

brownbaerchen (approved):
I will play around with similar initial conditions for optimisation, I really like your approach! Feel free to ignore my comments. I also don't expect you to have an answer to the questions. If we could understand the magic numbers, that would be really nice, though.

pySDC/playgrounds/Diagonal/dahlquist.py
```diff
@@ -0,0 +1,350 @@
...
+        self.lamIU = deque([[c() for _ in range(self.M)] for _ in range(2)])
+        self.lamEU = deque([[c() for _ in range(self.M)] for _ in range(2)])
+        if not self.leftIsNode:
+            self.lamEU0, self.lamIU0 = c(), c()
+
+        # Instanciate list of solver (ony first time step)
+        self.lhs = [None] * self.M
+
+        self.dt = None
+        self.axpy = blas.get_blas_funcs('axpy', dtype=self.u.dtype)
```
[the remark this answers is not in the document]
tlunet: The `axpy` depends on the selected data type (hence the need to associate it to the SDC solver instance), and is used later in the sweep to optimize the addition of supplementary terms to the RHS before local system solve. That's how I see things ... but tbh, I just took the same approach as they did in Dedalus, as they seem to be used to deal with this the best way :sweat_smile:

pySDC/playgrounds/Diagonal/dahlquist.py
```diff
@@ -0,0 +1,350 @@
...
+            subproblem. The default is False.
+        """
+        # Attribute references
+        qI = self.QDeltaI
+        for i in range(self.M):
+            self.lhs[i] = 1 - dt*qI[i, i]*self.lambdaI
+
+    def _evalImplicit(self, lamIU):
+        np.copyto(lamIU, self.u)
+        lamIU *= self.lambdaI
```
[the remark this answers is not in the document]
tlunet: It avoids data duplication : if you do

```python
lamIU = self.u * self.lambdaI
```
then each time you call the _evalImplicit method, you build a new array (that is the product between `u` and `lambdaI`), and assign `lamIU` to it. 

In the proposed approach, `lamIU` is a pre-allocated array that is built only when the SDC solver is instantiated, and only its values are modified during computation. It's a bit like the _efficient_ approach you would consider when programming in C++ : never use copy constructor, instantiate all the array you need before computation, and only modify their value during computation.

Of course, for this example it does not make a big difference, since the considered arrays are not very large. But I would say it's just trying to keep some _good habits when using Python for Scientific Computing_ :smile:

pySDC/playgrounds/Diagonal/optim.py
```diff
@@ -0,0 +1,65 @@
...
+import skopt
+
+
+def findLocalMinima(func, dim, bounds=(0, 15),
+                     nSamples=200, nLocalOptim=3, localOptimTol=1e-8,
+                     alphaFunc=1, randomSeed=None, threshold=1e-2):
+    print('Monte-Carlo local minima finder')
+
+    # Initialize random seed
+    np.random.seed(randomSeed)
```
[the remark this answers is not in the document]
tlunet: I'm aware of this. But here the `skopt` package actually use the random generator under the hook, and the proposed way to set the random seed is by using `np.random.seed`. For this part, I mainly compared different random points generator and took the one that cover the more uniformly the most ground, see

https://scikit-optimize.github.io/stable/auto_examples/sampler/initial-sampling-method.html#sphx-glr-auto-examples-sampler-initial-sampling-method-py

pySDC/playgrounds/Diagonal/optimDiagCoeffs.md
tlunet: That's because in this case, you are doing only one-dimensional optimization. Since the left node coincide with the initial time-step time, then you are looking for a diagonal $Q_{\Delta}$ matrix of the form

```math
Q_{\Delta,D} = \begin{pmatrix} 0 & 0 \\ 0 & \alpha \end{pmatrix}
```

while optimizing the spectral radius of the truncated system $1 - \alpha^-1 Q[1,1]$ (for Speck optimization) and $Q[1,1] - \alpha$ (for QmQd optimization). So in this case, taking $\alpha=Q[1,1]$ does give you the optimized coefficient, see :

```python
>>> genCollocation(2, 'LEGENDRE', 'LOBATTO')[-1]
array([[0. , 0. ],
       [0.5, 0.5]])
>>> genCollocation(2, 'LEGENDRE', 'RADAU-LEFT')[-1]
array([[0.        , 0.        ],
       [0.33333333, 0.33333333]])
```

pySDC/playgrounds/Diagonal/optimDiagCoeffs.md
tlunet: No the full Q matrix is not symmetric indeed :

```python
>>> genCollocation(2, 'LEGENDRE', 'GAUSS')[-1]
array([[ 0.25      , -0.03867513],
       [ 0.53867513,  0.25      ]])
```

and I did not find any intuitive reason to explain this. I guess at this point this should be investigated by hand ...

pySDC/playgrounds/Diagonal/dahlquist.py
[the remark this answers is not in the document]
brownbaerchen: Interesting! Thanks!

brownbaerchen:
Thanks for your responses!

PR #204, opened 2022-12-14
SE and Adaptivity

lisawim:
Adaptivity is now implemented in the switch estimator(SE). The results discussed with Thomas look reasonable. As info for @brownbaerchen: a lower error tolerande for adaptivity results in less restarts. That's good! I also changed the storage for the hooks for the switch. It is know storaged at time L.time with value self.t_switch. This leads to correct storage of the founding of multiple switches. Additionally, I added description for each method in the SE for better readability and a separate function for doing interpolation and root finding stuff.

The originally implementation of adaptivity results in a bug, which I found when I considered the case of using the SE after maxiter iterations. There was a case in which a switch is found and a time step is chosen. But the time step suggested by adaptivity is smaller than the time step suggested by the SE For this reason, I added an if statement such that the time step of the SE is preferred when a switch is found. 

This is now provided for the battery model with one switch. Next PR will be adaptivity provided for the second battery model with two switches in order to keep the PR small and clearly.


pySDC/projects/PinTSimE/battery_model.py
```diff
@@ -143,8 +151,8 @@ def main(problem=battery, restol=1e-12, sweeper=imex_1st_order, use_switch_estim
     min_iter = 20
     max_iter = 0
 
-    Path("data").mkdir(parents=True, exist_ok=True)
-    fname = 'data/battery_{}.dat'.format(sweeper.__name__)
+    Path("data/{}".format(problem.__name__)).mkdir(parents=True, exist_ok=True)
```
pancetta, line 154: No subdirectories, please

pySDC/projects/PinTSimE/battery_model.py
```diff
@@ -173,49 +181,67 @@ def run():
...
-    ax.legend(frameon=False, fontsize=12, loc='upper right')
+    ax.legend(frameon=False, fontsize=8, loc='upper right')
 
-    ax.set_xlabel('Time')
-    ax.set_ylabel('Energy')
+    ax.set_xlabel('Time', fontsize=8)
+    ax.set_ylabel('Energy', fontsize=8)
 
-    fig.savefig('data/{}_model_solution_{}.png'.format(problem, sweeper), dpi=300, bbox_inches='tight')
+    fig.savefig('data/{}/{}_model_solution_{}.png'.format(problem, problem, sweeper), dpi=300, bbox_inches='tight')
```
pancetta, line 244: No subdirectories, please

pancetta:
Does this test all 4 options? With and without SE, with and without adaptivity?

lisawim:
Sure. Creates all the plots that I sent to you (except the plots using the reference errors/solution)

brownbaerchen (changes requested):
I would advise to make the tests more strict. Maybe you have some strict tests in other files and then everything is fine. But I sometimes change things, particularly regarding the stats helper and then I am anxious if the tests run successfully with your problems because I know you use all those things. And when they complete I think I didn't break anything and all is good, but at least with the iteration test in this commit, which can never go wrong, this mode of changing the code doesn't really work...

brownbaerchen:
Looks like I was too late :D
But do you actually test the cases Robert mentioned and I missed it? Like in the CI and not just by making plots that some human has to verify? In that case I am sorry..

pancetta:
No, good points you raised, Obi Wan. Will need a new PR, though, since this has been merged.

brownbaerchen:
I'll try to be more quick next time, Qui-Gon. Breakfast is overrated anyways :D

lisawim:
I will fix it as soon as possible.. Thanks for the very good points you mentioned!

PR #208, opened 2022-12-20
Adaptivity with fewer restarts

brownbaerchen:
I added a convergence controller for estimating the contraction factor. This can be used to determine whether a restart makes sense when adaptivity finds the error estimate to be larger than the set tolerance.
A restart only makes sense when the prediction for how many more sweeps we would need is larger than the number of sweeps we use for a step usually. In the van der Pol example that I tried we can avoid all restarts this way and save a few iterations. There is also a test that is executed by the CI pipeline if we save the correct number of sweeps.

Since we compute the step size such that a scheme with a given order will hopefully achieve the tolerance, we don't want to mess with the order of the scheme too much when using adaptivity. But this is exactly what we do when we continue iterating. I could not reconcile this with Hot Rod so far, but for regular Adaptivity it seems to work well to compute the next step size only after reaching `maxiter`. That means when we continue iterating past `maxiter`, we don't compute the next step size another time but move on with whatever we computed after `maxiter` iterations. Since we plan to do `maxiter` iterations in the next time step, this should work out fine.

I have to confess that I haven't really checked what happens when we do multi-step SDC / single level PFASST / composite collocation problem. (I am still struggling with communicating what I mean.)
In principle it should yield something sensible-ish, but adaptivity in that setting is very wonky as of now anyways.
There is an estimate for the embedded error in each step of a MSSDC run, but it only works for linear problems.
So I thought I would first do this for basic SDC and then we can see how to move on.
If you want me to be more thorough in this regard, I can look into this a bit more.

Another part of this PR is increasing the default number of `max_restarts` to 10. My original thinking was that when we know the order of the scheme, we should only need one restart in serial and a maximum of two in parallel, and that's still true.
But it turns out that we don't know the order as well as we want to in some settings and it doesn't make sense to restrict the code in this way. If it works than the limitation doesn't do anything and if it doesn't, this limitation just breaks everything. We need some limit in order to avoid deadlocks and I think 10 is reasonable, but feel free to suggest something else.

Also, you can suggest a better name than `wiggleroom` for this feature. Perhaps I was thinking of puppies when I decided for this name, I don't know. I am not too convinced myself.

Some changes that come from a different branch of mine have sneaked themselves into this PR. Some problem files in the Resilience project. But since I plan on using them at some point I thought there is no point in removing them from here.


pySDC/projects/Resilience/dahlquist.py
```diff
@@ -155,8 +159,133 @@ def plot_stability(stats, ax=None, iter=None, colors=None, crosshair=True, fill=
     ax.legend(frameon=False)
 
 
+def plot_contraction(stats, fig=None, ax=None, iter=None, plot_increase=False, cbar=True, **kwargs):
```
pancetta, line 162: Description missing

pySDC/projects/Resilience/dahlquist.py
```diff
@@ -155,8 +159,133 @@ def plot_stability(stats, ax=None, iter=None, colors=None, crosshair=True, fill=
...
+    if cbar:
+        divider = make_axes_locatable(ax)
+        cbar_ax = divider.append_axes('right', 0.2, pad=0.1)
+        cb = fig.colorbar(cs, cbar_ax)
+        cb.set_label(r'$\rho$')
+        cbar_ax.set_rasterized(True)
+    return cs
+
+
+def plot_increment(stats, fig=None, ax=None, iter=None, cbar=True, **kwargs):
```
pancetta, line 226: Description missing

pySDC/projects/Resilience/heat.py
```diff
@@ -0,0 +1,113 @@
...
+from pySDC.implementations.problem_classes.HeatEquation_ND_FD import heatNd_unforced
+from pySDC.implementations.sweeper_classes.generic_implicit import generic_implicit
+from pySDC.implementations.controller_classes.controller_nonMPI import controller_nonMPI
+from pySDC.core.Hooks import hooks
+from pySDC.helpers.stats_helper import get_sorted
+from pySDC.projects.Resilience.hook import log_error_estimates
+import numpy as np
+
+
+def plot_embedded(stats, ax):
```
pancetta, line 12: Description missing

pySDC/projects/Resilience/heat.py
```diff
@@ -0,0 +1,113 @@
...
+    t = [get_sorted(stats, type='u', recomputed=False)[i][0] for i in range(len(u))]
+    e_em = np.array(get_sorted(stats, type='e_embedded', recomputed=False))[:, 1]
+    e_em_semi_glob = [abs(u[i][1] - uold[i][1]) for i in range(len(u))]
+    ax.plot(t, e_em_semi_glob, label=r'$\|u^{\left(k-1\right)}-u^{\left(k\right)}\|$')
+    ax.plot(t, e_em, linestyle='--', label=r'$\epsilon$')
+    ax.set_xlabel(r'$t$')
+    ax.legend(frameon=False)
+
+
+def run_heat(
```
pancetta, line 24: Description missing

pySDC/projects/Resilience/piline.py
```diff
@@ -7,98 +7,14 @@
...
-            level=L.level_index,
-            iter=0,
-            sweep=L.status.sweep,
-            type='sweeps',
-            value=step.status.iter,
-        )
+from pySDC.projects.Resilience.hook import log_error_estimates
 
 
 def run_piline(
```
pancetta, line 13: Description missing

pySDC/projects/Resilience/piline.py
```diff
@@ -186,21 +102,23 @@ def run_piline(
 def get_data(stats, recomputed=False):
```
pancetta, line 102: Description missing

pySDC/projects/Resilience/piline.py
```diff
@@ -186,21 +102,23 @@ def run_piline(
...
+        't_restarts': np.array(get_sorted(stats, type='restart', recomputed=None))[:, 0],
         'sweeps': np.array(get_sorted(stats, type='sweeps', recomputed=None))[:, 1],
     }
     data['ready'] = np.logical_and(data['e_ex'] != np.array(None), data['e_em'] != np.array(None))
+    data['restart_times'] = data['t_restarts'][data['restarts'] > 0]
     return data
 
 
-def plot_error(data, ax, use_adaptivity=True):
+def plot_error(data, ax, use_adaptivity=True, plot_restarts=False):
```
pancetta, line 121: Description missing

pySDC/projects/Resilience/vdp.py
```diff
@@ -301,13 +244,69 @@ def mpi_vs_nonMPI(MPI_ready, comm):
         check_if_tests_match(data[1], data[0])
 
 
+def check_adaptivity_with_wiggleroom(comm=None, size=1):
```
pancetta, line 247: Description missing

pySDC/projects/Resilience/vdp.py
```diff
@@ -42,86 +43,28 @@ def plot_step_sizes(stats, ax):
     ax.set_xlabel('time')
 
 
-class log_data(hooks):
-    def post_step(self, step, level_number):
+def plot_wiggleroom(stats, ax, wiggle):
```
pancetta, line 46: Description missing

pySDC/implementations/convergence_controller_classes/estimate_contraction_factor.py
```diff
@@ -0,0 +1,115 @@
...
+        Returns:
+            None
+        """
+        for L in S.levels:
+            if L.status.error_embedded_estimate_last_iter is not None:
+                L.status.contraction_factor = (
+                    L.status.error_embedded_estimate / L.status.error_embedded_estimate_last_iter
+                )
+                if self.params.e_tol is not None:
+                    L.status.iter_to_convergence = max([0, np.ceil(
```
pancetta, line 97: no safety factor here?

pancetta:
Concerning the naming: I'm also not such a big fan of `wiggleroom`. It's not clear what's meant here. Why not `override with_iteration_estimator` as a flag (or something along those lines)?

pySDC/implementations/convergence_controller_classes/estimate_contraction_factor.py
brownbaerchen: how would the safety factor look like? Since this is the estimate, I would not put in any safety factors here. In other places where you decide what to do with this estimate, you may want to put in a safety factor.

pancetta:
Do the tests cover both `nonMPI` and `MPI`?

brownbaerchen:
> Do the tests cover both `nonMPI` and `MPI`?

For regular adaptivity, I check if MPI and non MPI versions give the same result, but I don't really know what happens with MSSDC in either case here. The implementation avoids deadlocks if used with MSSDC, but that's about all that I checked.

brownbaerchen:
Honestly, I think this is a bad strategy with MSSDC as it is implemented so far. Picture the following: I do MSSDC with four sweeps and four steps, but the first step needs three more iterations to converge. So we keep iterating, but the last step in the block will compute the step size for the next block at the same time as the first step computes its final iteration. The information will maybe be propagated somehow, but it cannot be reflected in the step size for the next block. Also, since the relation between the local error and the global error is unclear, I am not convinced that we can maintain global accuracy in MSSDC with this strategy. 

I wanted to anyways try computing the step size only on the first step in the block, which would remedy the first problem kind of.
This only works when the dynamics aren't changing too quickly, in which case we might not care about this issue anyways, I don't know. I think it's difficult to tell what to expect when doing adaptivity and MSSDC in general.

pancetta (approved):

PR #226, opened 2022-12-21
Added tests for explicit RK methods

### LINKED ISSUE #224: Explicit RK methods need tests
Yeah, we like implicit RK methods, but it looks like we should [test the explicit ones](https://parallel-in-time.org/pySDC/coverage/d_b9c274c7ffe3e178_Runge_Kutta_py.html), too.

brownbaerchen:
Closes #224


pancetta:
Want to add a test for Heun?

Issue #215, opened 2022-12-21
Test coverage of problem_helpers.py


pancetta:
Test coverage of the [problem_helpers](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/helpers/problem_helper.py) look [terrible](https://parallel-in-time.org/pySDC/coverage/d_41564483689e14fd_problem_helper_py.html). We should change this.

PR #230, opened 2022-12-22
Upwind stencils

brownbaerchen:
This adds upwind stencils from previous versions of AdvectionEquation_ND_FD.py to the FD stencil generating framework as well as tests against some previous hardcoded stencils. Since I didn't find reference values for forward and backward stencils in advection and heat equation files, I took some values from Wikipedia for this.

This improves coverage quite significantly, but the generation of 3D stencils is not covered by any test, so I am not sure if this is sufficient to solve issue #215.


pancetta:
Nice!! Can you move the tests to, well, the test directory? This way the helper is shorter and clearer (not polluted with tests which add no functionality for users).

pySDC/tests/test_helpers.py
```diff
@@ -81,3 +85,136 @@ def get_derived_from_in_package(base_class, base_package):
                 derived.append(cls)
 
     return derived
+
+
+@pytest.mark.skip(reason="This function is called from a different test with the desired parameters")
```
brownbaerchen, line 90: I am a pytest noob, so maybe this is not the desired way of doing this, but pytest wants to call every function in the test file, I think. Since this function is called from the other test, I don't want to test it once more and so I marked it to be skipped. If you don't want any skipping, another option is to move the test of one parameter combination here.

PR #239, opened 2023-01-09
Started working on problem defaults

brownbaerchen:
Continuing #207 here ...


tlunet:
Still not understanding why you need the `**kwargs` parameter for the base problem ...

brownbaerchen:
> Still not understanding why you need the `**kwargs` parameter for the base problem ...

Because of ``` self.params = _Pars(kwargs) ``` Otherwise we need to import the `_Pars` class to the derived classes and this is more convenient in my opinion.

How would you add the parameters to the class?

tlunet:
The base class does not need this `**kwargs` option, right ? So if a children class need one, then she must import it herself.

Here the `advectionNd` class actually does not need any `_Pars` class, because each argument are given in the function signature explicitly. That's the whole idea of specifying the default arguments in the constructor signature, and also avoid the bug that motivated Robert to create this Frozen class : if you give a parameter through **kwargs that is not expected by the constructor of the class, then you get an error.

brownbaerchen:
Ah, so you just want to write in the derived class `self.c = c` and so on? That makes a lot of sense actually. Especially since the only place you access problem specific variables is inside the problem class itself. So I would be fine with removing the frozen class from the problem class altogether. Does @pancetta allow us to do that?

pancetta:
Not only in the problem class, but only THROUGH it (think: hook class). Anyway, yes, please go ahead!

PR #240, opened 2023-01-09
Removed frozen class from problem class

brownbaerchen:
What do you think about the dump in the controller module? I am a bit unfamiliar with this and don't know how python version specific this is.

### DIFF
--- a/pySDC/core/Controller.py
+++ b/pySDC/core/Controller.py
@@ -165,12 +165,12 @@ class controller(object):
                     else:
                         out += '            %s = %s\n' % (k, v)
             out += '-->         Problem: %s\n' % L.prob.__class__
-            for k, v in vars(L.prob.params).items():
-                if not k.startswith('_'):
+            for k in L.prob.__init__.__code__.co_varnames:
+                if k in L.prob.__dict__.keys():
                     if k in description['problem_params']:
-                        out += '-->             %s = %s\n' % (k, v)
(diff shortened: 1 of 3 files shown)


tlunet:
Not the best idea to manipulate `__[...]__` attributes in Python. In this case, I think it should be better to write the dump directly as a method of the Problem class, with eventual overwriting of the base method in children classes

tlunet:
PS : also, the `self.param = param` can be done at the end of the constructor, which avoid having to write `self.[...]` every time you have to manipulate it in the constructor. Also, it allows to put some parameter as property instead of attributes, if they can be determined with other attributes (for instance, dx and nX can be determined directly from the grid, ndim from the A matrix ...)

brownbaerchen:
Honestly, because of the setup dump, I prefer the old way. Writing a function for the setup dump in every derived class is just annoying extra work.

Also, I don't get why you want the property functions. `N` determines the `A` matrix, not the other way around. And the problem domain is constant during a run, so why not compute `dx` only once in the beginning?
This can barely be generalised because it is just domain size / dof, but the domain size is problem specific.
Introducing abstraction for this also makes it more weird with ODEs. 

I will close this PR, because my version of what I think you want is bad. Maybe you can open a PR with how you would do it?

tlunet:
Yeah, I understood a bit what you want to do, so I'll propose a tentative PR and then we'll maybe find the optimum approach. Let's keep this PR in the closet for now ;)

PR #244, opened 2023-01-12
Tentative solution to replace FrozenClass

tlunet:
To follow up on #240 : here is one way to put class parameters as argument, and still allow to display all the main parameters from the controller class.

It is based on a `RegisterParams` class, currently implemented in Problem.py, but that could serve as (one of the) parent class of each core classes. I've applied this to the `advectionNd` class, but there is still a few issues to solve ... :

1. the call of super constructor should be done at the beginning of the children constructor. But yet it's done after the settings of `nvars`, since it needs `nvars` as tuple. Is the format for this `init` attribute of the problem class a requirement for all children class, or it is like it just for `advectionNd` @pancetta ?
2. the `ndim` argument can always be determined from the `nvars` one, and should be since there was no "prolongation" of `nvars` to larger dimension (for instance, if `nvars=4` and ndim=`3`, then `nvars` is automatically transformed to `[4,4,4]`). But the current implementation actually does not allows different size for `nvars` since the the `A` matrix is generated taking only `nvars[0]` into account. So should we keep `nvars` multidimensional for `ndim>1` or can we just use an integer for it and set the number of dimension with `ndim` ? This could also simplify the generation of xv, and only keep `xvalues` as main attribute, from which the rest can be derived (`dx`, `xv`, etc ...).
3. continuing the previous topic, `xv` can be stored as sparse, which will save memory (especially in 3D) and if we indeed keep the same mesh on each dimension, we could simply store `xvalues`, which will enough to generate the analytical solution in `u_exact`.

Other commits may come on this PR (for instance for documentation, etc ...) but let me know what you think about it yet @brownbaerchen


pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py
```diff
@@ -48,91 +47,87 @@ def __init__(
...
+        ndim = len(nvars)
         if ndim > 3:
             raise ProblemError(f'can work with up to three dimensions, got {ndim}')
 
-        if type(freq) is tuple:
-            for f in freq:
-                if f % 2 != 0 and bc == 'periodic':
-                    raise ProblemError('need even number of frequencies due to periodic BCs')
-        else:
-            if freq % 2 != 0 and freq != -1 and bc == 'periodic':
```
tlunet, line 73: I supposed that the `freq != -1` was a residue from a previous implementation, since you can avoid this condition using `nvars=(4,)` and `freq=(-1,)`, which would not work. Or is there something I did not realize @pancetta ?

brownbaerchen: No actually I implemented `freq=-1` as a Gaussian. But only for 1D, and periodic, so it should fail when you supply a tuple with more than one entry, I forgot that you can also supply `freq=(-1)`. I guess it's not the cleanest implementation...

tlunet: Do we want to keep that option ?

brownbaerchen: I want to keep the Gaussian as an option, but I am open to changing the interface.

tlunet: Ok let me see what I can do. Just to be clear, Gaussian works only with periodic boundary condition in 1D ?

brownbaerchen: It only works in 1D. If you just use it for initial conditions than ofc the boundary conditions don't matter, but the exact solution at later times only works with periodic boundaries. Also, the domain is hardcoded from 0 to 1 [here](https://github.com/Parallel-in-Time/pySDC/blob/1b0d8989584f9b1983b64d003ca6ad208cb5d6a4/pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py#L188)

tlunet: Ok thanks. But the domain was already hardcoded to [0,1] with the generation of `xvalues`, so that's not an issue :wink:

tlunet: @brownbaerchen what's the default value you want to use for `sigma` btw ?

brownbaerchen: I used `simga=6e-2` at some point. But we can take care of that later. I guess that should be decided together with the spatial resolution

brownbaerchen:
I like the changes. I am fine with the register params thing. I guess you are right that there are many things that could use a cleanup...

pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py
```diff
@@ -114,25 +114,42 @@ def __init__(
...
         # store relevant attributes
-        self.c, self.freq, self.sigma, self.dx = c, freq, sigma, dx
-        self.stencil_type, self.order = stencil_type, order
-        self.lintol, self.liniter = lintol, liniter
-        self.direct_solver, self.bc = direct_solver, bc
+        self.freq, self.sigma = freq, sigma
+        self.lintol, self.liniter, self.direct_solve = lintol, liniter, direct_solver
+
+        # read-only attributes
+        self._readOnly = [ndim, c, stencil_type, order]
```
brownbaerchen, line 125: I like this, but I would prefer a dictionary here. Since later you access variables from this list, I think it's more straightforward if you can look for a string key, not just an index. But this is fewer lines, so I don't insist. :)

tlunet: Just did that. But black absolutely want to put those properties in two lines, which add unnecessary lines. @pancetta can we set black to allow one-line functions when it's just a return statement ?

pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py
```diff
@@ -122,30 +122,26 @@ def __init__(
         self.lintol, self.liniter, self.direct_solve = lintol, liniter, direct_solver
 
         # read-only attributes
-        self._readOnly = [ndim, c, stencil_type, order]
+        self._readOnly = dict(
+            ndim=ndim, c=c, stencil_type=stencil_type, order=order)
```
brownbaerchen, line 126: I think you need to make the dictionary keys string here

tlunet: ?? That's just a simpler equivalent to

```python
self._readOnly = {'ndim': ndim, ...}
```

tlunet: Oh wait, got it

pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py
```diff
@@ -187,24 +204,27 @@ def u_exact(self, t, **kwargs):
             t (float): current time
 
         Returns:
-            dtype_u: exact solution
+            me: exact solution
```
brownbaerchen, line 207: What's your scheme for this? The name of the variable is not returned, so I personally care only about the order and what it is when looking at the return values. I get that when looking though a function this can help. What about `me (dtype_u): exact solution`?

pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py
```diff
@@ -48,91 +48,105 @@ def __init__(
...
+        self.freq, self.sigma = freq, sigma
+        self.lintol, self.liniter, self.direct_solve = lintol, liniter, direct_solver
+
+        # read-only attributes
+        self._readOnly = {
+            name: eval(name)
+            for name in ['ndim', 'c', 'stencil_type', 'order']}
+
+        # register parameters
+        self._register('nvars', 'c', 'freq', 'stencil_type', 'order', 'lintol', 'liniter', 'direct_solver', 'bc')
```
brownbaerchen, line 130: Wait, when we register `'c'` here and also have it as a readonly parameter, that is double, no?
I wonder if it would be any nicer to extend the register params class to be able to register readonly parameters.
Then you could do 
```
self._register('nvars', 'freq', 'lintol', 'liniter', 'direct_solver', 'bc')
self._register_readonly({'c': c, 'stencil_type': stencil_type, 'order': order})
```
and we could then overload the `__getattr__` method to streamline the readonly stuff a bit.

However, I wonder what's the difference between readonly and regular parameters. Neither the order, nor the boundary conditions should be allowed to be changed during the run, I feel.

tlunet: Register here just put parameters in params dictionnary, so they can be displayed later by the controller, or simply displayed all together. The readOnly parameter specifies argument for which any change will be ignored, for instance 'order' since changing it won't change the whole `A` matrix. Other arguments (like `sigma`, `lintol`, `lintiter`, etc ...) can be changed and there change would be taken into account in the object methods.

Overwriting the __getattr__ method is usually not recommended, since it add some overlay each time you want to access an attribute or use method. But I may have an idea that I can try ...

tlunet: Ok, found a way with the new commit. This can be just the standard approach when inheriting from a base class and adding supplementary parameters. First one has to store every parameters as a class attribute. Then register those parameters so they can be listed in the params property, and (optional) secure some parameters as read-only. What do you think ?

pancetta:
To (finally) answer your question concerning the `init` format: this is heaviy problem-dependent. Whatever you put as `init` into the parent's super constructor will be available as `self.init`. Think of some crazy data you may need to instantiate new data objects (e.g. from FEniCS). This is what `init` is meant for: this should contain all information necessary to instantiate new data objects and the format is matched by the corresponding data type (`mesh` in this case).

tlunet:
> To (finally) answer your question concerning the `init` format: this is heaviy problem-dependent. Whatever you put as `init` into the parent's super constructor will be available as `self.init`. Think of some crazy data you may need to instantiate new data objects (e.g. from FEniCS). This is what `init` is meant for: this should contain all information necessary to instantiate new data objects and the format is matched by the corresponding data type (`mesh` in this case).

Thanks ! I may have some small improvement to propose for that. On the same topic, why do we need different data-types for u and f ?

pancetta:
I like improvements, bring it on!

Concerning data types: the IMEX sweeper expects a different type of `f` than other sweepers. This `f` needs an implicit and an explicit part .I really don't like splitting the RHS by index (e.g. `f[:,0]` is implicit, `f[:,1]` is explicit), so it has to be by data type configuration. Then, one way to make a problem with IMEX structure is to inherit from another problem and override `feval` and `solve`, but keeping the rest (more or less). If then the data type of `f` and `u` are the same, you need to change more.

Then think of particles, where the `u` data type is a particle and the `f` acceleration or velocity or energy or whatnot.

tlunet:
Ok, I think I finshed the implementation aspect of this PR. The main change is how parameters are given and stored in the base Problem class, with an example of inheritance using the `advection_ND` class. Children constructor don't need to pass children parameters to parent constructor, but simply store it as attributes at the end, and registers the selected attributes as class parameters. Some may be changed during run-time (for instance, if it will be taken into account for in any other class method that uses it), some may **not** and can be registered as read-only parameters, which raises an error when one want to modify them. All those parameters will be accessible through the `params` property, that returns a dictionary containing names and values of each main class parameters.

Also, the problem class has now two properties that can be used to instantiate a data types for u and f. It simplify a little bit the implementation in inherited Problem classes, but the main advantage is to tell explicitly in the code how the `init` argument should interact with `dtype_u` and `dtype_f` (and will be underlined by the documentation).

:bell: Note on testing : in a near future, each inherited problem class could be tested like this ...

```python
def testProblemImplementation(probClass):
    prob = probClass()
    params = prob.params  # check if all parameters where set as attributes
    u_init = prob.u_init  # check if dtype_u works with init
    u_init = prob.f_init  # check if dtype_f works with init
```

If you are OK with this @pancetta @brownbaerchen, I'll write down the full documentation for those parts ...

pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py
```diff
@@ -48,91 +48,84 @@ def __init__(
...
-        self.Id = sp.eye(np.prod(self.params.nvars), format='csc')
+        # store relevant attributes
+        self.nvars, self.ndim, self.c = nvars, ndim, c
+        self.stencil_type, self.order, self.bc = stencil_type, order, bc
+        self.freq, self.sigma = freq, sigma
+        self.lintol, self.liniter, self.direct_solver = lintol, liniter, direct_solver
+
+        # register parameters
+        self._register('nvars', 'ndim', 'c', 'stencil_type', 'order', 'bc', readOnly=True)
+        self._register('freq', 'sigma', 'lintol', 'liniter', 'direct_solver')
```
brownbaerchen, line 128: I personally would like the `_register` function to accept a dictionary instead of `*names`. Then don't need the separate steps of storing and registering. But I get the feeling you are not a fan of dictionaries. So if you prefer this, I am fine with it as well.

brownbaerchen (commented):
Looks good to me apart from minor things. Also, we should probably have either the `helpers/pysdc_helper.py` file or the `core/Common.py` file, but not both, since they have the same purpose. I don't care what name the file has, though.

pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py
tlunet: It's not that I'm not a fan of dictionary. Actually, one could even store automatically parameters as arguments without dictionaries, like this 

```python
self._register('nvars', 'ndim', 'c', 'stencil_type', 'order', 'bc', readOnly=True, localVars=locals())
# => get the parameters values from locals() and store it directly as attribute in _register method
```
But I'm following one of Python's Zen : _"Explicit is better than implicit"_

By setting all parameters as attributes in the _register method, you hide the storing of parameters as object attribute, and never write it explicitly, _e.g_ `self.nvars = nvars`. From a developer point of view, once you are aware of this mechanism, it is indeed simpler to write and avoid this whole `self.blabla = blabla` multiple times.
But this is only for one person, and she does it only one time ! While for someone who want read the code, specifying in the constructor `self.thisPar = thisPar` (or something else) is explicit and easier to understand, especially if you are not aware of the `_register` mechanism.

pySDC/core/Common.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @property
+    def params(self):
+        return {name: getattr(self, name) for name in self._readOnly.union(self._parNames)}
+
+    def __setattr__(self, name, value):
+        try:
+            if name in self._readOnly:
+                raise ReadOnlyError(name)
+        except AttributeError:
```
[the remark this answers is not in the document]
tlunet: Actually it's not, and I believe it's even more simpler. I've put this in the new commit, along with a more explicit and consistent naming for `_readOnly` => `_parNamesReadOnly`

tlunet:
> Looks good to me apart from minor things. Also, we should probably have either the `helpers/pysdc_helper.py` file or the `core/Common.py` file, but not both, since they have the same purpose. I don't care what name the file has, though.

`helpers/pysdc_helper.py` is supposed to disappear once we apply this "default argument" approach to every class and get rid of the `FrozenClass` type. But I would keep `core/Common.py` for base classes used as parent of core class, and are not used explicitly by people implementing specialized class in `implementations` (they'll only inherited from the `Problem` class, or others, that inherit from the common classes). `helpers` could however keep all the components that could be used later for specialized class in `implementations` (for instance, the `get_finite_difference_matrix` function).

tlunet:
Just added the documentation (and also, did the same mistake as @brownbaerchen did before by merging master into my local v6 :sweat_smile:).

I used Numpy docstyle instead of Google docstyle. Yes it's produce longer documentation, but at least those can have more details and can be easily read from the code. Also, if necessary, most IDE's and editors have a way to hide docstring if necessary (for instance, if you are developing the class and don't care about documentation). Also, there are a lot of tools that generate automatically a template for the docstring from the signature of the function ...

Additional points :  

- Constructor documentation is written in the class docstring. Its parameters are implicitly considered as attributes, and additional attributes (in particular properties) are given afterward in the `Attributes` section.
- Math formula can be added in latex format, and can be useful as most IDE have now a way to render docstrings with Sphinx (I'm sure there should be something like this in vim and pycharm ;). For instance with Spyder :
grafik
- Return argument have an explicit name and associated documentation. That way, it's easy to see by looking at the code what does this function returns and what this is.

It does (probably) not break the website documentation rendering. Any comment/thought ?

brownbaerchen:
Alright! Good job! For a second I thought you already changed all the problems. Very annoying... 

While we have to go through all the problems: I am in the process of adding a [hook for storing the local error wrt to `u_exact`](https://github.com/brownbaerchen/pySDC/blob/070d16e2a6f84ca21239b7b99b911c1c4c70bc2d/pySDC/implementations/hooks/log_errors.py#L31). However, for the local error we need to solve a single step starting from the initial conditions of the step, not of the run. For this reason I added two additional arguments `u_init` and `t_init` to the [signature of `u_exact`](https://github.com/brownbaerchen/pySDC/blob/070d16e2a6f84ca21239b7b99b911c1c4c70bc2d/pySDC/implementations/problem_classes/Van_der_Pol_implicit.py#L43) for the van der Pol problem.

Currently this is the only problem supporting this and so I am unsure if everything associated should just go in my project folders rather than the implementations folder. But I do think this is a useful feature and I suggest to add the arguments to all `u_exact` functions and raise an error if you supply sth other than `None` if they don't support this feature.

Given that the [`scipy` reference solution thing](https://github.com/brownbaerchen/pySDC/blob/070d16e2a6f84ca21239b7b99b911c1c4c70bc2d/pySDC/implementations/problem_classes/Van_der_Pol_implicit.py#L75) works rather well, I think this can easily be extended to more problems then just vdp later.

What do you think? Should I keep this to myself or should we add the extra lines to all classes?

pancetta:
@brownbaerchen I'd be ok with that.

pancetta:
> Just added the documentation (and also, did the same mistake as @brownbaerchen did before by merging master into my local v6 😅).
> 
> I used Numpy docstyle instead of Google docstyle. Yes it's produce longer documentation, but at least those can have more details and can be easily read from the code. Also, if necessary, most IDE's and editors have a way to hide docstring if necessary (for instance, if you are developing the class and don't care about documentation). Also, there are a lot of tools that generate automatically a template for the docstring from the signature of the function ...
> 
> Additional points :
> 
>     * Constructor documentation is written in the class docstring. Its parameters are implicitly considered as attributes, and additional attributes (in particular properties) are given afterward in the `Attributes` section.
> 
>     * Math formula can be added in latex format, and can be useful as most IDE have now a way to render docstrings with Sphinx (I'm sure there should be something like this in vim and pycharm ;). For instance with Spyder :
>       grafik
> 
>     * Return argument have an explicit name and associated documentation. That way, it's easy to see by looking at the code what does this function returns and what this is.
> 
> 
> It does (probably) not break the website documentation rendering. Any comment/thought ?

Looks good indeed, but I worry that we're going to mess up things. There is no consistency in the documentation anymore, if we have (1) `rst` and `md` files as readmes and (2) NumPy and Google docstyle mixed throughout the code. I agree that some choices may not be optimal, but they are not dealbreakers. So, I'd rather not mix up different docstyles.

tlunet:
> > Just added the documentation (and also, did the same mistake as @brownbaerchen did before by merging master into my local v6 sweat_smile).
> > I used Numpy docstyle instead of Google docstyle. Yes it's produce longer documentation, but at least those can have more details and can be easily read from the code. Also, if necessary, most IDE's and editors have a way to hide docstring if necessary (for instance, if you are developing the class and don't care about documentation). Also, there are a lot of tools that generate automatically a template for the docstring from the signature of the function ...
> > Additional points :
> > ```
> > * Constructor documentation is written in the class docstring. Its parameters are implicitly considered as attributes, and additional attributes (in particular properties) are given afterward in the `Attributes` section.
> > 
> > * Math formula can be added in latex format, and can be useful as most IDE have now a way to render docstrings with Sphinx (I'm sure there should be something like this in vim and pycharm ;). For instance with Spyder :
> >   grafik
> > 
> > * Return argument have an explicit name and associated documentation. That way, it's easy to see by looking at the code what does this function returns and what this is.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It does (probably) not break the website documentation rendering. Any comment/thought ?
> 
> Looks good indeed, but I worry that we're going to mess up things. There is no consistency in the documentation anymore, if we have (1) `rst` and `md` files as readmes and (2) NumPy and Google docstyle mixed throughout the code. I agree that some choices may not be optimal, but they are not dealbreakers. So, I'd rather not mix up different docstyles.

This is just a suggestion for v6. We will anyway have to rework many classes with this idea of default parameters and `_register` approach intended to replace the FrozenClass (not allowing addition of new attribute as FrozenClass does can also be added ...). Also, if we tend to apply the new contribution guidelines for v6, there will also some additional rework. In the meantime, going through a rework of all class documentation won't be that much in comparison to what will already have to be done (and there is no timer on a v6 release, right ?).

So the point here is : Numpy of Google docstyle, what do you think is best ?

And about the mix `rst` and `md` : this is also a temporary solution that will have to be fixed, cf #250

tlunet:
> Alright! Good job! For a second I thought you already changed all the problems. Very annoying...
> 
> While we have to go through all the problems: I am in the process of adding a [hook for storing the local error wrt to `u_exact`](https://github.com/brownbaerchen/pySDC/blob/070d16e2a6f84ca21239b7b99b911c1c4c70bc2d/pySDC/implementations/hooks/log_errors.py#L31). However, for the local error we need to solve a single step starting from the initial conditions of the step, not of the run. For this reason I added two additional arguments `u_init` and `t_init` to the [signature of `u_exact`](https://github.com/brownbaerchen/pySDC/blob/070d16e2a6f84ca21239b7b99b911c1c4c70bc2d/pySDC/implementations/problem_classes/Van_der_Pol_implicit.py#L43) for the van der Pol problem.
> 
> Currently this is the only problem supporting this and so I am unsure if everything associated should just go in my project folders rather than the implementations folder. But I do think this is a useful feature and I suggest to add the arguments to all `u_exact` functions and raise an error if you supply sth other than `None` if they don't support this feature.
> 
> Given that the [`scipy` reference solution thing](https://github.com/brownbaerchen/pySDC/blob/070d16e2a6f84ca21239b7b99b911c1c4c70bc2d/pySDC/implementations/problem_classes/Van_der_Pol_implicit.py#L75) works rather well, I think this can easily be extended to more problems then just vdp later.
> 
> What do you think? Should I keep this to myself or should we add the extra lines to all classes?

I do aggree (too), see Discussion #253

pancetta:
Agreed, agreed. I just wanted to make sure we're not mixing up too many things. Concerning docstyles: I'm ok with either of these. NumPy's would be probably more fitting from what I see and it would give us the chance of checking all docstrings for consistency and correctness.

pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py
pancetta: I'd rather have one mechanism of "registering" attributes than two (which have to be done both). If you are clear about what the `_register` method does, I think there will be no confusion. Maybe with a more explicit name for this method (e.g. `_make_attribute_and_register`) this is not such a problem? The current proposal is not clear to me (if I were someone who wants to read the code): why do I need to store AND register? And what happens if I store but don't register?

tlunet: @pancetta I do like you idea of making the method name more explicit. I've implemented this in the last commit. Also on a side note : paramters are stored in `set`, as looking up an element is faster than in a list. But then the order they are given in the `_makeAttributeAndRegister` function does not matter, only the alphabetic order is used, see :

image

Since setting attribute of the class method is (theoretically) done only during the instantiation with `__init__`, it should not degrade that much performance to use `list` instead, which would allow to keep the parameter order in the `params` dictionary. Would you prefer that ?

pancetta: Any scenario where the order would matter? If not, a `set` it is.

pancetta:
> helper

Concerning helpers: the fewer we have the better. If you see a way to get rid of some of then by integrating them into some core functionality, feel free!

tlunet:
@brownbaerchen @pancetta can we merge this into v6 ? So I can focus on #254 (see all steps proposed in #255 )

pancetta:
> @brownbaerchen @pancetta can we merge this into v6 ? So I can focus on #254 (see all steps proposed in #255 )

LGTM

PR #246, opened 2023-01-13
Multiple Hooks simultaneously 

### LINKED ISSUE #243: Allow multiple hooks simultaneously
In the current implementation, you can supply one user-defined hook class, which has to inherit from the core hook class.
The problem I want to resolve is the following: I have written a hook class recording parameters to my specific problem, but now for one plot I want to record one additional parameter. In my case that is the error compared to some reference solution which is not cheap to obtain, thus I don't just want to add it to the hook I use all the time. Currently that means I have to copy paste the hook class or add a level of inheritance, which get's messy quickly.

I would like the hooks module to have a similar structure to the convergence controller module.
That is to say rather than doing sth like
```
self.hook.pre_run(...)
```
I want 
```
for hook in self.hooks:
    hook.pre_run(...)
```
and so on in the controller.

That way I can easily write a separate hook class recording only the error wrt a reference solution and supply both to the controller as an array like object.

In my projects I could save a bunch of lines of code. In addition to not repeating statements in different hook classes, I could also forgo calling the parent methods via `super` all the time. Currently, doing this is mandatory, since timing variables will otherwise be missing.
Also, we could add some default hooks in an implementation folder, for instance for recording the solution at the end of the step, or I could write a hook that records the embedded error estimate that could be added automatically when adding the respective convergence controller and so on.

The `__stats` dictionary would have to be moved from the hook class to the controller class such that all hooks add entries to the same dictionary. Additionally, I would `pass` in all methods of the core hook class and instead write a default hook class that records all the stuff the core module currently does.

What do you think, @pancetta ? Yes, no, maybe?

brownbaerchen:
You can now add hooks either as a single hook, so the API is backwards compatible, or as an array of hooks. All hooks will be called one after another in the order that they are added whenever previously a single hook was called. The order should not matter since the hooks are not supposed to modify anything.
I wrote a few hooks that record small pieces of data that is interesting when using a certain convergence controller, which are added automagically when the respective convergence controller is added. For instance after each step, the step size is recorded as 'dt' when using adaptivity and so on.

Since the hooks have no access to the controller, I left the `__stats` variable in the hooks and merge them after the fact in the controller.

Since I need the `__num_restarts` variable for `add_to_stats` such that I can filter recomputed values afterwards, it is unfortunately still necessary to call the parent method with `super().whatever()` just like before.
However, for most people it doesn't matter. Actually, my tests pass without calling the parent's methods, but there is a number of fringe cases where it might make a difference. So whenever we write general hooks, we should keep calling the parent's method, but individual projects can choose to forgo this if they never restart anything.

Closes #243

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -2,7 +2,7 @@ name: CI pipeline for pySDC
 
 on:
   push:
-  pull_request:
+  pull_request_target:
   schedule:
     - cron: '1 5 * * 1'
 
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -3,6 +3,7 @@ name: CI pipeline for pySDC
 on:
   push:
   pull_request_target:
+    types: [opened, edited, closed, reopened, synchronize, ready_for_review, review_requested , review_request_removed, auto_merge_enabled,  auto_merge_disabled]
   schedule:
     - cron: '1 5 * * 1'
 
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -2,7 +2,7 @@ name: CI pipeline for pySDC
 
 on:
   push:
-  pull_request_target:
+  pull_request:
   schedule:
     - cron: '1 5 * * 1'
 
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -3,7 +3,6 @@ name: CI pipeline for pySDC
 on:
   push:
   pull_request_target:
-    types: [opened, edited, closed, reopened, synchronize, ready_for_review, review_requested , review_request_removed, auto_merge_enabled,  auto_merge_disabled]
   schedule:
     - cron: '1 5 * * 1'
 
--- a/pySDC/core/Controller.py
+++ b/pySDC/core/Controller.py
@@ -42,12 +42,11 @@ class controller(object):
 
         # check if we have a hook on this list. If not, use default class.
         self.__hooks = []
-        self.hook_classes = [default_hooks]
+        hook_classes = [default_hooks]
         user_hooks = controller_params.get('hook_class', [])
-        self.hook_classes += user_hooks if type(user_hooks) == list else [user_hooks]
-        for hook in self.hook_classes:
-            self.__hooks += [hook()]
-        controller_params['hook_class'] = controller_params.get('hook_class', self.hook_classes)
+        hook_classes += user_hooks if type(user_hooks) == list else [user_hooks]
+        [self.add_hook(hook) for hook in hook_classes]
+        controller_params['hook_class'] = controller_params.get('hook_class', hook_classes)
 
         for hook in self.hooks:
             hook.pre_setup(step=None, level_number=None)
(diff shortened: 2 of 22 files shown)


brownbaerchen:
Why is `mirror_to_gitlab` not working? It works now on my fork, but not here. Jakob mentioned that you should use `on: pull_request_target` instead of `on: pull_request` [here](https://github.com/Parallel-in-Time/pySDC/blob/1b0d8989584f9b1983b64d003ca6ad208cb5d6a4/.github/workflows/ci_pipeline.yml#L5), I think.
This is also what he writes in the [readme](https://github.com/jakob-fritz/github2lab_action) now.
No idea, if this is what causes the problem here, though.

jakob-fritz:
This is exactly, what causes the problem.
With "on: pull_request" the secrets from the source are used, whereas with "on: pull_request_target" the secrets from target are used.

This was missing in the first version of the readme and is now added as brownbaerchen mentioned.

brownbaerchen:
@jakob-fritz Now the pipeline is not executed when I start a PR. What did I do wrong?

jakob-fritz:
I don' know, why it is not shown. When I look at the commit itself I can see a GitHub-Workflow that [started](https://github.com/brownbaerchen/pySDC/actions/runs/3911542552)

I have no idea, why this is not shown in the merge-request

brownbaerchen:
> I don' know, why it is not shown. When I look at the commit itself I can see a GitHub-Workflow that [started](https://github.com/brownbaerchen/pySDC/actions/runs/3911542552)
> 
> I have no idea, why this is not shown in the merge-request

No that's different. That is run on my fork on push. But since I can do whatever on my fork, we need to run it here as well. The actions tab does not show any action run for the latest commit on this PR on the main repo.

jakob-fritz:
Then, maybe @pancetta needs to confirm this execution first? As the secrets (and potentially the compute-time in other cases) from his repo are used?

pancetta:
Can't do. Nothing to confirm here!

brownbaerchen:
People seem to have some security concerns regarding `pull_request_target`. And it makes sense that enabling anybody to run anything on KIT machines is not what we want to do here. As #194 demonstrates, this is actually a valid concern. Before we merge something like this into master, we should consider opening this only to known collaborators, if that's an option. Maybe the gitlab stuff should be skipped if a first time collaborator opens a PR with Robert having to start the gitlab stuff manually?

This actually raises the following point: If anyone can change the pipeline to `pull_request_target`, than anyone can just enable themselves to download pictures of famous people to KIT machines. Is this not working because of safety reasons? Does Robert have to change to `on: pull_request_target` in the master branch first?

jakob-fritz:
> People seem to have some security concerns regarding `pull_request_target`. And it makes sense that enabling anybody to run anything on KIT machines is not what we want to do here. As #194 demonstrates, this is actually a valid concern. Before we merge something like this into master, we should consider opening this only to known collaborators, if that's an option. Maybe the gitlab stuff should be skipped if a first time collaborator opens a PR with Robert having to start the gitlab stuff manually?
> 
> This actually raises the following point: If anyone can change the pipeline to `pull_request_target`, than anyone can just enable themselves to download pictures of famous people to KIT machines. Is this not working because of safety reasons? Does Robert have to change to `on: pull_request_target` in the master branch first?

That totally makes sense to be sceptic on that. In the documentation I read, that first-time contributors need to be confirmed to the workflow to run in pull-requests (this is why I was wondering if he has to confirm it). This is even more important when using other services (e.g. Gitlab on KIT-machines) and after seeing #194 (no words on that).

So your point is valid, that it makes sense that "pull_request" can only be changed to "pull_request_target" from within the main repo and not from the outside.
If you do not replace it but add it, does might the pipeline then run again (with an error of not accessing the secrets from the main repo)? This might be an indicator, that the "_target"-part is unused until part of the main repo.

pancetta:
> That totally makes sense to be sceptic on that. In the documentation I read, that first-time contributors need to be confirmed to the workflow to run in pull-requests (this is why I was wondering if he has to confirm it).

And I did confirm that @brownbaerchen can run stuff during his PRs. I think it is the duty of the maintainer to take care of PR's content, i.e. if a PR from a first-time contributor contains crap, then get rid of it. I cannot prevent a known contributor from going rogue, though.

pancetta:
I think I like this one!

PR #256, opened 2023-01-17
Test faults

brownbaerchen:
I add soft faults via a hook and I used to do it with inheritance. However, this was not compatible with the current hook API anymore.
I used the opportunity to clean up most of the hooks in the resilience project folder and adapted the notebooks to the new changes.

Also, I added some tests for generating fault statistics. These are part of the mpi tests because different processors generate statistics for different strategies in parallel. All individual experiments are run in serial.
The tests are a bit flaky because I get overflow errors on GitHub, but not on my laptop. The result is a test that is not as strict as I would like, but I think it's better than nothing. Unfortunately I did not find a satisfactory solution to fix the overflow behaviour across machines so far.

Since I need to generate plots with the fault stuff, I wanted to get everything working again and maybe come back to the flaky test when I have more experience/patience/wisdom/beer/time.


brownbaerchen:
I am confused how adding more tests reduced the coverage this much... Do files that don't get tested at all not count into the percentage? Otherwise it is weird to me that testing half of the `fault_stats` file is worse than not testing it at all. The other changes in coverage are rather small...

pancetta:
> I am confused how adding more tests reduced the coverage this much... Do files that don't get tested at all not count into the percentage? Otherwise it is weird to me that testing half of the `fault_stats` file is worse than not testing it at all. The other changes in coverage are rather small...

Don't know. Untested files are of course counted toward the overall coverage (unless they are in the `test` or `playground` directories, which are explicitly excluded from coverage. You can look at the coverages of both `master` branch and your PR [here](https://app.codecov.io/gh/Parallel-in-Time/pySDC/pull/256). Does that help?

The "classical" coverage report (see artifacts of your CI run) show the same numbers, btw.

tlunet:
grafik

Seems like adding custom-made classes for a given project (here hook) without specific testing is a coverage killer :sweat_smile:

brownbaerchen:
I failed at the coverage thing :( Still convinced that this PR improves coverage, though

pancetta:
Please merge latest commits from `master`.. happy now? 😢

pancetta:
What's the status here?

brownbaerchen:
Codecov is complaining that the test coverage is too low in the files that I touched. Given that the overall test coverage is still up by 2%, I suggest merging this anyways.
I resolved some merge conflicts, let's see if the tests go through before doing anything.

PR #258, opened 2023-01-18
Inheritance for Battery, hardcoded references and more

lisawim:
See PR #247 for description.


pySDC/projects/PinTSimE/estimation_check_2capacitors.py
```diff
@@ -59,18 +72,17 @@ def run(dt, use_switch_estimator=True):
 
     # fill description dictionary for easy step instantiation
     description = dict()
-    description['problem_class'] = battery_2condensators  # pass problem class
+    description['problem_class'] = battery_n_capacitors  # pass problem class
     description['problem_params'] = problem_params  # pass problem parameters
     description['sweeper_class'] = imex_1st_order  # pass sweeper
     description['sweeper_params'] = sweeper_params  # pass sweeper parameters
     description['level_params'] = level_params  # pass level parameters
     description['step_params'] = step_params
```
brownbaerchen, line 80: Instead of generating the description object from scratch, you could separate the `run` function into `generate_description` and `run`. Then you can put the `generate_description` function into `batter_2capacitors_model.py` and in the run here you only change the parameters you need different. Meaning
```
def run(dt):
    description = generate_description()
    description['level_params']['dt'] = dt
    controller = ...
    controller.run(...)
    ...
```
The same applies to the estimation check for a single capacitor.

pySDC/projects/PinTSimE/battery_2capacitors_model.py
```diff
@@ -0,0 +1,304 @@
...
+from pySDC.implementations.controller_classes.controller_nonMPI import controller_nonMPI
+from pySDC.projects.PinTSimE.battery_model import get_recomputed, proof_assertions_description
+from pySDC.projects.PinTSimE.piline_model import setup_mpl
+import pySDC.helpers.plot_helper as plt_helper
+from pySDC.core.Hooks import hooks
+
+from pySDC.projects.PinTSimE.switch_estimator import SwitchEstimator
+
+
+class log_data(hooks):
```
brownbaerchen, line 18: Maybe one hook for the whole project is enough? You log really similar things most of the time. If the solution is different, you can just log the whole solution and separate it into components later.

I recently tried to streamline this a bit, so, if you want to keep the names in the stats object, you can also have one hook that logs all quantities that are the same and then make separate hooks that log things that may be different across problems. Then you just add multiple hooks as a list.

pySDC/projects/PinTSimE/estimation_check_2capacitors.py
```diff
@@ -86,56 +98,37 @@ def run(dt, use_switch_estimator=True):
...
-
     return stats, description
 
 
 def check(cwd='./'):
     """
     Routine to check the differences between using a switch estimator or not
+
+    Args:
+        cwd: current working directory
```
brownbaerchen, line 109: `cwd (str): current working directory`

pySDC/tests/test_projects/test_pintsime/test_estimation_check_2capacitors.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+
+@pytest.mark.base
+def test_main():
+    from pySDC.projects.PinTSimE.estimation_check_2capacitors import check
+
+    check()
```
brownbaerchen, line 8: Why do you test both the estimation check and the standard model for 2 capacitors, but for 1 capacitor you test nothing?

pancetta:
LGTM. Any comments, @brownbaerchen ?

brownbaerchen (changes requested):
One more iteration...

I am sorry for complaining so much! The code is already really good, but I think we can get even better by removing yet more lines :)

While fewer lines is certainly not always more readable, repeating similar lines makes understanding code difficult because you have to search for the important parts. Black doesn't help with this when it spreads a dictionary with two entries across 1000+ lines...
Also, if you decide on a change for parameters for running a problem, for instance, you need to go through all files that define the parameters and change it everywhere.
If you do this only in a single location and import it from other files rather than repeating it, both reading and changing the code becomes much faster.

Whenever I implement a new feature, I think "What feature is already there that is similar to what I want and how can I implement this new thing with utilising as much of the old stuff as possible?" Abstraction often means more work now, but less work later. For instance, if the old stuff is well tested and the new stuff doesn't work, I know the problem is only in the few things I changed.

Finally, I want to reiterate that I know nothing about your problem, so I am sure many of my comments are plain stupid and you can just ignore them.

pySDC/projects/PinTSimE/estimation_check_2capacitors.py
lisawim: I did another kind of doing this. I haven't pushed it yet, because I want to change some other things that you mentioned. I hope you will be satisfied with the new changes.

pySDC/projects/PinTSimE/battery_2capacitors_model.py
lisawim: I saw that you used type='u' for the whole solution u? So, it is possible to do this in my hooks? In case of yes I can adapt my hooks, and for each number of capacitors I can use only one hook I guess.

pySDC/projects/PinTSimE/battery_model.py
```diff
@@ -62,11 +62,31 @@ def post_step(self, step, level_number):
...
 
 
 def main(dt, problem, sweeper, use_switch_estimator, use_adaptivity):
     """
     A simple test program to do SDC/PFASST runs for the battery drain model
+
+    Args:
+        dt (float): time step for computation
+        problem (problem_class.__name__): problem class that wants to be simulated
+        sweeper (sweeper_class.__name__): sweeper class for solving the problem class numerically
```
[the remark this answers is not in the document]
lisawim: You mean pySDC.core.Sweeper.sweeper I guess.

pySDC/projects/PinTSimE/battery_2capacitors_model.py
brownbaerchen: Yes, you can do it too!
In your hook currently you add `L.uend[1]` as `voltage C` for instance.
If you add `L.uend` as `u`, for instance by using the [predefined hook](https://github.com/Parallel-in-Time/pySDC/blob/53d6faf64827db05d6274f44002e0ab3f44a4b9d/pySDC/implementations/hooks/log_solution.py#L4), you can sort everything out like this:
```
u = get_sorted(stats, type='u')
V_C = [me[1][1] for me in u]
```
You can just add the index back when sorting the stats if you leave it out when logging them.

I started the way you did it because it's nice to have sensible names for everything, but for the sake of abstraction, I just went ahead and called everything `u` :D

pySDC/implementations/problem_classes/Battery.py
```diff
@@ -280,11 +270,131 @@ def solve_system(self, rhs, factor, u0, t):
...
+            dtype_u=dtype_u,
+            dtype_f=dtype_f,
+            params=problem_params,
+        )
+
+        v = np.zeros(n + 1)
+        v[0] = 1
+
+        self.A = np.zeros((n + 1, n + 1))
+        self.switch_A = {k: np.diag(-1 / (self.params.C[k] * self.params.R) * np.roll(v, k + 1)) for k in range(n)}
```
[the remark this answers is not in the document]
lisawim: And if I understand correctly, you want to say that I should define I function outside __init__ where I can define self.switch_A, self.switch_f as dictionaries and then I can say, class battery should inherit from class battery_n_capacitors?

I defined a function get_problem_dict() and __init__() calls then switch_A, switch_f = get_problem_dict()

lisawim:
Done!! Is @brownbaerchen now a happy baerchen? :D Or can I improve some more for this PR?

pySDC/implementations/problem_classes/Battery.py
```diff
@@ -115,57 +270,64 @@ def u_exact(self, t):
...
+            if u[m][1] - self.params.V_ref[0] <= 0:
+                switch_detected = True
+                m_guess = m - 1
+                break
+
+        vC_switch = [u[m][1] - self.params.V_ref[0] for m in range(1, len(u))] if switch_detected else []
+
+        return switch_detected, m_guess, vC_switch
+
+    def count_switches(self):
```
brownbaerchen, line 304: This function is the same that is inherited, I think.

brownbaerchen:
Yes, very happy! Well done! Just one tiny thing: I think you forgot to remove the `count_swtiches` function in the battery problem class. But this is my final complaint. I really like what you've done! :)

brownbaerchen:
Just one comment on preconditioners: I believe you don't set a specific explicit preconditioner for IMEX problems, which means you are using explicit Euler. Did you try Picard? According to Robert that often gives faster convergence.

lisawim:
No, I did not try it. I only used implicit and explicit Euler as default. But it should worth trying it.

brownbaerchen:
Preconditioner performance is problem specific, so this might be totally worse for your problems. Feel free to stick with explicit Euler, in particular because you need to rewrite all test result otherwise. I was just wondering.

lisawim:
Say, this may be something that can be explored further in the future.

brownbaerchen:
Yes, let's merge this now, finally! (provided the CI agrees)

PR #260, opened 2023-01-19
TL: minor doc correction

tlunet:


tlunet:
Trying to solve a weird behavior here :

- merged latest changes of master into v6 (#259)
- pipeline for the new v6 failed (since v6 is yet in rework, this is expected)
- pipeline for the last commit on master (1c235f941bf0c04367d3b70b211a070b3ccce59e) failed now ... :confused:

PR #261, opened 2023-01-19
Added Lorenz attractor problem

brownbaerchen:
Simple script that solves the non-linear equations implicitly. I also added a scipy reference solution analogously to the van der Pol problem. I used this reference solution to verify the implementation of the Jacobian. The solution also has the familiar butterfly shape.
I tried to write everything such that the changes going to v6 will be minimal, that is to say I already put in defaults instead of essential keys.

I added a hook for computing the global error at the end of the run. Because the step sizes get redistributed between the last `post_step` hook and the `post_run` hook, this is a bit more messy than anticipated, but it is also tested.

@lisawim, now is your turn to point out the many places where I can improve ;)


lisawim:
Looks really good to me! As usual, it looks all very clean coded. But there is one thing I noticed: Due to consistency it makes sense to rename the file Lorentz.py to Lorenz.py ;) That's all!

pancetta:
Merged it. You could consider testing the MPI part or removing it (if you don't need it anyway).

Issue #262, opened 2023-01-20
Exclude code from coverage


pancetta:
We should consider excluding code from coverage reports, e.g. plotting, error handling (won't test that anyway), terribly slow parts (unless we find a way to test it nonetheless)...

One way to do that is to use `# pragma: no cover`, see https://coverage.readthedocs.io/en/7.0.5/excluding.html.

Thoughts?

tlunet:
I wonder if this won't induce more use of `# pragma: no cover` instead of adding test. And the indication of code coverage is still good in my opinion. Why not just having coverage tests indicated for each PR (with improvements/degradations) rather as condition for the PR to pass CI tests ?

tlunet:
On an other note, maybe if we switch in a near future to jupyter notebooks for project/tutorials description, instead of using one script that generates plots and figures, maybe that could solve the problem (considering that every jupyter would be tested by a simple re-run ...) ?

pancetta:
Well, we can always just ignore degradation.. but if the degradation comes from stuff like plotting, then I fail to see how to test that reasonably. 

Yet, if people just put in this pragma everywhere, nothing is gained, of course. But hey, no matter what we decide, this is what reviews are for..

tlunet:
Actually, excluding code from coverage seems to be a research subject itself, see this (interesting) article : https://homepages.dcc.ufmg.br/~andrehora/pub/2021-msr-test-coverage-exclusion.pdf

From a first quick look, seems that the `# pragma: no cover` approach is actually a good idea, if we already define somewhere what could be excluded and what should not be  (and finally, review to assess that :wink:)

brownbaerchen:
I like the statements. Jupyter is all well and good, but I do not want to rely on it. When I am making plots for a talk, for instance, I want nothing to do with Jupyter, but I still want the script to be in the repo. Also, now that you included markdown support with plots, I may choose that over Jupyter.

I tested these pragmas [here](http://parallel-in-time.org/pySDC/coverage/d_5893b65af8042ae2_Lorenz_py.html#t130).
Since individual plotting scripts are not part of the functionality that others care about I think we can safely ignore testing them with these statements.
"Serious" plotting scripts that generate plots for papers are after all executed by the pipeline.

I am not sure what to do about debugging statements. If I check that someone has put in the correct kind of parameters, I don't want to write a test for every way of triggering a debug statement, but the error messages may be broken, which others may care about.

So I am in favour of excluding plotting only functions but nothing else.

[...]

PR #265, opened 2023-01-22
Added coverage details on contribution guidelines

tlunet:
Following #262


pancetta:
In case you wonder why this had an effect on coverage: there's one project which uses 1000s of tests to generate statistics. Looks like this does not have a random seed, so every once in a while the tests behave slightly differently and a few lines are not visited.

tlunet:
Looks like fun :laughing: ... I was wondering indeed, thanks for the explanaitions

PR #266, opened 2023-01-23
Convergence controller for adapting the parameters in the collocation problem during runtime

brownbaerchen:
The goal is to do adaptivity with different collocation methods. We can get a sort of embedded error estimate by having two solutions of different order by using two different collocation methods. The advantage compared to using different numbers of iterations is that the impact of the preconditioner is eliminated. That allows to use weird stuff like diagonal preconditioners and so on.

Producing multiple solutions to different collocation problems is expensive, but we are hoping to accelerate convergence a great deal by interpolating between changes of collocation problems.

The first step is to change the collocation problem during runtime, which can be done with this convergence controller.
You setup the problem as usual, but then you can give this convergence controller an array of options that you want to loop through. For instance you can pass
```
description['convergence_controllers'] = {CollocationInexactness: {'num_nodes': [2, 3]}}
```
to setup an inexactness with two different collocation methods determined by whatever you passed to the sweeper parameters except for the number of nodes.

Before the first iteration and whenever convergence of a step is declared, the next sweeper in line will be activated. This means before the first iteration, the number of nodes will be changed to 2 regardless of what you supplied in the sweeper parameters after however many iterations are needed for convergence, a new collocation problem is set up with three nodes and iterated to convergence.
During all collocation changes, the solution is interpolated to the new problem using Thibaut's magic.

For the adaptivity, we want to solve all collocation problems exactly. When the concern is gaining efficiency, however this makes no sense, so you can also pass `restol = [1e-4, 1e-9]` for instance, which will be passed to the level.

@lisawim feel free to complain if you want, but don't feel obligated if you have better things to do such as writing conference papers ;)


pySDC/implementations/convergence_controller_classes/collocation_inexactness.py
```diff
@@ -0,0 +1,161 @@
+import numpy as np
+
+from pySDC.core.Lagrange import LagrangeApproximation
+from pySDC.core.ConvergenceController import ConvergenceController, Status
+from pySDC.core.Collocation import CollBase
+
+
+class CollocationInexactness(ConvergenceController):
```
pancetta, line 8: Why is that called `Inexactness`?

lisawim:
Nothing is more important than giving my complaint partner complaints! ;) In my view, it is really interesting, but somewhat confusing for the first time.. As I understand correctly, you want to do some iterations using a sweeper with 2 iterations, and after a few iterations the inexactness controller changes the number of nodes to 3 and do few iterations with this. Then you compute the estimated embedded error to have base for adaptivity. But what is the difference to using only one collocation method, i.e., only using one type of nodes? When you use description['convergence_controllers'] = {CollocationInexactness: {'num_nodes': [2, 3]}}, why one cannot use only 3 nodes. Using this you have higher computational costs. Does that really give you an improvement regarding the accuracy of the solution?

brownbaerchen:
> So, how do you deal with parameters that get overwritten by the convergence controller? Ideally, one does not have to provide these values in the first place, right? Less ideal, but more explicit: maybe issue a warning?

In principle, you are right that it is confusing. However, we need to set everything up before the convergence becomes active, so we need some parameters for the sweeper during instantiation of the levels.

What I can do is add parameters to the description when the convergence controller is setup. I believe I can find a place before the levels are instantiated. There, I could check if conflicting parameters are already present and raise a warning.

But neither is particularly good. I think I prefer not to change the description because that might decrease reproducibility in some cases. For instance if you store the description after the run, it will not be the same as before. With this convergence controller it should not cause any problems, it would just potentially trigger a warning.

What do you think? Should we allow any changes to the description after passing it to the controller and make this less confusing or should we risk confusion here for sake of possibly preventing confusion elsewhere? :D

pySDC/implementations/convergence_controller_classes/collocation_inexactness.py
brownbaerchen: I think the most intuitive use for this convergence controller by itself is to gradually refine the collocation problem. I am not clear on the exact definition of inexactness, but this is what I picture: To solve an adjacent, less accurate problem during early iterations and refine to the full problem later on.
Of course you can also do the opposite here or do iterations with less resolution but the same computational cost.
Do you have suggestions for a better name?

pancetta:
There are quite a few situations where changing the description on the fly would actually help (e.g. when repeating a test with different parameters). So far, you need to instantiate a new controller to get a clean start, but it would be easier if we could simply change the description. Yet, in this case one has to be careful that there are no remainders from past runs (e.g. in the stats object). So, carefully done this could help!

pySDC/implementations/convergence_controller_classes/collocation_inexactness.py
pancetta: I guess it's rather refinement and coarsening, but since it it both people call this adaptivity. But I'm not sure if that's not even more confusing.

pancetta: How about `AdaptiveCollocation`?

brownbaerchen:
> Nothing is more important than giving my complaint partner complaints! ;) In my view, it is really interesting, but somewhat confusing for the first time.. As I understand correctly, you want to do some iterations using a sweeper with 2 iterations, and after a few iterations the inexactness controller changes the number of nodes to 3 and do few iterations with this. Then you compute the estimated embedded error to have base for adaptivity. But what is the difference to using only one collocation method, i.e., only using one type of nodes? When you use description['convergence_controllers'] = {CollocationInexactness: {'num_nodes': [2, 3]}}, why one cannot use only 3 nodes. Using this you have higher computational costs. Does that really give you an improvement regarding the accuracy of the solution?

Thanks for taking the time to look at this!

The goal will be to obtain two solutions to the same initial value problem with different order of accuracy. That way we know one solution is significantly more accurate than the other and subtracting them gives a reasonable estimate of the local error of the lower order method.
So far we used the fact that the order of SDC rises by one with each iteration up to the order of the underlying quadrature.

However, this is only really true for implicit Euler sweeps! If we do something else like LU, chances are we still gain an order with each iteration, but I don't think we are guaranteed to. We can convince ourselves that this is the case for our setup and use adaptivity as we did so far.
But SDC is not really parallel-in-time by itself, so our supervisors want us to use ever crazier preconditioners that are diagonal to make SDC parallel across the method. These preconditioners are total which craft and their convergence properties are probably hard to define. 

So we are maybe not guaranteed that we gain one order of accuracy in every sweep with all preconditioners, but what we do know is that once we solved the collocation problem exactly, the order of accuracy of the scheme is equal to the order of the quadrature, no matter how we arrived there.
This way we can make sure that we satisfy the assumptions that went into the step size update equation and so on.
Solving the collocation problem exactly is expensive and switching the quadrature method is also expensive it seems, but nevertheless, we can try this out to get some combination of adaptivity and parallel-across-the-method SDC, which is mathematically justified.
Will it be way too expensive to be useful? Yes, maybe, probably. But do we want to play with this? Also yes, definitely, for sure!

brownbaerchen:
> There are quite a few situations where changing the description on the fly would actually help (e.g. when repeating a test with different parameters). So far, you need to instantiate a new controller to get a clean start, but it would be easier if we could simply change the description. Yet, in this case one has to be careful that there are no remainders from past runs (e.g. in the stats object). So, carefully done this could help!

What's wrong with instantiating a new controller? I think that is the easiest way to ensure you are doing a clean start when doing multiple runs of the same problem. I think if we want to use the same controller twice, we are asking for bugs in the resulting statistics :D

pancetta:
> > There are quite a few situations where changing the description on the fly would actually help (e.g. when repeating a test with different parameters). So far, you need to instantiate a new controller to get a clean start, but it would be easier if we could simply change the description. Yet, in this case one has to be careful that there are no remainders from past runs (e.g. in the stats object). So, carefully done this could help!
> 
> What's wrong with instantiating a new controller? I think that is the easiest way to ensure you are doing a clean start when doing multiple runs of the same problem. I think if we want to use the same controller twice, we are asking for bugs in the resulting statistics :D

Yeah, agreed. That's why changing anything on the fly is discouraged at the moment. But sometimes instantiating a controller can be costly (multi-level, FEM, ...). I guess I'd rather pay these costs than running into troubles with left-overs.

brownbaerchen:
I added some documentation which should clear confusion without having to alter the description. Because I made sure the output is really comprehensive when setting `logger_level = 15`, I think it should be fine. Here is an example of the produced text output when using the parameters
```
params = {
    'num_nodes': [4, 3, 2, 1],
}
```
for the convergence controller:
```
...
controller - INFO: Setup overview (--> user-defined, -> dependency) -- END                                                              

AdaptiveCollocation - Level 15: Process  0 on time 0.000000 - Switching to collocation 1 of 4:                                          
    quad_type: RADAU-RIGHT                                          
--> num_nodes: 4                                                    
    QI: IE                                                          
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  1 -- Sweep:  1 -- residual: 2.25802094e-02  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  2 -- Sweep:  1 -- residual: 1.43562682e-03  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  3 -- Sweep:  1 -- residual: 9.48776849e-05  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  4 -- Sweep:  1 -- residual: 6.46493715e-06  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  5 -- Sweep:  1 -- residual: 4.55301737e-07  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  6 -- Sweep:  1 -- residual: 3.21941697e-08  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  7 -- Sweep:  1 -- residual: 2.29171532e-09  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  8 -- Sweep:  1 -- residual: 1.64530028e-10  
AdaptiveCollocation - Level 15: Process  0 on time 0.000000 - Switching to collocation 2 of 4:                                          
    quad_type: RADAU-RIGHT                                          
--> num_nodes: 3                                                    
    QI: IE                                                          
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration:  9 -- Sweep:  1 -- residual: 4.00499261e-06  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 10 -- Sweep:  1 -- residual: 2.51483598e-07  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 11 -- Sweep:  1 -- residual: 1.49890329e-08  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 12 -- Sweep:  1 -- residual: 1.70295000e-09  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 13 -- Sweep:  1 -- residual: 1.89544047e-10  
AdaptiveCollocation - Level 15: Process  0 on time 0.000000 - Switching to collocation 3 of 4:                                          
    quad_type: RADAU-RIGHT                                          
--> num_nodes: 2                                                    
    QI: IE                                                          
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 14 -- Sweep:  1 -- residual: 2.46861879e-04  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 15 -- Sweep:  1 -- residual: 3.10622124e-05  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 16 -- Sweep:  1 -- residual: 3.90857854e-06  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 17 -- Sweep:  1 -- residual: 4.91817470e-07  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 18 -- Sweep:  1 -- residual: 6.18853254e-08  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 19 -- Sweep:  1 -- residual: 7.78700197e-09  
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 20 -- Sweep:  1 -- residual: 9.79831549e-10  
AdaptiveCollocation - Level 15: Process  0 on time 0.000000 - Switching to collocation 4 of 4:                                          
    quad_type: RADAU-RIGHT                                          
--> num_nodes: 1                                                    
    QI: IE                                                          
hooks - INFO: Process  0 on time 0.000000 at stage         IT_FINE: Level: 0 -- Iteration: 21 -- Sweep:  1 -- residual: 2.58681965e-14  
```

pySDC/projects/Resilience/collocation_adaptivity.py
```diff
@@ -0,0 +1,296 @@
...
+    """
+    Compute the maximal order achievable by a given collocation method
+    """
+    pass
+
+
+# define a few hooks
+class LogSweeperParams(hooks):
+    """
+    Log the sweeper parameters after every iteration to check if the collocation inexactness is doing what it's supposed
```
pancetta, line 63: collocation inexactness -> adaptive collocation

pySDC/projects/Resilience/collocation_adaptivity.py
```diff
@@ -0,0 +1,296 @@
...
+            level=L.level_index,
+            iter=step.status.iter,
+            sweep=L.status.sweep,
+            type='e_local_post_iter',
+            value=abs(L.prob.u_exact(t=L.time + L.dt) - L.uend),
+        )
+
+
+# plotting functions
+def compare_collocation_inexactness(prob):
```
pancetta, line 131: see above

pancetta: and below in the doc string

pySDC/tests/test_projects/test_resilience/test_collocation_inexactness.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+
+@pytest.mark.base
+def test_main():
+    from pySDC.projects.Resilience.collocation_adaptivity import main
+
+    main()
```
pancetta, line 8: Rename test please, see above

pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
```diff
@@ -0,0 +1,196 @@
...
+        'num_nodes': [2, 3],
+    }
+    ```
+    will use collocation methods like you passed to the `sweeper_params` in the `description` object, but will change
+    the number of nodes to 2 before the first iteration and to 3 as soon as the 2-node collocation problem is converged.
+    This will override whatever you set for the number of nodes in the `sweeper_params`, but you still need to set
+    something to allow instantiation of the levels before this convergence controller becomes active.
+    Make sure all lists you supply here have the same length.
+
+    Feel free to set `logger_level = 15` in the controller parameters to get comprehensive text output on what exactly
```
pancetta, line 30: So, no output with `loglevel > 15`? Maybe add something for `20` (e.g. only the first line of output from `15`?)

pySDC/projects/Resilience/collocation_adaptivity.py
```diff
@@ -0,0 +1,296 @@
...
+            level=L.level_index,
+            iter=step.status.iter,
+            sweep=L.status.sweep,
+            type='coll_order',
+            value=L.sweep.coll.order,
+        )
+
+
+class LogLocalErrorPostIter(hooks):
+    """
```
pancetta, line 101: Sounds like a hook class which could be used for more examples? Make this generic?

pySDC/implementations/hooks/log_errors.py
```diff
@@ -24,14 +34,47 @@ def post_step(self, step, level_number):
...
+class LogGlobalErrorPostStep(LogError):
+    def post_step(self, step, level_number):
+        super().post_step(step, level_number)
+        self.log_global_error(step, level_number, '_post_step')
+
 
 class LogGlobalErrorPostRun(hooks):
     """
     Compute the global error once after the run is finished.
+    Because of some timing issues, we cannot inherit from the `LogError` class here.
```
pancetta, line 77: Huh? What's the issue here?

brownbaerchen: The problem is that the convergence controllers which change the step size are called after the `post_step` hooks but before the `post_run` hook. That gives the convergence controllers a chance to change the step size before recording the error at `t=L.time + L.dt` in the `post_run` hook, which will then be a different time than belongs to the solution, which was recorded at `t=L.time + L.dt` as well, but `L.dt` had a different value.
To take care of this, I record the time at which the solution was recorded in the `LogGlobalErrorPostRun` hook. I can make the evaluation time a parameter in the function that computes the error, but since there is not a huge amount of abstraction, I thought I just don't abstract here.

pancetta: Gives me headaches.. but ok.

brownbaerchen: It was a weird bug that surprised me a little, but it is just a result of abstraction of the convergence controllers in the end. Why should they care what the controller decides to do next?
That is why the Lorenz test checks if the hook does what it's supposed to. I originally wanted to remove the computation from there and only use the hook, but then the bug showed itself ;)

pancetta: All hail automated tests!

brownbaerchen:
If there are no further complaints, I will merge this after lunch because I am working on other stuff in the same files and I don't want more merge conflicts than necessary...

PR #267, opened 2023-01-25
Resilience statistics for Lorenz attractor problem

brownbaerchen:
For the upcoming paper I thought it's interesting to look at a chaotic problem such as the Lorenz attractor in terms of resilience. I made a file discussing what I found, although it's largely the same as with the van der Pol equation, actually.

I changed a few more things that I needed to generate the statistics.
 - When the maximum number of restarts is reached, instead of just moving on to the next step, the new default behaviour is to raise a ConvergenceError and crash the code.
 - You can now add convergence controllers manually if you want to overwrite the parameters supplied to them when they where automatically added. I test this in the `vdp.py` file when I check the step size limiter. I supply `np.inf` as the minimum step size to adaptivity, which will add the step size limiter convergence controller, but since I add it manually with a more reasonable minimum step size, it works anyways. This let me remove the `max_restarts` option from the description.
 - New function for the resilience strategies to filter faults that can't be fixed
 - Parameters for fault insertion for Lorenz problem

Because I changed the default behaviour for dealing with multiple restarts, I had to mess with one of @lisawim's files very slightly. I hope we don't get merge conflicts because I believe you are working on the same files, but it should be easy to fix anyways.


pySDC/core/ConvergenceController.py
```diff
@@ -66,6 +66,9 @@ def log(self, msg, S, level=15, **kwargs):
     def setup(self, controller, params, description):
         """
         Setup various variables that only need to be set once in the beginning.
+        If the convergence controller is added automatically, you can give it params by adding it manually.
```
pancetta, line 69: I... am not sure I get what you're trying to say here.

pySDC/core/ConvergenceController.py
```diff
@@ -75,7 +78,8 @@ def setup(self, controller, params, description):
         Returns:
             (dict): The updated params dictionary after setup
         """
-        return params
+        # allow to change parameters by adding the convergence controller manually
+        return {**params, **description.get('convergence_controllers', {}).get(type(self), {})}
```
pancetta, line 82: Why is that necessary?

pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
```diff
@@ -78,7 +78,7 @@ def setup(self, controller, params, description):
 
                 defaults['num_colls'] = max([defaults['num_colls'], len(params[key])])
 
-        return {**defaults, **params}
+        return {**defaults, **super().setup(controller, params, description, **kwargs)}
```
pancetta, line 81: This is rather ugly and seems not very explicit. Why do you need this?

brownbaerchen:
The problem this aims to solve is the following: Many convergence controllers are added automatically with parameters that are defined by whatever added them. But what if I want other parameters than the automatically added version gets?
In the current master branch I am just screwed in that case...
Here, when adding the convergence controller, the core modules setup function looks for parameters of manually added versions of the same convergence controller. These will then override the parameters given by default, but the convergence controller is still added only once.
So far, there are honestly no cases where this is needed, but I want to stay general. It might become useful at some point.
For instance when it comes to spreading step sizes. That convergence controller is loaded automatically by the basic restarting convergence controller and if we want to supply any specific parameters to that, this is the most straightforward way, I believe.
We can do sth similar by adding the convergence controller before it is added automatically with custom parameters in the current implementation, actually. But that is a mess waiting for bugs to happen.

pancetta:
> The problem this aims to solve is the following: Many convergence controllers are added automatically with parameters that are defined by whatever added them. But what if I want other parameters than the automatically added version gets? In the current master branch I am just screwed in that case... Here, when adding the convergence controller, the core modules setup function looks for parameters of manually added versions of the same convergence controller. These will then override the parameters given by default, but the convergence controller is still added only once. So far, there are honestly no cases where this is needed, but I want to stay general. It might become useful at some point. For instance when it comes to spreading step sizes. That convergence controller is loaded automatically by the basic restarting convergence controller and if we want to supply any specific parameters to that, this is the most straightforward way, I believe. We can do sth similar by adding the convergence controller before it is added automatically with custom parameters in the current implementation, actually. But that is a mess waiting for bugs to happen.

So, users can add an existing/automatically added convergence controller again, providing their own set of parameters, and this will then overwrite the automatically added one, yes?

brownbaerchen:
Yes, there is a test where two parameters are added automatically and one of them is overwritten by adding the convergence controller again manually. I added a test to make sure that only one instance of this convergence controller is added. It's all in the `check_step_size_limiter` function in the `vdp.py` file.

pySDC/core/ConvergenceController.py
pancetta: Can you make this more clear in this docstring, please?

PR #268, opened 2023-01-26
Minor modification on PETSc datatype

tlunet:
`__new__` is a class method => does not need to instantiate a `PETSc.VEC` object to be used.

This PR may improve performance when using `petsc_vec` datatypes (didn't test it tho :sweat_smile: ...)


pancetta:
Wow, good catch!!!

PR #271, opened 2023-01-27
Application of SDC to fully-implicit DAE: sweeper, example problems, performance tests.

kho19:


pancetta:
Looks good to me so far, but test coverage could be better. Please add tests for the other problem classes/cases and hooks. You can either do this as part of your project or directly in the test directory.

kho19:
Added tests achieving 100% coverage. 
@brownbaerchen mentioned that the .npy files storing the reference solutions may be too large. Any comments on this?

PR #273, opened 2023-02-02
Bugfix for SE

lisawim:
I found some bug in the SE that I have fixed now (hopefully). I observed for some specific case that after jumping over the switch, the time step is not setting back to the initial dt in the switch estimator's post_step_processing function (only using SE in this specific case). The cause was the reset of the status variables. Before post_step_processing is reached, the status variables have already been reset to None. I assume that post_step_processing is already called after a successfully computed step, but status variables will be called after each step. For this reason, I moved switch_detected, t_switch back to the default dict, so that I can set it back to None after the next successful step. I also updated the hardcode references.

Probably @brownbaerchen has some idea for a cleaner alternative.

### DIFF
--- a/pySDC/core/ConvergenceController.py
+++ b/pySDC/core/ConvergenceController.py
@@ -21,7 +21,6 @@ class Status(FrozenClass):
     """
 
     def __init__(self, status_variabes):
-
         [setattr(self, key, None) for key in status_variabes]
 
         self._freeze()
@@ -415,7 +414,6 @@ class ConvergenceController(object):
 
         # check if we have arrived at the end of the path to the variable
         if len(where) == 0:
-
             variable_exitsts = name in place.__dict__.keys()
             # check if the variable already exists and raise an error in case we are about to introduce a bug
             if not allow_overwrite and variable_exitsts:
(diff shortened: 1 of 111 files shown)


lisawim:
I'm a bit confused about these several black errors.. Locally I hadn't any ones.

pancetta:
> I'm a bit confused about these several black errors.. Locally I hadn't any ones.

Tried an update of black? Maybe a new version is unhappy about stuff old versions let through?

lisawim:
As discussed with @brownbaerchen we did set the variables back to Status and did some adaptions in post_step_processing to make it work.

lisawim:
Funny, my command line added the git add command to my commit..

pancetta:
Welcome to the `git` jungle... you have 111 changes in this PR.

brownbaerchen:
> Welcome to the `git` jungle... you have 111 changes in this PR.

Thanks to the black update, I guess......

pancetta:
Merge `master` and you should be fine.

lisawim:
You wrote I should do update my black. I did. I did blackening for 105 files, on Friday. Now you also did do this today. This is the reason why I have 111 files in this PR..

brownbaerchen:
> Merge `master` and you should be fine.

We found that GitHub unfortunately doesn't update the reference in the past, right? As is the case already since Lisa did the same blackening as you. Did someone find a solution to this other than closing and reopening the PR?

pancetta:
Worked!

brownbaerchen:
> Worked!

Maybe we were doing some strange BS earlier then :D

PR #275, opened 2023-02-10
Adding toy problem to model superconductor with temperature leak

brownbaerchen:
This problem is interesting since the temperature of the domain at some point rises suddenly after a period of staying virtually constant, which is a good showcase for adaptivity.

The equation was proposed by @eschnaubelt during the Time-X Apps Hackathon as a toy version of something he works with in a "serious" code. 

In some cases it is required to impose an upper bound on the step size when using adaptivity with this problem. Otherwise the transition between low temperature and rapid heating is sometimes not resolved in time.
There is both an IMEX version and a fully implicit version, with the fully implicit version seeming more robust w.r.t. this issue.

Since it's the first time for me with Neumann boundary conditions in finite differences, maybe someone can check that I implemented this correctly. I followed the finite difference book by Leveque, but it does not go into great detail on boundary conditions, so I am a bit insecure about this implementation. If anyone knows a good test that I can implement to check the generic boundary conditions, that would be great!
Keep in mind that the git diff does not necessarily show that we assert the maximum order of the finite difference discretisation to be no more than two in case non-periodic boundary conditions are used. So for now it's fine that the Neumann boundaries are hardcoded to first order, I guess.

In terms of testing, I only check that the IMEX and fully implicit versions give somewhat of the same result. This is only really a check on the Newton solver, however.. I also confirmed by eye that the solution looks as expected. Since the thermal conductivity is really high, the temperature profile is almost constant across the spatial domain, even though there is heating in the middle. This behaviour is maintained until the end of the simulation.
See below the maximum temperature and the step size over time for one set of parameters.
Figure_2

For now, this is only 1D, but it should not be too difficult to extend this to more dimensions if we think this is useful. I think this is a nice problem to showcase adaptivity with because it models a real application, even though it is still pretty simple.

@eschnaubelt Let me know what you think and if we can cooperate a bit further on similar problems, maybe! We are currently working on a publication on adaptive step sizes in spectral deferred corrections and I think this problem or possibly slightly more complicated versions of it would fit very nicely in there.

### DIFF
--- a/pySDC/implementations/problem_classes/LeakySuperconductor.py
+++ b/pySDC/implementations/problem_classes/LeakySuperconductor.py
@@ -49,6 +49,10 @@ class LeakySuperconductor(ptype):
             'lintol': 1e-8,
         }
 
+        # optional parameters for solving the linear problems within the Newton solver iteratively
+        defaults['liniter_within_newton'] = problem_params.get('liniter_within_newton', defaults['liniter'])
+        defaults['lintol_within_newton'] = problem_params.get('lintol_within_newton', defaults['lintol'])
+
         for key in problem_params.keys():
             if key not in defaults.keys():
                 raise ParameterError(
@@ -137,7 +141,7 @@ class LeakySuperconductor(ptype):
 
     def solve_system(self, rhs, factor, u0, t):
         """
-        Simple linear solver for (I-factor*A)u = rhs
+        Simple Newton solver for (I-factor*f)(u) = rhs
 
         Args:
             rhs (dtype_f): right-hand side for the linear system
@@ -190,7 +194,17 @@ class LeakySuperconductor(ptype):
             J = self.Id - factor * (self.A + get_non_linear_Jacobian(u))
 
             # solve the linear system
-            delta = np.linalg.solve(J, G)
+            if self.params.direct_solver:
+                delta = np.linalg.solve(J, G)
+            else:
+                delta = cg(
+                    J,
+                    G,
+                    x0=G * 0,
+                    tol=self.params.lintol_within_newton,
+                    maxiter=self.params.liniter_within_newton,
+                    atol=0,
+                )[0].reshape(self.params.nvars)
 
             # update solution
             u = u - delta
@@ -223,7 +237,7 @@ class LeakySuperconductor(ptype):
                 Returns:
                     (numpy.1darray): RHS
                 """
-                return self.eval_f(u, t)
+                return self.eval_f(u.reshape(self.init[0]), t).flatten()
 
             me[:] = self.generate_scipy_reference_solution(eval_rhs, t, u_init, t_init)
         return me
@@ -253,7 +267,7 @@ class LeakySuperconductorIMEX(LeakySuperconductor):
 
     def solve_system(self, rhs, factor, u0, t):
         """
-        Simple linear solver for (I-factor*A)u = rhs
+        Simple linear solver for (I-factor*f_expl)(u) = rhs
 
         Args:
             rhs (dtype_f): right-hand side for the linear system
@@ -278,4 +292,38 @@ class LeakySuperconductorIMEX(LeakySuperconductor):
                 maxiter=self.params.liniter,
                 atol=0,
             )[0].reshape(self.params.nvars)
+        return me
+
+    def u_exact(self, t, u_init=None, t_init=None):
+        """
+        Routine to compute the exact solution at time t
+
+        Args:
+            t (float): current time
+
+        Returns:
+            dtype_u: exact solution
+        """
+        me = self.dtype_u(self.init, val=0.0)
+
+        if t == 0:
+            me[:] = super().u_exact(t, u_init, t_init)
+
+        if t > 0:
+
+            def eval_rhs(t, u):
+                """
+                Function to pass to `scipy.solve_ivp` to evaluate the full RHS
+
+                Args:
+                    t (float): Current time
+                    u (numpy.1darray): Current solution
+
+                Returns:
+                    (numpy.1darray): RHS
+                """
+                f = self.eval_f(u.reshape(self.init[0]), t)
+                return (f.impl + f.expl).flatten()
+
+            me[:] = self.generate_scipy_reference_solution(eval_rhs, t, u_init, t_init)
         return me
--- a/pySDC/implementations/problem_classes/LeakySuperconductor.py
+++ b/pySDC/implementations/problem_classes/LeakySuperconductor.py
@@ -142,7 +142,7 @@ class LeakySuperconductor(ptype):
         Simple Newton solver for (I-factor*f)(u) = rhs
 
         Args:
-            rhs (dtype_f): right-hand side for the linear system
+            rhs (dtype_f): right-hand side
             factor (float): abbrev. for the local stepsize (or any other factor required)
             u0 (dtype_u): initial guess for the iterative solver
             t (float): current time (e.g. for time-dependent BCs)
--- a/pySDC/implementations/problem_classes/LeakySuperconductor.py
+++ b/pySDC/implementations/problem_classes/LeakySuperconductor.py
@@ -1,0 +1,281 @@
+import numpy as np
+import scipy.sparse as sp
+from scipy.sparse.linalg import cg, spsolve
+
+from pySDC.core.Errors import ParameterError, ProblemError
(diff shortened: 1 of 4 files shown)


pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -0,0 +1,281 @@
...
+        Returns:
+            dtype_f: The right hand side
+        """
+        f = self.dtype_f(self.init)
+        f[:] = (self.A.dot(u.flatten()).reshape(self.params.nvars) + self.eval_f_non_linear(u, t)) / self.params.Cv
+        return f
+
+    def solve_system(self, rhs, factor, u0, t):
+        """
+        Simple linear solver for (I-factor*A)u = rhs
```
pancetta, line 140: Yeah, no.

pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -0,0 +1,281 @@
...
+
+        Args:
+            rhs (dtype_f): right-hand side for the linear system
```
pancetta, line 143: see above

pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -0,0 +1,281 @@
...
+
+            res = np.linalg.norm(G, np.inf)
+            if res <= self.params.lintol or np.isnan(res):
+                break
+
+            # assemble Jacobian J of G
+            J = self.Id - factor * (self.A + get_non_linear_Jacobian(u))
+
+            # solve the linear system
+            delta = np.linalg.solve(J, G)
```
pancetta, line 193: Could make this iterative, too, using the flag you already have..

pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -0,0 +1,281 @@
...
+        Args:
+            t (float): current time
+
+        Returns:
+            dtype_u: exact solution
+        """
+
+        me = self.dtype_u(self.init, val=0)
+
+        if t > 0:
```
pancetta, line 213: Could you also test this?

pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -0,0 +1,281 @@
...
+        Returns:
+            dtype_u: solution as mesh
+        """
+
+        me = self.dtype_u(self.init)
+
+        if self.params.direct_solver:
+            me[:] = spsolve(self.Id - factor * self.A, rhs.flatten()).reshape(self.params.nvars)
+        else:
+            me[:] = cg(
```
pancetta, line 273: Want to test this?

pySDC/projects/Resilience/leaky_superconductor.py
```diff
@@ -0,0 +1,195 @@
...
+
+class live_plot(hooks):  # pragma no cover
+    """
+    This hook plots the solution and the non-linear part of the right hand side after every step. Keep in mind that using adaptivity will result in restarts, which is not marked in these plots. Prepare to see the temperature profile jumping back again after a restart.
+    """
+
+    def pre_run(self, step, level_number):
+        self.fig, self.axs = plt.subplots(1, 2, figsize=(10, 4))
+
+    def post_step(self, step, level_number):
```
pancetta, line 24: Exclude from coverage via pragma

pySDC/projects/Resilience/leaky_superconductor.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # call main function to get things done...
+    try:
+        uend, stats = controller.run(u0=uinit, t0=t0, Tend=Tend)
+    except ConvergenceError:
+        stats = controller.return_stats()
+    return stats, controller, Tend
+
+
+def plot_solution(stats, controller):  # pragma no cover
```
pancetta, line 132: Why is the pragma not working here?

pySDC/projects/Resilience/leaky_superconductor.py
```diff
@@ -0,0 +1,195 @@
...
+        custom_description['convergence_controllers'] = {Adaptivity: {'e_tol': 1e-6, 'dt_max': dt_max[imex]}}
+        stats, controller, _ = run_leaky_superconductor(
+            custom_description=custom_description,
+            custom_controller_params=custom_controller_params,
+            imex=imex,
+            Tend=5e2,
+        )
+
+        res[imex] = get_sorted(stats, type='u')[-1][1]
+        if plotting:
```
pancetta, line 184: pragma

brownbaerchen:
Hmmmmmmmmm conda says there is no update to black and I ran black locally before committing, of course. Why is black doing sth else here than locally?

pancetta:
> Hmmmmmmmmm conda says there is no update to black and I ran black locally before committing, of course. Why is black doing sth else here than locally?

What's you version of black, locally? The CI has `23.1.0`.

brownbaerchen:
```$ black --version
black, 23.1.0 (compiled: yes)
Python (CPython) 3.10.9```

pancetta:
Up-to-date with master?

pySDC/implementations/problem_classes/LeakySuperconductor.py
brownbaerchen: I used zeros as initial guess because I am computing a correction here, which should be small. Does that make sense? There is no information about the spatial profile in zeros left, but I couldn't think of anything better. Suggestions?

pancetta: No, makes sense. You want the delta to be zero at some point, so this is meaningful.

pancetta:
`test_resilience/test_order.py::test_main` and, well, `test_resilience/test_leaky_superconductor.py::test_scipy_reference_solution` take a bit long, no?

eschnaubelt:
Nice work! Thanks (again) for the nice interaction at the Hackathon! You results are indeed a nice approximation of the dynamics of more complex real-world problems. I believe that a 1D problem should be enough to highlight the need for adaptive time steps while still keeping short time-to-prototype. 

If you wanted to increase the effect even further, you could consider temperature-dependent material relations for your thermal conductivity kappa and heat capacity CV. Furthermore, the heat source could be modeled as an exponential rather than a linear function to make the problem "more non-linear". In the end, the level of complexity depends on what you'd like to show, e.g., for the publication. Sky's the limit. 

I'd surely appreciate further collaboration! I'm available for discussion if needed/helpful, feel free to contact me. I'll also report to my supervisors in the coming days to identify our future directions. I believe there's potential for PinT to accelerate quench simulations.

tlunet:
> `test_resilience/test_order.py::test_main` and, well, `test_resilience/test_leaky_superconductor.py::test_scipy_reference_solution` take a bit long, no?

I don't think the `scipy_reference_solution` should be tested with every problem available ... at least if you use the default implementation that you added in the `Problem` class @brownbaerchen ?

brownbaerchen:
> > `test_resilience/test_order.py::test_main` and, well, `test_resilience/test_leaky_superconductor.py::test_scipy_reference_solution` take a bit long, no?
> 
> I don't think the `scipy_reference_solution` should be tested with every problem available ... at least if you use the default implementation that you added in the `Problem` class @brownbaerchen ?

I agree, I added the test after Robert requested it, but I can ofc remove it again :D

pancetta:
Yes please, did not anticipate the extend of this test.

brownbaerchen:
> `test_resilience/test_order.py::test_main` and, well, `test_resilience/test_leaky_superconductor.py::test_scipy_reference_solution` take a bit long, no?

Where can you see how long the tests take? I always make sure they don't take too long locally, but I don't know how that translates to performance on GitHub..

pancetta:
After each call to `pytest` you can see the duration, e.g. [here](https://github.com/Parallel-in-Time/pySDC/actions/runs/4162349024/jobs/7201413925#step:4:288).

brownbaerchen:
I looked a bit closer and noticed the temperature did not enter the runaway stage in my tests with iterative solvers. So I increased their accuracy and now there were a lot of `nan` and the code would crash. I don't understand CG well enough to know what's going on, but I don't think we really need it. It was just a copy-paste artefact so I removed it.
I also removed the test for the scipy reference solution and added a test that the runaway process has started.

PR #276, opened 2023-02-14
Resilience in Schrödinger equation

brownbaerchen:
This PR adds some stuff for simulating faults in the Schrödinger equation and a file to generate plots that are supposed to go in some publication at some point.

I didn't want to test the function for running the Schrödinger equation because that doesn't contain much new stuff. I did add a test for the flipping of bits in complex numbers, which is also new in this PR.

Also, I changed the way I generate stats in parallel so I can do this on the cluster to get some statistics for PDEs in reasonable time.

Finally, I removed some stuff from the test to check the order of error estimates. Since they produce a plot that goes on the website, I couldn't remove too much stuff, but I am hoping to cut the time in half.

This PR should not change anything related to other peoples projects.


pancetta:
That's quite a hit in coverage. How much does running the example cost?

brownbaerchen:
> That's quite a hit in coverage. How much does running the example cost?

I guess it doesn't cost too much, but I don't really know what to test here. All I do is setup the problem with certain parameters.

pancetta:
Just run it and see that it does meaningful things in terms of iteration counts, etc.. You don't want to end up with a setup that crashes because.. reasons.

pancetta:
Looks like the faulty case is not tested, correct?

brownbaerchen:
I still don't know why this flakeheaven complaint does not show up on my local machine. I even tried setting up a new conda environment to no avail. Anyways, the "fix" changes the behaviour of the code. Now the except behaviour is entered if there is either `dill.PicklingError` or `TypeError`, not both as before. To me it looked like this is actually desired, if not please complain. Anyways, thanks to [these guys](https://github.com/PyCQA/flake8-bugbear/pull/346) the previous line is no longer allowed. Another solution would be, of course, to exclude `B030` from flakeheaven. Your choice.

PR #277, opened 2023-02-17
Added A-stable DIRK34 method

brownbaerchen:
Taken from [here](https://doi.org/10.1007/BF01934920). The tests make sure that the embedded error estimate obtained with this method is order 4 and that the method is A-stable (within the patch of the complex plane that is tested).


pancetta:
I see, you found another [Balrog](https://www.youtube.com/watch?v=PfhZB7rQ7iA)..

PR #278, opened 2023-02-21
Bugfixes in Quench problem

brownbaerchen:
This PR includes a few things.

Most notably bug fixes in the Quench problem. I forgot a diffusion coefficient in the solvers for both IMEX and fully implicit. Interestingly, the solution doesn't really change after the fix. I guess this is another case where you can do pretty much whatever you want in the preconditioner, as long as you compute the residual accurately. Since I remembered the coefficient in the rhs evaluations, this, at least, was correct.
At the end of the day it is not dissimilar to the IMEX idea: I was solving implicitly a problem with much more diffusion (I think) and used that to precondition the full "normal" system. Not that this is something reasonable to do. I just think the bug is actually sort of interesting.

Also, I tried some resilience with this, but it's a bit of a mess. Since the problem is so stiff, it is too expensive to generate a scipy reference solution for this during runtime. I can produce a reference solution myself, but that is also a bit of a mess. I noticed that faults can impact the solution a little bit, but usually not too much. What I was hoping for was that flipping a bit in the exponent would inject huge amounts of energy into the system, which triggers the runaway heating. However, it seems the energy is just dissipated because the FD discretisation is not good enough. (I am just guessing here, at the end of the day I have no idea...)
So while the solution is impacted a bit, faults don't do anything particularly interesting in the cases I tried. So I just imposed an absolute tolerance on the final temperature that I want to have at the end of the simulation, which leads to most faults being recovered. Given the huge amount of diffusion in this problem, this is perhaps not unexpected. Maybe I should try with an exponential term in the transition to runaway heating to see if this is more sensitive to faults.

Something that I implemented along the way and I thought I could just merge is checking for convergence by reaching an increment tolerance. I also added a test for that.

Also, I started on a function in the plot helper for generating figures for specific journals. At the moment maybe there are not too many papers written with pySDC, but at least the preset for the JSC beamer template I find useful. Basically, I want people to hardcode the LaTeX text width there such that you can easily generate figures with a certain size relative to the tex target. I think rather than everybody making their own plotting utilities maybe we can work together to use some tricks others have learned.

There is also a hook logging the embedded error estimate after each iteration, which I needed for the test of the check_convergence convergence controller. It's a bit hacky because the estimate is computed after the hook is called, but it should work fine and the user does not have to worry about this. 

Sorry for clustering many small messes into one large mess like this..

Final note on the quench problem: After the bugfix I need to limit the step size for the fully implicit version as well in order for it not to under-resolve the temperature jump. On the bright side, even if it is under-resolved, the runaway process is still captured, which is the most important thing. Also, I want to try SDC vs. the fourth order embedded DIRK scheme that I implemented the other day to see if SDC is actually more efficient than RK, of only compared to explicit RK, don't hope for good news on that front though, probably..


pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, problem_params, dtype_u=mesh, dtype_f=mesh):
             self.dx = 1.0 / (self.params.nvars + 1)
             xvalues = np.array([(i + 1) * self.dx for i in range(self.params.nvars)])
         elif self.params.bc == 'neumann-zero':
-            self.dx = 1.0 / (self.params.nvars - 1)
+            self.dx = 1.0 / (self.params.nvars + 1)
             xvalues = np.array([(i + 1) * self.dx for i in range(self.params.nvars)])
```
pancetta, line 75: You're sure these `xvalues` are correct? It looks like they run from `1/(N+1)` to `(N+1)/(N+1)` and I guess you're missing the `0` here.

pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -182,18 +184,22 @@ def get_non_linear_Jacobian(u):
             G = u - factor * self.eval_f(u, t) - rhs
 
             res = np.linalg.norm(G, np.inf)
-            if res <= self.params.newton_tol or np.isnan(res):
+            if res <= self.params.newton_tol and _n > 0:
```
pancetta, line 187: Why the `_n`?

pancetta: I mean, why the `_` in the name?

pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -182,18 +184,22 @@ def get_non_linear_Jacobian(u):
...
                 break
 
             # assemble Jacobian J of G
-            J = self.Id - factor * (self.A + get_non_linear_Jacobian(u))
+            J = self.Id - factor * (self.A + get_non_linear_Jacobian(u)) / self.params.Cv
```
pancetta, line 191: I guess you should put that parameter into the `A` and the nonlinear Jacobian..

pySDC/implementations/problem_classes/LeakySuperconductor.py
```diff
@@ -265,7 +271,7 @@ def solve_system(self, rhs, factor, u0, t):
         """
 
         me = self.dtype_u(self.init)
-        me[:] = spsolve(self.Id - factor * self.A, rhs.flatten()).reshape(self.params.nvars)
+        me[:] = spsolve(self.Id - factor * self.A / self.params.Cv, rhs.flatten()).reshape(self.params.nvars)
```
pancetta, line 274: see above

pySDC/projects/Resilience/fault_stats.py
```diff
@@ -212,14 +218,19 @@ def get_custom_description(self, problem, num_procs):
         elif problem == run_Schroedinger:
             e_tol = 4e-6
             dt_min = 1e-3
+        elif problem == run_leaky_superconductor:
```
pancetta, line 221: Looks like you're not testing the leaky superconductor, but changed the parameter in `fault_stats.py` to this (see below)?

pySDC/projects/Resilience/fault_stats.py
```diff
@@ -1659,20 +1732,22 @@ def bar_plot_thing(
 
 def main():
     stats_analyser = FaultStats(
-        prob=run_Schroedinger,
+        prob=run_leaky_superconductor,
```
pancetta, line 1735: here

brownbaerchen: With this script I actually make the statistics for resilience so the problem here is just whatever I last checked. There is a test checking that the overall script works, which generates stats for van der Pol. I am not sure how to test that parameters work. I can ofc. test the stats generation for all problems, but its actually quite expensive. I think testing the parameters for all problems would be excessive.

pySDC/implementations/problem_classes/LeakySuperconductor.py
brownbaerchen: Better now? We want to include both boundaries, a.k.a. 0 and 1, right?

pySDC/implementations/problem_classes/LeakySuperconductor.py
brownbaerchen: Flakeheaven makes you add underscores to iteration variables that don't appear inside the loop, but now it does appear in the loop, so the underscore should be removed, you're right!

PR #279, opened 2023-02-23
Counters for measuring work

brownbaerchen:
I put in a counter class which can be passed to scipy solvers like gmres to count iterations, but you can also use it to count all sorts of things like Newton iterations and right hand side evaluations. This is useful because it is much more finely grained than SDC iterations for measuring the cost of the scheme.
There is also a hook that logs the increment in all counters in every step.

I did it this way because the problems are not supposed to interface with the hooks, but you can access problem attributes from the hooks.

I also implemented GMRES into the fully implicit version of the quench problem. However, to make it converge, I had to apply a preconditioner which is the linear part of the problem I want to solve with the Newton solver a.k.a. the diffusion part. Since the IMEX problem only solves the linear part implicitly, I didn't find a suitable preconditioner to make GMRES or CG converge, so there is no iterative linear solver in the IMEX version.

### DIFF
--- a/pySDC/core/Problem.py
+++ b/pySDC/core/Problem.py
@@ -12,6 +12,18 @@ class _Pars(FrozenClass):
         self._freeze()
 
 
+class WorkCounter(object):
+    """
+    Class for counting iterations
+    """
+
+    def __init__(self):
+        self.niter = 0
+
+    def __call__(self, *args, **kwargs):
+        self.niter += 1
+
+
 class ptype(object):
     """
     Prototype class for problems, just defines the attributes essential to get started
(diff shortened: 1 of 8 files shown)


pancetta:
Good idea! The only issue I have is the naming: `niter` could be confused with SDC iterations, no? Rather `inner_niter` or so?

brownbaerchen:
> Good idea! The only issue I have is the naming: `niter` could be confused with SDC iterations, no? Rather `inner_niter` or so?

I actually don't think it's too confusing. After all you interface with this like 
```
# retrieve number of iterations in the counter
newton_iter = prob.work_counters['newton'].niter

# increase iteration count for Newton solver
prob.work_counters['newton']()
```
And in the hook class that logs the work in every step the variables are stored as `work_newton` etc.

Since the Newton solver can itself have inner iterations of a linear solver, I think we would make it more confusing by renaming `niter` to `inner_niter`.

PR #281, opened 2023-03-08
Interpolate after restarts

brownbaerchen:
New convergence controller which interpolates after a restart. There is a test that makes sure that the residual after solving the step for the second time is smaller than without interpolation and that that the interpolated value shows up the next step.

Also, since the non-MPI and MPI controllers deal with the steps differently (`controller.MS` and `controller.S`), I need some controller specific behaviour in the convergence controllers. Since I got tired of importing some controller and type-checking, I added a variable to both the controller as well as the convergence controllers that store if MPI is used or not.

### DIFF
--- a/pySDC/implementations/controller_classes/controller_MPI.py
+++ b/pySDC/implementations/controller_classes/controller_MPI.py
@@ -25,7 +25,7 @@ class controller_MPI(controller):
         """
 
         # call parent's initialization routine
-        super(controller_MPI, self).__init__(controller_params, description)
+        super(controller_MPI, self).__init__(controller_params, description, useMPI=True)
 
         # create single step per processor
         self.S = step(description)
--- a/pySDC/implementations/controller_classes/controller_MPI.py
+++ b/pySDC/implementations/controller_classes/controller_MPI.py
@@ -25,7 +25,7 @@ class controller_MPI(controller):
         """
 
         # call parent's initialization routine
-        super(controller_MPI, self).__init__(controller_params, description, useMPI=True)
+        super().__init__(controller_params, description, useMPI=True)
 
         # create single step per processor
         self.S = step(description)
--- a/pySDC/implementations/controller_classes/controller_nonMPI.py
+++ b/pySDC/implementations/controller_classes/controller_nonMPI.py
@@ -30,7 +30,7 @@ class controller_nonMPI(controller):
             raise ControllerError('predict flag is ignored, use predict_type instead')
 
         # call parent's initialization routine
-        super(controller_nonMPI, self).__init__(controller_params, description)
+        super(controller_nonMPI, self).__init__(controller_params, description, useMPI=False)
 
         self.MS = [stepclass.step(description)]
 
--- a/pySDC/implementations/controller_classes/controller_nonMPI.py
+++ b/pySDC/implementations/controller_classes/controller_nonMPI.py
@@ -30,7 +30,7 @@ class controller_nonMPI(controller):
             raise ControllerError('predict flag is ignored, use predict_type instead')
 
         # call parent's initialization routine
-        super(controller_nonMPI, self).__init__(controller_params, description, useMPI=False)
+        super().__init__(controller_params, description, useMPI=False)
 
         self.MS = [stepclass.step(description)]
 
--- a/pySDC/implementations/convergence_controller_classes/interpolate_between_restarts.py
+++ b/pySDC/implementations/convergence_controller_classes/interpolate_between_restarts.py
@@ -1,5 +1,5 @@
 import numpy as np
-from pySDC.core.ConvergenceController import ConvergenceController
+from pySDC.core.ConvergenceController import ConvergenceController, Status
 from pySDC.core.Lagrange import LagrangeApproximation
 from pySDC.core.Collocation import CollBase
 
@@ -26,7 +26,6 @@ class InterpolateBetweenRestarts(ConvergenceController):
         """
         defaults = {
             'control_order': 50,
-            'initial_guess': description['sweeper_params'].get('initial_guess', 'spread'),
         }
         return {**defaults, **super().setup(controller, params, description, **kwargs)}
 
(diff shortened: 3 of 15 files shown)


pancetta:
Nice reduction of imports!

pySDC/implementations/controller_classes/controller_MPI.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, controller_params, description, comm):
         """
 
         # call parent's initialization routine
-        super(controller_MPI, self).__init__(controller_params, description)
+        super(controller_MPI, self).__init__(controller_params, description, useMPI=True)
```
tlunet, line 28: Can probably already simplify to `super().__init__`

pySDC/implementations/controller_classes/controller_nonMPI.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, num_procs, controller_params, description):
             raise ControllerError('predict flag is ignored, use predict_type instead')
 
         # call parent's initialization routine
-        super(controller_nonMPI, self).__init__(controller_params, description)
+        super(controller_nonMPI, self).__init__(controller_params, description, useMPI=False)
```
tlunet, line 33: Same comment as before ...

pySDC/implementations/convergence_controller_classes/interpolate_between_restarts.py
```diff
@@ -0,0 +1,102 @@
...
+        Returns:
+            None
+        """
+        self.status = Status(['u_inter', 'f_inter', 'perform_interpolation'])
+
+        self.status.u_inter = []
+        self.status.f_inter = []
+        self.status.perform_interpolation = False
+
+    def post_spread_processing(self, controller, S, **kwargs):
```
tlunet, line 48: Could you rename `S` to `step` so it is more explicit ? And follow the new contribution guidelines :stuck_out_tongue: ...

pySDC/implementations/convergence_controller_classes/interpolate_between_restarts.py
```diff
@@ -0,0 +1,102 @@
...
+        Args:
+            controller (pySDC.Controller.controller): The controller
+            S (pySDC.Step.step): The current step
+
+        Returns:
+            None
+        """
+        if self.status.perform_interpolation:
+            for i in range(len(S.levels)):
+                L = S.levels[i]
```
tlunet, line 61: `level` instead of `L` ?

pySDC/implementations/convergence_controller_classes/interpolate_between_restarts.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def post_spread_processing(self, controller, S, **kwargs):
+        """
+        Spread the interpolated values to the collocation nodes. This overrides whatever the sweeper uses for prediction.
+
+        Args:
+            controller (pySDC.Controller.controller): The controller
+            S (pySDC.Step.step): The current step
+
+        Returns:
```
tlunet, line 56: Not sure the `Returns: None` is needed, saves a few documentation lines ...

pySDC/projects/Resilience/vdp.py
```diff
@@ -442,6 +443,78 @@ def check_step_size_limiter(size=4, comm=None):
...
+    from pySDC.helpers.plot_helper import figsize_by_journal
+
+    fig, axs = plt.subplots(4, 1, figsize=figsize_by_journal('Springer_Numerical_Algorithms', 1.0, 1.0), sharex=True)
+    restart_ax = axs[2].twinx()
+
+    colors = ['black', 'red', 'blue']
+    labels = ['interpolate', 'regular', 'keep iterating']
+
+    for i in range(3):
+        convergence_ontrollers = {
```
tlunet, line 464: Typo on `convergence_ontrollers`

pySDC/tests/test_convergence_controllers/test_InterpolateBetweenRestarts.py
```diff
@@ -0,0 +1,215 @@
...
+                level=L.level_index,
+                iter=step.status.iter,
+                sweep=L.status.sweep,
+                type='u_inter_double_check',
+                value=(interpolator.getInterpolationMatrix(np.append(0, nodes_new)) @ L.u[:])[:],
+            )
+
+
+@pytest.mark.base
+def test_InterpolateBetweenRestarts(plotting=False):
```
tlunet, line 85: Those test only check residuum after restart and iteration, if I understand correctly ?

It would be good to also check that the order corresponds, that is :

- overwrite the M-node solution on one step with order M-1 polynomial
- interpolate on a smaller (or bigger) time step, and verify that the interpolation is exact

tlunet (changes requested):

pySDC/tests/test_convergence_controllers/test_InterpolateBetweenRestarts.py
brownbaerchen: I don't really understand what you want here, to be honest. The tests make sure that the residual is smaller than usual after a restart, which means the interpolated values must have been a better initial guess than the usual `spread`.

I can test this more finely with what you suggested if you can elaborate a bit more. How do I verify that the interpolation is exact? Do I need to do a Dahlquist problem for that?

brownbaerchen: Ok. I did something weird, but I think this does what you wanted...? :D

tlunet: Seems like it, nice ! Minor detail tho : you should probably use `numpy.polynomial.Polynomial` instead of `numpy.poly1d`, see https://numpy.org/doc/stable/reference/routines.polynomials.html

tlunet:
Seems all good to me !

tlunet (approved):

PR #282, opened 2023-03-10
Estimate embedded error using different collocation problems

brownbaerchen:
This is step 2 out of 3 for adaptivity based on different collocation problems. We already had the switch between quadrature rules between iterations and this convergence controller basically records the solution when after collocation problem is converged and before it is exchanged for the next one.

I just integrated the tests for this in the tests of the convergence controller that switches the quadrature rule.
The plot below shows the error estimate compared to the local error wrt a reference solution for an advection problem and they agree really well and have the correct order, which is what the test confirms.
In the left panel the type of the nodes is changed, in the middle panel the number of nodes is gradually increased and in the right panel the number of nodes is decreased. In all cases we can estimate the local error of all but the most accurate method.
The dashed lines show how many more iterations are needed to switch to the new collocation method, but don't worry about that now. Performance will be analysed when the adaptivity is ready.
Figure_1

### DIFF
--- a/pySDC/implementations/convergence_controller_classes/estimate_embedded_error.py
+++ b/pySDC/implementations/convergence_controller_classes/estimate_embedded_error.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-from pySDC.core.ConvergenceController import ConvergenceController, Pars
+from pySDC.core.ConvergenceController import ConvergenceController, Pars, Status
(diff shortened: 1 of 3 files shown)


pancetta:
Other than that: LGTM

pancetta:
Aha! There you are!

PR #283, opened 2023-03-13
Reference solution for Quench problem

brownbaerchen:
Turns out I have been very stupid. The reason that scipy wouldn't give me a reference solution in time was because the problem is so stiff that RK45 is not a good method to solve the problem. But `scipy.solve_ivp` can also do implicit time stepping with a "BDF" scheme. This seems to work reasonably well to produce reference solutions rather fast. At least the error is decreased when I increase the accuracy requirements of SDC. I failed to show any sort of order, but I don't know if the reference solution is bad or if the problem is nasty. At least I can get errors down to machine precision with the reference solution, so I thought it's probably fine.

Also, I implemented an exponential transition term in the leak. However, it doesn't lead to faster heat up than the linear part because it is rescaled to fit between 0 and 1. In that scenario linear can be steeper. I am not a modelling expert. Maybe there is a different way to model exponential growths that connects smoothly to the other terms that I don't know about.

Also, the interpolation stuff resulted in a flaky test. When you replace the solution in a van der Pol problem by a random polynomial, the Newton solver does not converge every time. I hope I fixed that in the last commit where I just deactivated crash when Newton didn't converge. We'll see. At least on my fork this went through.

### DIFF
--- a/pySDC/tests/test_convergence_controllers/test_InterpolateBetweenRestarts.py
+++ b/pySDC/tests/test_convergence_controllers/test_InterpolateBetweenRestarts.py
@@ -155,6 +155,7 @@ def run_vdp(hook, adaptivity=True):
         'newton_tol': 1e-9,
         'newton_maxiter': 99,
         'u0': np.array([2.0, 0.0]),
+        'crash_at_maxiter': False,
     }
 
     # initialize step parameters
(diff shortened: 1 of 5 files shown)


pySDC/tests/test_convergence_controllers/test_InterpolateBetweenRestarts.py
```diff
@@ -155,6 +155,7 @@ def run_vdp(hook, adaptivity=True):
         'newton_tol': 1e-9,
         'newton_maxiter': 99,
         'u0': np.array([2.0, 0.0]),
+        'crash_at_maxiter': False,
```
pancetta, line 158: So... where is that parameter used, then?

Issue #255, opened 2023-01-17
Rework problem class implementation implementation


tlunet:
Following a template applied to the [advectionNd](https://github.com/Parallel-in-Time/pySDC/blob/v6/pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py) problem class, rework all other implementations to get to the same standards :

- [ ] use of the default parameter approach and inheritance from `RegisterParams` base class (see #206)
- [ ] update (with eventual rewrite) of full documentation
- [ ] compliance to new problem base interface (see #254)
- [ ] compliance to new Contribution Guidelines

Issue #206, opened 2022-12-16
Default parameters for problem classes


brownbaerchen:
As pySDC is a prototyping code, more often than not we don't care about the precise setup of a problem. Instead, we just want something that we can use to test the order of a scheme of something like that.
Doing things like that could be greatly accelerated by providing default parameters for the problems.

I am unclear on what the best approach for this is.
My first proposal would be to make a class method called something like `get_default_description`, which will setup an entire description object such that you can immediately get started.
Since there is usually only a unique sweeper class allowed for each problem, it makes sense to include more parameters than just problem parameters.

But what about the quadrature? I guess this is a point where it is difficult to setup a default that makes sense for everybody.
I am still in favour of just setting up something that works and then you change it rather than you need to setup everything every time. For instance, I might want to use a fixed number of iterations, while you might not, but we may be able to agree on a type of quadrature nodes.

A different option would be to have the class method ask for a dictionary of quadrature parameters including residual tolerance, step size and iteration count, which are typically spread across different parts of the description object. Of course, I want to leave them where they are, but this could be sorted out automatically in this class method.

I am unsure about whether we need defaults for the controller parameters, because these already have defaults that are pretty problem unspecific.
This means you have to setup a hook just if you want to have the solution after the step, but the hook is probably the part that the least number of people will agree on anyways, since we all want to extract different information for our tests. For instance, in other hooks I have seen, people usually store the solution at the time corresponding to the beginning of the step, but I always store it at the end of the step, since the plots look messed up if you use adaptive step sizes otherwise.

So I suggest adding this to the core problem class:
 ```
@classmethod
def get_default_description(cls, quadrature_params):
    # check that the quadrature_params contain everything that we need
    required_keys = ['maxiter', ...]
    for k in required_keys:
        assert k in quadrature_params.keys(), f'Need "{k}" in quadrature_params to setup defaults'

    # initialise a dictionary for each of the entries of the description class
    step_params = dict()
    ...

    # allocate the quadrature params to where they belong in the description
    step_params['maxiter'] = quadrature_params['maxiter']
    ...

    return description 
```

Then the individual problem classes would inherit and add their problem parameters on top:
```
@classmethod
def get_default_description(cls, quadrature_params):
     description = super().get_default_description(quadrature_params)
     description['problem_params']['power_level'] = 9001
     return description
```

Any problems with this / better ideas welcome!

tlunet:
[...]

PR #285, opened 2023-03-20
Proposition for release 5.2

tlunet:
This PR mostly solve #206, and prepare `pySDC` for the next changes toward #255.

**Main changes**

- **problem class define their parameters explicitly with the constructor, and not with a dictionary**. This allows any user to see directly in the class signature what parameters are needed and the eventual default values. Documentation can be easily added to the class for additional help on the choices of the parameters (some have been added already).
- **'use of the `FrozenClass` dictionnary has been removed**. Now parameters are stored directly as attributes of the problem class, thanks to the convenient method `_makeAttributeAndRegister`, inherited from the base problem class, that creates attributes for each parameters, eventually as read-only. Also, it allows to access all parameters with an inherited `params` property that return a dictionary (_e.g_ for output).

:scroll: A starting documentation has been added in the contribution guide on how to define a custom problem with this new interface, see [here](https://github.com/Parallel-in-Time/pySDC/blob/v6/docs/contrib/04_custom_implementations.md).

> :bell: this does not break the API _stricto sensu_, as dictionaries containing the parameters for the problem class can still be passed to the controller classes. However some behavior are not allowed anymore. For instance, passing a parameters that is not expected by the problem class automatically throw an explicit error :

```python
>>> LeakySuperconductor(Kv=1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: LeakySuperconductor.__init__() got an unexpected keyword argument 'Kv'
```

**Additional changes**

- datatype classes are now defined as class attributes of the problem class, which avoid passing them in the constructor.
- base problem class has now a `u_init` and `f_init` properties that create a new datatype for `u` or `f` evaluation, that is `prob.u_init = prob.dtype_u(prob.init)` and `prob.f_init = prob.dtype_f(prob.init)`.
- advection and diffusion ND_FD problem classes now inherit from a common `GenericNDimFinDiff` class that initializes the mesh and derivation matrix, and provide the main methods shared by all classes.
- the `type` parameter for the differentiation matrix computation has been changed to `stencil_type`. :warning: this is actually the most API-breaking change of this PR.
- some `pragma: no cover` have been added to the GPU problem classes and tests, since they are not tested by CI, hence fail coverage tests. But the new interface and tests for GPU have been tested separately by @brownbaerchen on JSC a cluster.
- a new playground with experimental improved datatype classes for V6 has been introduced [here](https://github.com/Parallel-in-Time/pySDC/tree/v6/pySDC/playgrounds/datatypes), and will be used to compare performances  with current datatype classes of pySDC.

And if I forgot something, I'll add it later :ghost:


pySDC/core/Common.py
```diff
@@ -0,0 +1,75 @@
...
+        readOnly : bool, optional
+            Wether or not store the parameters as read-only attributes
+        """
+        if len(names) > 1 and localVars is None:
+            raise ValueError("a dictionnary must be provided in localVars with parameters values")
+        # Set parameters as attributes
+        for name in names:
+            try:
+                super().__setattr__(name, localVars[name])
+            except KeyError:
```
pancetta, line 59: Add coverage pragma

pySDC/core/Problem.py
```diff
@@ -1,74 +1,96 @@
...
-            params (dict): set or parameters
-        """
+    logger = logging.getLogger('problem')
+    dtype_u = None
+    dtype_f = None
 
-        self.params = _Pars(params)
-        self.work_counters = {}
+    def __init__(self, init):
+        self.work_counters = {}  # Dictionnary to store WorkCounter objects
```
pancetta, line 60: Typo

pySDC/implementations/datatype_classes/mesh.py
```diff
@@ -3,6 +3,13 @@
 from pySDC.core.Errors import DataError
 
 try:
+    # Hack in case dolfin is used => must be imported before mpi4py
+    try:
+        import dolfin as df
```
pancetta, line 8: Why is this needed here? Or: why wasn't it needed before?

pancetta (commented):
LGTM, albeit very extensive update. Tests seem to go through, which is reassuring. Do you plan to continue to work on this PR (I see some TODOs in the documentation)?

pySDC/implementations/datatype_classes/mesh.py
tlunet: At some point I had a test that imported a `fenics_mesh` right after importing the standard `mesh`. Because `mesh` also import `mpi4py` (and also initialize MPI), and `dolfin` does it's own initialization of MPI with C routines, then it produces a call of `MPI_Finalize()` twice.
However, importing `dolfin` before `mpi4py` avoid any clashes of this sort, since `mpi4py` recognizes that MPI has been already initialized before, and uses in the end the same MPI library as `dolfin`.

I spend some time trying to understand that problem, and put this modification in `mesh.py` mostly as a reminder. There is probably a cleaner approach, by handling pySDC imports globaly on the main package level (_i.e_ in `pySDC.__init__.py`), and then let the other sub-modules use packages that have already been imported. But I leave it for latter ... can put a TODO there eventually :wink:

tlunet:
> LGTM, albeit very extensive update. Tests seem to go through, which is reassuring. Do you plan to continue to work on this PR (I see some TODOs in the documentation)?

Probably some time to time, or once I get to travel again (and a lot of time in trains). Next stage is the new datatype approach. If it works (more efficient, more simple to use, etc ...), then I'll be able to move to the more generic problem class following #253, which will (eventually) finish the rework of the problem classes.

About the TODOs in the documentation, that could be material for `good first issues` : anyone working on pySDC can improve the documentation of each problem one at the time, and propose doku PR targeting one specific problem (reworking the documentation of all problem classes all-together seems a bit too much, I think ... better use a step-by-step approach ?)

pySDC/core/Common.py
```diff
@@ -54,6 +54,7 @@ def _makeAttributeAndRegister(self, *names, localVars=None, readOnly=False):
             raise ValueError("a dictionnary must be provided in localVars with parameters values")
         # Set parameters as attributes
         for name in names:
+            # pragma: no cover
```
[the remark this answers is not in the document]
brownbaerchen: why do you want to exclude this from coverage anyways? I thought we decided only to exclude plotting only stuff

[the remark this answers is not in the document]
pancetta: And errors we're not going to test anytime soon.

docs/contrib/04_custom_implementations.md
```diff
@@ -1,6 +1,61 @@
...
+
+$$
+\frac{du}{dt} = f(u, t), \quad u(0)=u_0, \quad t \in [0, T].
+$$
+
+where $u$ a vector or scalar, representing one or more variables.
+
+The type of variable, the definition of the right-hand side $f(u,t)$ and initial solution value $u_0$ are defined in a given `Problem` class.
+
+... to be continued ...
```
pancetta, line 13: What's supposed to go here?

pancetta:
> > LGTM, albeit very extensive update. Tests seem to go through, which is reassuring. Do you plan to continue to work on this PR (I see some TODOs in the documentation)?
> 
> Probably not on the PR itself, but on the following rework for sure (once I get to travel again and a lot of time in trains). Next stage is the new datatype approach. If it works (more efficient, more simple to use, etc ...), then I'll be able to move to the more generic problem class following #253, which will (eventually) finish the rework of the problem classes.
> 
> About the TODOs in the documentation, that could be material for `good first issues` : anyone working on pySDC can improve the documentation of each problem one at the time, and propose doku PR targeting one specific problem (reworking the documentation of all problem classes all-together seems a bit too much, I think ... better use a step-by-step approach ?)

Good point. Could you add that to the `contribution.md` file, then?

docs/contrib/04_custom_implementations.md
tlunet: The full definition of a problem class (with base methods, abstract ones, etc ...). Supposed to come later following #253

pySDC/implementations/datatype_classes/mesh.py
tlunet: Yeah, I was quickly testing problem classes at some point with [`pySDC.tests.test_problem.py`](https://github.com/Parallel-in-Time/pySDC/blob/v6/pySDC/tests/test_problem.py), willing to generalize it for different problem classes.
This requires importing all problem classes on the top of the module (and I added one fenics problem at some point ...). I didn't continue on this way because of the mpi4py/dolfin clash, but I still wanted to include somewhere the safety, because we will probably get this problem again later if we implement more generic/complete testing procedure later

tlunet: And there is later the performance tests for all datatypes => will need to import all datatypes globaly at some point (PS : I'm not very found of local imports, because they can hide problems like this one, which is mostly an ostrich approach)

pancetta:
Yeah, let's just do that.

pySDC/implementations/problem_classes/AllenCahn_2D_FD.py
```diff
@@ -119,39 +119,39 @@ def solve_system(self, rhs, factor, u0, t):
...
             # newton update: u1 = u0 - g/dg
             # u -= spsolve(dg, g)
-            u -= cg(dg, g, x0=z, tol=self.params.lin_tol, atol=0)[0]
+            u -= cg(dg, g, x0=z, tol=self.lin_tol, atol=0)[0]
             # increase iteration count
             n += 1
             # print(n, res)
 
-        # if n == self.params.newton_maxiter:
+        # if n == self.newton_maxiter:
```
brownbaerchen, line 150: while we are in the process of cleaning up, I guess we can either remove this or add a flag for this.

pySDC/implementations/problem_classes/AllenCahn_2D_FD.py
```diff
@@ -327,40 +305,40 @@ def solve_system(self, rhs, factor, u0, t):
...
```
brownbaerchen, line 337: cleanup? same for line 331 maybe

pySDC/implementations/problem_classes/AllenCahn_2D_FD.py
```diff
@@ -461,39 +428,39 @@ def solve_system_2(self, rhs, factor, u0, t):
...
```
brownbaerchen, line 459: why was this copy pasted everywhere at some point? :D

pySDC/implementations/problem_classes/FastWaveSlowWave_0D.py
```diff
@@ -13,27 +13,17 @@ class swfw_scalar(ptype):
     Attributes:
     """
 
-    def __init__(self, problem_params, dtype_u=mesh, dtype_f=imex_mesh):
+    dtype_u = mesh
+    dtype_f = imex_mesh
+
+    # TODO : choose default values
```
brownbaerchen, line 19: did you check if there are some parameters in the paper that we could use for defaults? I guess @pancetta would know.

pancetta:
So, when do you suggest to tag the new release? After this PR or after #253 ?

pySDC/implementations/problem_classes/GeneralizedFisher_1D_PETSc.py
```diff
@@ -106,19 +104,19 @@ class Fisher_reaction(object):
     Helper class to generate residual and Jacobian matrix for PETSc's nonlinear solver SNES
     """
 
-    def __init__(self, da, params, factor):
+    def __init__(self, da, prob, factor):
```
brownbaerchen, line 107: also here defaults or TODO?

tlunet:
After this PR. It's already a big change, and it's maybe better to go to v6 incrementally, right ?

pySDC/projects/Resilience/advection.py
```diff
@@ -40,7 +40,7 @@ def run_advection(
     sweeper_params['num_nodes'] = 3
     sweeper_params['QI'] = 'IE'
 
-    problem_params = {'freq': 2, 'nvars': 2**9, 'c': 1.0, 'type': 'backward', 'order': 5, 'bc': 'periodic'}
+    problem_params = {'freq': 2, 'nvars': 2**9, 'c': 1.0, 'order': 5, 'bc': 'periodic'}
```
brownbaerchen, line 43: why did you remove the type here? Did you mean to replace it with stencil_type or did I do sth stupid?

brownbaerchen (commented):
You where really busy! 👍 

I was just scrolling through parts of the code and nitpicking a little. Mainly, I am not a huge fan that there is quite some stuff commented out. But that's of course not your fault and you can't fix everything. So feel free to ignore the comments.

Did you fix all the problems, even those that are not tested already? Unfortunately, I wasn't able to do anything about this since Amsterdam...

pySDC/implementations/problem_classes/FastWaveSlowWave_0D.py
pancetta: You mean for `lambda_s`, `lambda_f` and `u0`, yes?

brownbaerchen:
> After this PR. It's already a big change, and it's maybe better to go to v6 incrementally, right ?

I agree. I think the changes that existing users need to make are rather small and hopefully they would be happy about simplifying their code like Thibaut did for the stuff in master. I don't know if we will do any more API breaking things anytime soon and I don't want poor Thibaut to have to merge master all the time.

pySDC/implementations/problem_classes/FastWaveSlowWave_0D.py
pancetta: If so, then I'd suggest
`lambda_s = -1`
`lambda_f = -1000`
`u0 = 1`
if that actually works..

pySDC/core/Common.py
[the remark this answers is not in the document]
pancetta: Yeah, you're right. Move the pragma to the `except` statement, then.

pySDC/projects/Resilience/advection.py
tlunet: Haem ... was maybe due to standard tests that were done with `stencil_type="center"`, even if you could put `type`: `backward` in the previous version, which was simply ignored. I may have done a replacement on all files of this ...
But I'm not so sure : do you want `backward` stencil here specifically ?

pySDC/core/Common.py
```diff
@@ -0,0 +1,76 @@
...
+                raise ValueError(f'value for {name} not given in localVars')
+        # Register as class parameter
+        if readOnly:
+            self._parNamesReadOnly = self._parNamesReadOnly.union(names)
+        else:
+            self._parNames = self._parNames.union(names)
+
+    @property
+    def params(self):
+        """Dictionnary containing names and values of registered parameters"""
```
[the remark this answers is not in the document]
tlunet: Corrected with a global search, thanks !

pySDC/core/Common.py
```diff
@@ -0,0 +1,76 @@
...
+        localVars : dict
+            Dictionnary containing key=names and values=paramValues for each
+            parNames given in names. Can be provided, for instance, using
+            `locals()` built-in dictionary. MUST BE provided as soon as
+            names contains anything.
+        readOnly : bool, optional
+            Wether or not store the parameters as read-only attributes
+        """
+        if len(names) > 1 and localVars is None:
+            raise ValueError("a dictionnary must be provided in localVars with parameters values")
```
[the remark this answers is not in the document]
tlunet: Corrected with a global search

pySDC/implementations/problem_classes/AllenCahn_2D_FD.py
tlunet: Naaaa ... :joy:

pancetta:
Let me know when you're done, I've lost track...

tlunet:
@pancetta almost all modifications are done (in particular, the addition of a page for docstring documentation in contribution guidelines). The only point left is the modification for the `dolfin`/`mpi4py` clash ...

pancetta:
> @pancetta almost all modifications are done (in particular, the addition of a page for docstring documentation in contribution guidelines). The only point left is the modification for the `dolfin`/`mpi4py` clash ...

OK, if you remove the hack, will it still work? If so, nothing changed with respect to the state we currently have and I'd rather like to see this as a todo for later.

tlunet:
Oh, actually it does not ! And I remember why :sweat_smile: ... When collecting all the tests, pytest import each test files, in particular `pySDC.tests.test_problem.py`, which has the following header :

```python
# Test some functionality of the core problem module
import pytest
import numpy as np
from pySDC.core.Problem import ptype
from pySDC.implementations.problem_classes.LogisticEquation import logistics_equation

PROBLEMS = {
    logistics_equation: {
        'probParams': dict(u0=2.0, newton_maxiter=100, newton_tol=1e-6, direct=True, lam=0.5, stop_at_nan=True),
        'testParams': {'tBeg': 0, 'tEnd': 1.0, 'nSteps': 1000, 'tol': 1e-3},
    }
}
```

This allows to generically define tests for several problems (very useful for later ...), and run it with 

```python
@pytest.mark.base
@pytest.mark.parametrize('probType', PROBLEMS.keys())
def testProblem(probType):
    # ...
```

pancetta:
I assume this goes away if you don't to global stuff but wrap it in a class and setup/teardown methods? I'm really not a big fan of the hack, esp. because it is in a very central place for a very special case.

pySDC/tests/test_projects/test_AC/test_simple_forcing.py
```diff
@@ -17,6 +17,8 @@ def test_main_parallel():
     # try to import MPI here, will fail if things go wrong (and not later on in the subprocess part)
     import mpi4py
 
+    del mpi4py
```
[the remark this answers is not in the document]
tlunet: Remove a pyflakes warning : `mpi4py` imported by unused. I'm usually in favor of avoiding unused imports in general, and in this case it make sense looking at the comment below. What about 

```python
# try to import MPI here, will fail if things go wrong (and not in the subprocess part)
try:
    import mpi4py
except ImportError:
    raise ImportError('petsc tests need mpi4py')
finally:
    del mpi4py
```

More explicit and elegant ... ?

tlunet:
So I've found a solution for the `dolfin` problem. But now codecov is beeing annoying ... :disappointed:

tlunet:
So ... this is a bit problematic :

grafik

Why codecov wants to test `if` statements ?

pancetta:
Yeah, whatever. Don't worry about that. I think we're good to merge. Thanks a lot for your hard work, again and again.

PR #288, opened 2023-04-19
Error estimate by Taylor expansion across the quadrature nodes

brownbaerchen:
After solving a collocation problem with N nodes exactly, you have a solution to the problem of order up to 2N, but you also have stages, which are all order N accurate.
By doing a Taylor expansion of the solution around the last collocation node and using both the solutions as well as right hand side evaluations to cancel terms, we can cancel 2N terms in the Taylor expansion. This is much more accurate than the stages, so what remains is not the ninth Taylor term, but the local error of the last stage.

In this PR, the extrapolation based error estimate, which is used in Hot Rod to compute the local error of a step by cancelling terms in Taylor expansions using solutions of previous steps, is applied to the stages inside the collocation problem. There is a test that verifies that the order of accuracy of this error estimate is equal to N+1. The +1 comes from local vs. global errors, I assume. The test is run for an ODE and a PDE and for Gauss as well as Radau-right nodes.

I didn't do anything with this yet because I am mainly writing the paper, but I got excited and had to quickly do this.
It took slightly longer than expected because of a bug in the hook that records local errors, which changed initial conditions of the steps. I don't think this caused major problems so far because this only happened after steps were converged, but it's good that this bug is fixed now.
The implementation is not too polished, because I need to figure out how to best use this when the time comes.

In the future, we want to use adaptivity in SDC with converged collocation problems, because then we can use inexactness, parallel SDC, all the things. We already have an implementation based on interpolating between different collocation problems available, but this may be nicer because it's virtually free.

Because the CI pipeline is generally broken atm, I can't verify that I didn't break everything, but I didn't change much, so should be fine.


pySDC/tests/test_projects/test_resilience/test_extrapolated_error_within_Q.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+from pySDC.projects.Resilience.extrapolation_within_Q import check_order
+from pySDC.projects.Resilience.advection import run_advection
+from pySDC.projects.Resilience.piline import run_piline
```
pancetta, line 4: Please add the imports to the test function, because otherwise MPI doesn't like it.

PR #296, opened 2023-04-26
Quench resilience

brownbaerchen:
This is the rest of the mess that I created over the past couple of weeks.

Changes include:

 - Three different reference solutions for Quench problem: Scipy, SDC with small dt and DIRK
 - Moved the strategies for quick setup of different SDC versions to their own file and added a few
 - Added script for making work-error plots
 - Added script to make all the plots that go in the paper
 - small stuff

The work error stuff is not tested, because that's difficult to do, so coverage may go down a bit :(


pancetta:
Looks like you managed to create the new leader in the list of slowest tests.. https://github.com/Parallel-in-Time/pySDC/actions/runs/4819699085/jobs/8583242135?pr=296#step:4:263

PR #299, opened 2023-05-05
Copy generated plots into subdir of created html

jakob-fritz:
The previous version of the generated website did not include the plots that were created in the same CI.

Now, the plots are copied into a subdir of the website and are therefore also published, as all content of that directory is used for publication. The links in the RST-files have been updated accordingly.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -267,6 +267,7 @@ jobs:
           sphinx-build -b html docs/source docs/build/html
           mkdir -p docs/build/html/coverage
           mv htmlcov/* docs/build/html/coverage/.
+          cp -r data docs/build/html/
 
 #      - name: Build html II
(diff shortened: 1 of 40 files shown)


brownbaerchen:
I like the changes because it makes the paths in the Readmes easier! However, I am not sure if it solves the problem that we're having. The issue is that the plots are not displayed on GitHub because the paths don't lead anywhere in the repository. That is still the case, no? The plots are not persistent in the repository beyond the action, as far as I understand. Since we don't want to check in the plots, I am not sure if what I want is possible, though.

jakob-fritz:
I assume, that it helps, as the plots are now part of the directory that is deployed [in the CI](https://github.com/Parallel-in-Time/pySDC/blob/f2e51e4f35cadad8fad98aad0a96ddf377f97d28/.github/workflows/ci_pipeline.yml#LL288).

Therefore, it should also be persistent (i.e. the links should continue to work), as they don't link to the repo, but to another relative location.

It may be seen whether this approach works, when the deploy-job in the current check is executed.

PR #302, opened 2023-05-08
Updated links to images and result-data again

jakob-fritz:
Now also the links to result-files are included. These files are included literally; not as an image

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -267,7 +267,6 @@ jobs:
           sphinx-build -b html docs/source docs/build/html
           mkdir -p docs/build/html/coverage
           mv htmlcov/* docs/build/html/coverage/.
-          cp -r data docs/build/html/
 
 #      - name: Build html II
 #        if: "!contains(github.event.head_commit.message, '[CI-no-benchmarks]')"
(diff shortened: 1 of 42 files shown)


jakob-fritz:
In artifact links are still wrong...

jakob-fritz:
> Regardless of the results, the problem we wanted to solve still persists. The plots are shown on the website, but not on GitHub. Maybe this is not worth spending lot's of time on and we just go back to the way things were?

Where on Github would you like the results to appear? Since they are not checked in (i.e. staged and commited) they are not shown in the files-section, but are (in principel) present on the website (currently not correctly linked, but uploaded)

jakob-fritz:
Ok. I don't assume that to be possible, because it means having versioned files and unversioned files in the same place.

I would rather refer to an artifact containing the files for that usecase.

I assumed, that the issue is that the website did not include the plots because they are outside of the "docs" directory and therefore not uploaded

pancetta:
Sounds like this PR can be closed, then?

jakob-fritz:
If you want to, you can close this PR and roll back.

As an alternative, I can update the PR to roll back to the original state, as the current state is worse than the original state in that it does not create a working website.

pancetta:
Yes please!

jakob-fritz:
Now it seems to work again (at least it does in the artifact).

So this PR can be merged

PR #306, opened 2023-05-12
Adaptivity based on the new error estimate and more lean MPI tests

brownbaerchen:


brownbaerchen:
Looks like I need to look into why these tests are failing again a bit more. I think I got confused by my many branches a bit. Sorry!

PR #320, opened 2023-06-13
Implementing Local and Global Convergence Plots for Second-Order SDC Paper

ikrom96git:
The codes enable the generation of local and global convergence plots for the second-order SDC paper.


pancetta:
Depending on how long these scripts will run: can you write tests for them?

ikrom96git:
Sure, I will write tests for them later also.

PR #322, opened 2023-06-26
Small tweaks for RK sweepers

brownbaerchen:
This PR includes two small changes to RK implementations.

First, @ikrom96git noticed that when supplying a residual tolerance larger than 0, the RK sweepers would do nothing and don't tell you about it. This is now fixed by setting the residual tolerance to -1 in the setter for the level as well as raising a warning when this is done.
Because RK methods are direct with a single iteration, the residual drops to machine precision after the first iteration. We can compute this to confirm that the method does what it's supposed to, but this is very computationally expensive and could dominate the computational time for explicit methods. So the residual is usually never computed and set to 0 instead.

The second change is regarding tests. The same things are testet as before (although I removed some redundant tests, I think), but in a more finely grained fashion. Instead of testing if all methods are A stable at the same time, there are now individual tests for the different methods. So if a test failed, you don't see: Some RK method is no longer A stable, but you see: These specific RK methods are no longer A stable. This also makes it easier to add new methods.
Because I want to test that the methods now work when giving a residual tolerance larger than 0, annoyingly I had to rewrite the script that computes the order a bit... I guess when I tried to write one script that does all, I failed to make things more simple and instead just complicated everything more than necessary.

@tlunet: Since Robert is on vacation and this is only a small change, I think you can merge this PR if you see fit. I don't think we need to wait until he comes back. Feel free to suggest improvements of course!

### DIFF
--- a/pySDC/tests/test_Runge_Kutta_sweeper.py
+++ b/pySDC/tests/test_Runge_Kutta_sweeper.py
@@ -8,6 +8,10 @@ COLORS = {
     'Cash_Karp': 'teal',
 }
 
+EXPLICIT_METHODS = ['RK1', 'MidpointMethod', 'CrankNicholson']
+IMPLICIT_METHODS = ['RK1', 'MidpointMethod', 'RK4', 'Cash_Karp']
+EMBEDDED_METHODS = ['Cash_Karp', 'Heun_Euler', 'DIRK34']
+
 
 def get_sweeper(sweeper_name):
     """
@@ -144,8 +148,7 @@ def plot_stability_single(sweeper_name, ax=None, description=None, implicit=True
     ax.legend(frameon=False)
 
 
-@pytest.mark.base
-def test_all_stability():
+def plot_all_stability():
     """
     Make a figure showing domains of stability for a range of RK rules, both implicit and explicit.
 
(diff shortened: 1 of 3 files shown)


pySDC/tests/test_Runge_Kutta_sweeper.py
```diff
@@ -8,6 +8,10 @@
     'Cash_Karp': 'teal',
 }
 
+EXPLICIT_METHODS = ['RK1', 'MidpointMethod', 'CrankNicholson']
```
tlunet, line 11: Is `RK1` Forward Euler or Backward Euler ? And why is `MidpointMethod` both in `EXPLICIT_METHODS` and `IMPLICIT_METHODS` ?

tlunet (commented):
Need probably a little clarification on the names for RK methods, but other than that seems all right.

pySDC/tests/test_Runge_Kutta_sweeper.py
brownbaerchen: I implemented most RK rules following the [Wikipedia page on list of RK methods](https://en.wikipedia.org/wiki/List_of_Runge–Kutta_methods). There is for instance both explicit and implicit Midpoint. 
Maybe it wasn't the best decision, but instead of separate classes, you supply as a parameter which version you want. Same is true for RK1. If you give `RK1` as sweeper class with `sweeper_params = {'implicit': True}`, you get backward Euler and vice versa.

tlunet: Then maybe `ExplicitMidpoint`, `ImplicitMidpoint`, `ForwardEuler` and `BackwardEuler` so the names are more explicit ?

tlunet:
Seems good now ! Should I do the merge myself once all tests pass, or do we wait for @pancetta  to do it ?

@ikrom96git be aware of those changes ...

PR #326, opened 2023-07-07
Refined tutorial for 12th PinT Workshop

brownbaerchen:
# Installation
We need `mpi-fft` not for the tutorial, but for the tests. So I added that to the environment file. Also, I added `ipython` because it lets us make a Jupyter kernel from the virtual environment.
On my laptop, the python 3 kernel in Jupyter used the python from the base conda environment and this was the most elegant solution I could come up with.

Also, I excluded `slow` tests from the tests during installation. I believe we are not skipping anything too important there.

# Tutorials
I split the tutorial in three notebooks. If you are happy with this, we can remove the `playground` notebook.

The first notebook runs the AC problem and plots the radius over time. It's the same as the first half of the playground notebook, but I added more comments as to what the problem is and why the parameters are what they are.

The second notebook implements the heat equation "from scratch". This was not part of the playground notebook.
I basically stripped everything from the implementation in pySDC that I didn't deem necessary and inherited directly from the core problem class rather than the generic ND FD class.

The third and final notebook is the second half of the playground notebook with the diagonal sweeper.
I added a few equations to explain quick and dirty what we want to do.
I also added a few assert statements to highlight when we are only implementing special cases of parameters.
For instance, we where using initial guess "zero", but only "spread" was implemented.

Throughout the notebooks I made an effort to include good comments.
That means docstrings for the functions and I removed some comments that I found superfluous.
I am sure I missed a few in both cases.

I tried to write the notebooks such that they are pretty much self-explanatory. I think they may in total be too long to do at the workshop, but people can finish them at home if they want. Or they can maybe choose if they want to implement either the heat equation or the sweeper if we find time to be an issue.

GitHub is not doing the best job with notebooks, I think. I hope you come up with a good way of added some criticism...


pancetta:
Thanks, looking great! Did you check that the environment actually builds? I vaguely remember that in my case (Apple ARM as in your case) I had to remove one of these mpi dependencies to get this environment to build.

brownbaerchen:
Yes, I was using the tutorial environment installed with micromamba. I am about 81% certain I remembered to include all the steps I had to take.

pancetta:
I'll test this in more detail on Tuesday. Thanks for the PR!

Issue #327, opened 2023-07-11
Implement Multistep Methods for Comparison


brownbaerchen:
We have diagonally implicit and explicit Runge-Kutta methods implemented for comparison between SDC and other algorithms.
If anyone ever feels bored, I think it would be nice to implement multistep methods as well.

I believe this can be more or less easily done by writing a new sweeper which replaces the SDC functionality with functions that are called at the appropriate time, but which instead do a different thing from what their name is.
For instance, the sweeper is based on a collocation problem, but multistep methods don't do collocation. They do something that looks quite similar, though. Instead of generating solutions at intermediate collocation nodes and updating all of them from each other, multistep methods update the solution at the current step from solutions at previous time points.

So, to generate a multistep method interface, we need to do: 
 - `predict`: Fill `u` and `f` in the levels with the values from previous time steps
 - `update_nodes`: Compute the solution to the current step from the previous solutions, no loop over "nodes"
 - `compute_end_point`: Return the solution at the last "node"
 - `compute_residual`: We don't need to compute the residual at all, so do nothing.

Then, we would need to write some actual methods that only need to implement the `update_nodes` function to compute the solution to the current step.
And, we would need to profile this to make sure it is not doing any unnecessary work!

Again, this is not desperately needed! But I think it would be nice to have something in the interface of pySDC that does other methods in the language of SDC. If you feel like implementing this, please go ahead!

Please remember to test your methods. Ideally, you would test that they
 - have the expected order of accuracy,
 - are as stable as expected,
 - and perform exactly as many right hand side evaluations as needed.

But feel free to deviate from this.

lisawim:
That's a good idea! For DAEs, BDF methods are very familiar for solving them. 

I want to add an important thing:
For implementing multistep methods, one also have to think about the start. For example, the k-step BDF Method needs k-1 initial values to start. Mostly, these initial values are not known, only the one at t0. So, in this case, the other k-2 values can be computed via other methods. When this is done, the computation is not started at time t0 anymore, but instead from time t_{k-1}. This is true for all multistep methods.
Thus, it makes sense to implement this in the controller class as well.

I was already starting with implementing BDF for DAEs.

tlunet:
Are you thinking of multistep through the nodes or through the time-steps ? Because in both case, I see some issue there ...

brownbaerchen:
I believe you can also do the startup period in the `predict` function. I would avoid involving the controller at all cost ;)

[...]

PR #328, opened 2023-07-13
Added multistep methods as a sweeper

brownbaerchen:
As discussed in #327, you can abuse the sweeper class to do multistep methods.
I ended up doing this differently from what I initially thought and I don't know if this can be extended to arbitrary multistep methods.
In particular, at the moment it is required to use a fixed step size and restarts will probably also mess everything up.

I thought I start with a simple implementation of the most basic things and anybody is welcome to improve on this.
@lisawim, maybe we can discuss how we get the methods you want into this notation? I don't know exactly what you did with multistep methods so far.

I implemented methods with at most 2 steps, so not overly exciting, but for this I didn't have any issues with generating starting values, as they can be one order below the scheme.

I test the order of accuracy in time, the stability and the number of right hand side evaluations.

### DIFF
--- a/pySDC/implementations/sweeper_classes/Multistep.py
+++ b/pySDC/implementations/sweeper_classes/Multistep.py
@@ -11,17 +11,17 @@ class Cache(object):
         - t (list): Contains time of previous steps
     """
 
-    def __init__(self, num_entries):
+    def __init__(self, num_steps):
         """
         Initialization routing
 
         Args:
-            num_entries (int): Number of entries for the cache
+            num_steps (int): Number of entries for the cache
         """
-        self.num_entries = num_entries
-        self.u = [None] * num_entries
-        self.f = [None] * num_entries
-        self.t = [None] * num_entries
+        self.num_steps = num_steps
+        self.u = [None] * num_steps
+        self.f = [None] * num_steps
+        self.t = [None] * num_steps
 
     def update(self, t, u, f):
         """
@@ -45,14 +45,17 @@ class Cache(object):
         Print the contents of the cache for debugging purposes.
         """
         string = ''
-        for i in range(self.num_entries):
+        for i in range(self.num_steps):
             string = f'{string} t={self.t[i]}: u={self.u[i]}, f={self.f[i]}'
 
         return string
 
 
 class MultiStep(sweeper):
-    def __init__(self, params, alpha, beta):
+    alpha = None
+    beta = None
+
+    def __init__(self, params):
         """
         Initialization routine for the base sweeper.
 
@@ -62,10 +65,12 @@ class MultiStep(sweeper):
         The first element in the coefficients belongs to the value furthest in the past and vice versa. Values from previous time steps are stored in a `Cache` object.
         Be careful with the sign of the alpha values. You can look at the implementations of the Euler methods for guidance.
 
-        Args:
-            params (dict): parameter object
+        Class attributes:
             alpha (list): Coefficients for the solutions of previous steps
             beta (list): Coefficients for the right hand side evaluations
+
+        Args:
+            params (dict): parameter object
         """
         import logging
         from pySDC.core.Collocation import CollBase
@@ -89,10 +94,8 @@ class MultiStep(sweeper):
         self.parallelizable = False
 
         # proprietary variables for the multistep methods
-        self.steps = len(alpha)
+        self.steps = len(self.alpha)
         self.cache = Cache(self.steps)
-        self.alpha = alpha
-        self.beta = beta
 
     def predict(self):
         """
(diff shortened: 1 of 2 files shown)


pancetta:
So, this is multi-step methods within a single time-step, yes? I'd not call it multi-step methods then,  because people would expect something else under this name.

brownbaerchen:
> So, this is multi-step methods within a single time-step, yes? I'd not call it multi-step methods then, because people would expect something else under this name.

No, it's multistep methods. It stores the solutions from past steps and uses these to compute a solution to the current step. At least in the book I copied these methods from they referred to these methods as multistep. Of course every but the two step Adams-Moulton method are very boring examples of multistep methods, but I believe taxonomy allows to call single-stage one-step methods either one-step or multistep.

pySDC/implementations/sweeper_classes/Multistep.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def __init__(self, params):
+        alpha = [-1.0]
+        beta = [1.0, 0.0]
+        super().__init__(params, alpha, beta)
+
+
+class BackwardEuler(MultiStep):
+    """
+    Do you like mess? Me neither! Let me lure you in!
```
pancetta, line 177: Yeah.. nice try.

pancetta:
Ok, ok, I see. Not sure I get it, yet: You store the solution of how many previous steps and how many previous nodes? Is it: last step, `len(alpha)` nodes? Or only the last node of the `len(alpha)` steps?

brownbaerchen:
> Ok, ok, I see. Not sure I get it, yet: You store the solution of how many previous steps and how many previous nodes? Is it: last step, `len(alpha)` nodes? Or only the last node of the `len(alpha)` steps?

I only store the solutions to previous steps, not any of the nodes. That would also be interesting, but this PR is just about getting more established methods for comparisons into pySDC. Lisa expressed some interest in using this for her DAE project, so I went ahead and did this simple stuff.
Actually, there should not be any particular nodes. Just one for initial conditions and one for the solution of the step. But I did forget to fix this...

As always, I don't know anything about multistep methods, but I followed my favourite book, the Leveque finite difference book, and that had the convention: N alpha values for the previous solutions and N+1 beta values for the right hand side evaluations at the previous solutions and at the current one.

I should maybe elaborate on the documentation...

So I have two todos now:
- [ ] More documentation on alpha and beta parameters
- [ ] Fix the "collocation" to single node

Anything else?

pySDC/implementations/sweeper_classes/Multistep.py
```diff
@@ -0,0 +1,219 @@
+from pySDC.core.Sweeper import sweeper, _Pars
+
+
+class Cache(object):
```
lisawim, line 4: That's cool! I like the class, pretty good idea! So you wrote a class having the values from previous steps and can update them using the `update` method? I was thinking way too complicated when I was implementing that..

brownbaerchen: Yes, but this is actually a bit too simple. I know both of us want to restart time steps and that is not really supported with this simple cache. But we can add that later. For instance, we can keep one extra solution in the cache and do a reverse `update` if we restart a step. 
We should sit down while we're here at Hamburg at some point so you can show me what you already did and we can merge our efforts.

pancetta:
So, what's the status of this?

lisawim:
After some brainstorming with @brownbaerchen in Hamburg starting values for the LM methods are not always available and in case of using more than one previous time step to start with, starting values has to be generated by another method, e.g. RK methods. This seems to be not implemented yet, but I'm not sure if he wants to add this here. Please @brownbaerchen correct me if I'm wrong.

brownbaerchen:
Of course @tlunet's input is always appreciated! But don't feel obligated if you don't have the time.

I intended this is as a starting point if someone ever really wants to use LMMs. So far, the methods are of limited use. I guess the third order implicit two step method is not too bad, as long as you have fixed step size and no restarts, which has been the only way of running the code for a long time. #331 made it easy to access Butcher tableaux outside of RK sweepers. Now, all that needs to be done to add LMMs of order larger than 3 is to basically copy-paste the `update_nodes` function from the RK sweeper and plug in the Q matrices from the RK methods. I did not plan on doing this currently, because I am not sure that anybody will be using it.

Of course, @lisawim wants to use LMMs, but I was unable to get high order in time with the simple DAE problem from @kyrillh without putting any sort of thought into this. So I guess it makes more sense to move forward with this when we know how to do DAEs with RK.
@lisawim and I did almost the same thing apart from her doing DAEs and me ODEs. However, I tested that the LMMs do not perform unnecessary right hand side evaluations. If it's fine with @lisawim, we could merge this and you could inherit from this, only changing the `update_nodes` function to DAE stuff and implementing the coefficients?

lisawim:
That's a good idea, I'm completely fine with that! I'll also have to see what's going on with the order of accuracy..

Issue #286, opened 2023-03-24
Improve Problem classes Documentation


tlunet:
Currently problem class constructors explicitly define in their signature the problem parameters, for instance :

```python
class acoustic_1d_imex(ptype):
    """
    Example implementing the one-dimensional IMEX acoustic-advection

    TODO : doku
    """

    dtype_u = mesh
    dtype_f = imex_mesh

    def __init__(self, nvars, cs, cadv, order_adv, waveno):
        # ... implementation ...
``` 

But only a few of the problem classes have a fully detailed documentation describing the problem parameters (for an example, see the [advectionNd class](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py)).

When working with a given problem, it would be good to also add in the same time the documentation for its class, following the [documentation guidelines](https://github.com/Parallel-in-Time/pySDC/blob/master/docs/contrib/05_documenting_code.md).

> **Note** : for the problem class, since parameter are automatically stored as attributes (following the [contribution guidelines](https://github.com/Parallel-in-Time/pySDC/blob/master/docs/contrib/04_custom_implementations.md#implementing-a-custom-problem-class)), it is then not necessary to describe the parameters a second time in the `Attributes` doc section. For example :

```python
class AdvectionSpectral(ptype):
    r"""
    This class implement the advection problem

    .. math::
        \frac{\partial u}{\partial t} = -c\frac{\partial u}{\partial x},

    on :math:`[0,1]^N` using periodic boundary conditions and spectral Fourier
    discretization to estimate the space derivative term.

    Parameters
    ----------
    nvars : int of tuple, optional
        Spatial resolution (same in all dimensions). Using a tuple allows to
        consider several dimensions, e.g nvars=(16,16) for a 2D problem.
    c : float, optional
        Advection speed (same in all dimensions).

    Attributes
    ----------
    A : np.ndarray
        Discretization matrix for the :math:`\frac{\partial u}{\partial x}`
        term (dense).
    """

    dtype_u = mesh
    dtype_f = imex_mesh

    def __init__(self, nvars, c):
        super().__init__((nvars, None, np.dtype('float64')))
        self._makeAttributeAndRegister('nvars', 'c', localVars=locals(), readOnly=True)
        self.A = generateSpectralFourierMatrix(nvars)
        self.A += -c
```

[...]

Issue #223, opened 2022-12-21
Better tests for problem classes


pancetta:
- [ ] [`Advection_ND_FD`](https://parallel-in-time.org/pySDC/coverage/d_91faa57f8583c837_AdvectionEquation_ND_FD_py.html): Dirichlet and higher dimensions 
- [ ] [`AllenCahn_MPIFFT`](https://parallel-in-time.org/pySDC/coverage/d_91faa57f8583c837_AllenCahn_MPIFFT_py.html): IMEX and `rand_circle` (do we need IMEX here?)
- [ ] [`AllenCahn_Temp_MPIFFT`](https://parallel-in-time.org/pySDC/coverage/d_91faa57f8583c837_AllenCahn_Temp_MPIFFT_py.html): `rand_circle` (what about IMEX here?)
- [ ] [`HarmonicOscillator`](https://parallel-in-time.org/pySDC/coverage/d_91faa57f8583c837_HarmonicOscillator_py.html): init and damping
- [ ] [`HeatEquation_1D_FEniCS_weak_forced`](https://parallel-in-time.org/pySDC/coverage/d_91faa57f8583c837_HeatEquation_1D_FEniCS_weak_forced_py.html): `fenics_heat_weak_fullyimplicit` not tested whatsoever
- [ ] [`HeatEquation_ND_FD`](https://parallel-in-time.org/pySDC/coverage/d_91faa57f8583c837_HeatEquation_ND_FD_py.html): Gaussian and higher dimensions
- [ ] [`AcousticAdvection_1D_FD_imex`](https://parallel-in-time.org/pySDC/coverage/d_c032a2f084ee6f96_buildFDMatrix_py.html): all those helper classes (ugh!)

PR #333, opened 2023-07-23
Rework of documentation in problem classes

lisawim:
I reworked the problem classes (#286). I wrote some more detailed documentation for some problem classes. For all classes I also reworked the documentation for the class methods. Unfortunately, there is still much to do, but I think we are much closer to finish the issue #286. 

I have no knowledge about many of the problems.. @tlunet told me documentation for problem classes can be reworked if someone will work with that but I guess some problem classes will be untouched in the near future. I could do the rest but I need references for the problems to finish up the documentations to write down the equations for them..


pySDC/implementations/problem_classes/AdvectionDiffusionEquation_1D_FFT.py
```diff
@@ -7,50 +7,60 @@
...
-            if key not in problem_params:
-                msg = 'need %s to instantiate problem, only got %s' % (key, str(problem_params.keys()))
-                raise ParameterError(msg)
+        # invoke super init, passing number of dofs, dtype_u and dtype_f
+        super().__init__(init=(nvars, None, np.dtype('float64')))
+        self._makeAttributeAndRegister('nvars', 'c', 'freq', 'nu', 'L', localVars=locals(), readOnly=True)
 
         # we assert that nvars looks very particular here.. this will be necessary for coarsening in space later on
-        if (problem_params['nvars']) % 2 != 0:
+        if (self.params.nvars) % 2 != 0:
```
tlunet, line 61: Are you sure this works ? `params` is supposed to be a dictionnary only now ...

tlunet: PS : what was said during the pySDC workshop was actually wrong :sweat_smile:

tlunet:
Nice work !

tlunet:
Seems like the pySDC/implementations/problem_classes/AdvectionDiffusionEquation_1D_FFT.py was not tested ... I've added this to the list of problem not tested yet in #223

pancetta:
Great work, thanks a lot for this!

PR #337, opened 2023-07-25
TL: added MIN-SR implicit sweepers

tlunet:
Following some ongoing work with @caklovicka, I've added the new diagonal SDC preconditionners :

- MIN-SR-NS : diagonal $Q_{\Delta}$ minimizing the spectral radius of the non-stiff limit of the SDC iteration matrix, i.e $Q-Q_{\Delta}$
- MIN-SR-S : diagonal $Q_{\Delta}$ minimizing the spectral radius of the stiff limit of the SDC iteration matrix, i.e $I-Q_{\Delta}^{-1}Q$

Actually, the `MIN-SR-NS` was already implemented under the name `MIN_GT`, so `MIN-SR-NS` an alias.
I've added automated tests that verify that the spectral radius is numerically zero by checking if the $Q-Q_{\Delta}$ and $I-Q_{\Delta}^{-1}Q$ are nilpotent to the power $M$.

The whole theory behind it is in progress (but you can already check that in the https://github.com/tlunet/why repo). For now, the non-stiff limit is obtained "exactly", while the stiff limit is obtained using a mix between an analytic and a numerical approach. The later still need some improvements to go over $M=5$, but in the meantime those preconditionners can be tested for different problems and arbitrary nodes, if some others want to play a bit with it ... @pancetta @brownbaerchen @lisawim

### DIFF
--- a/pySDC/tests/test_sweeper.py
+++ b/pySDC/tests/test_sweeper.py
@@ -1,0 +1,46 @@
+import numpy as np
+import pytest
+
+from pySDC.core.Sweeper import sweeper as Sweeper
+
+node_types = ['EQUID', 'LEGENDRE'] + [f'CHEBY-{i}' for i in [1, 2, 3, 4]]
+quad_types = ['GAUSS', 'LOBATTO', 'RADAU-RIGHT', 'RADAU-LEFT']
+
+
+@pytest.mark.base
+@pytest.mark.parametrize("node_type", node_types)
+@pytest.mark.parametrize("quad_type", quad_types)
+def test_MIN_SR(node_type, quad_type):
+    for M in [2, 3, 4, 5]:
+        params = {'num_nodes': M, 'quad_type': quad_type, 'node_type': node_type}
+
+        sweeper = Sweeper(params)
+        Q = sweeper.coll.Qmat[1:, 1:]
+
+        # Check non-stiff limit
+        QDelta = sweeper.get_Qdelta_implicit(sweeper.coll, 'MIN-SR-NS')[1:, 1:]
+        assert np.all(np.diag(np.diag(QDelta)) == QDelta), "no diagonal QDelta"
+        K = Q - QDelta
+        Km = np.linalg.matrix_power(K, M)
+        nilpotency = np.linalg.norm(Km, ord=np.inf)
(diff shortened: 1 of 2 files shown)


pySDC/tests/test_sweeper.py
pancetta, line 1: Want to give it a less generic name here? E.g. `test_MIN_sweeper` or so?

pancetta (changes requested):

brownbaerchen:
LGTM. The stiff limit looks complicated. I couldn't find anything about it in the why repository. Do you have some notes on that? I will play around with this for the next project which is estimating the error for adaptivity of converged collocation problems. Then the convergence properties of the preconditioner don't matter to the step size selection and a good diagonal preconditioner would be great!

pySDC/tests/test_sweeper.py
tlunet: Other QDelta are not standalone tested (like `BEpar`, `LU`, etc ...). Could be done here actually ?

tlunet:
> LGTM. The stiff limit looks complicated. I couldn't find anything about it in the why repository. Do you have some notes on that? I will play around with this for the next project which is estimating the error for adaptivity of converged collocation problems. Then the convergence properties of the preconditioner don't matter to the step size selection and a good diagonal preconditioner would be great!

There is no notes on this, but you can already look at the associated scripts (fact7) : https://github.com/tlunet/why/tree/gaia/scripts/fact7

pySDC/tests/test_sweeper.py
tlunet: Can I keep the generic name if this file also tests `LU`, `Qpar`, `BE`, `BEpar` and `PIC` ?

pancetta: Absolutely. I just wanted to leave you the choice!

pancetta (approved):

pySDC/tests/test_sweeper.py
tlunet: There you go ... there is just some weird edge case for `LU` with `M > 3 and node_type == 'EQUID' and quad_type in ['GAUSS', 'RADAU-RIGHT']`, that may need a bit more theoretical understanding. But I've put a TODO for now, since it's not very that critical ...

lisawim:
I now I'm a bit late with that. Nice work! I'm looking forward to playing around with that hopefully in the near future!

PR #338, opened 2023-07-25
Brand-new switch estimator + new discontinuous ODE model!

lisawim:
Finally, I finished this PR to update the switch estimator (SE) which is improved a lot now! I've already presented the new framework of the SE in Hamburg. 

Furthermore I added the discontinuous scalar test ODE which I also presented there. It is tested in a hopefully comprehensive way.

As result of the SE's big update I need to adapt all battery model tests.. which took a lot time. The reason was that there are some plots based on the event time.. If the SE founds a potential event but this event does not satisfy the tolerance the event time is not logged. In any case (it doesn't matter if the event time satisfy the tolerance or not) it is restarted using a new time step size. Sometime those many restarts leads to a very good resolution of the event at the end of the time step and the SE stops to looking ahead.. But here an event is still not logged..
I adapted that by logging the event when the event is already resolved well. 

What do you think? @pancetta @brownbaerchen @tlunet ? I am very confused if there are no comments on that ;)


pySDC/implementations/problem_classes/DiscontinuousTestODE.py
```diff
@@ -0,0 +1,216 @@
...
+from pySDC.core.Problem import ptype
+from pySDC.implementations.datatype_classes.mesh import mesh
+
+
+class DiscontinuousTestODE(ptype):
+    r"""
+    This class implements a very simple test case of a ordinary differential equation consisting of one discrete event. The dynamics of
+    the solution changes when the state function :math:`h(u) := u - 5` changes the sign. The problem is defined by:
+
+    if :math:`u - 5 < 0:`
```
tlunet, line 13: Why this choice of $5$ for the switch ? Cannot it be a parameter of the problem ?

pySDC/implementations/problem_classes/DiscontinuousTestODE.py
```diff
@@ -0,0 +1,216 @@
...
+    dtype_u = mesh
+    dtype_f = mesh
+
+    def __init__(self, nvars=1, newton_maxiter=200, newton_tol=1e-8):
+        """Initialization routine"""
+
+        super().__init__(init=(nvars, None, np.dtype('float64')))
+        self._makeAttributeAndRegister('nvars', 'newton_maxiter', 'newton_tol', localVars=locals(), readOnly=True)
+
+        if self.nvars != 1:
```
tlunet, line 49: Don't put `nvars` as argument, if it's only value can be one. But you can let it as read-only parameter so it can be used later fo the data initialization (but since it's stored already in `init`, I'm not sure that you need it as problem parameter)

pySDC/implementations/problem_classes/DiscontinuousTestODE.py
```diff
@@ -0,0 +1,216 @@
...
+            raise ParameterError('nvars has to be equal to 1!')
+
+        self.t_switch_exact = np.log(5)
```
tlunet, line 52: `t_switch_exact` can be put as read-only parameter to avoid any accidental modification during run-time

tlunet:
Nice work ! I've put a few comments on the discontinuous scalar test ODE problem class ...

pySDC/implementations/problem_classes/DiscontinuousTestODE.py
lisawim: It is chosen arbitrarily since it is a self-constructed problem. The exact event time and also everything based on this value.

pySDC/implementations/problem_classes/DiscontinuousTestODE.py
```diff
@@ -0,0 +1,216 @@
...
+        """
+
+        t_switch = np.inf if self.t_switch is None else self.t_switch
+
+        h = rhs[0] - 5
+        u = self.dtype_u(u0)
+
+        n = 0
+        res = 99
+        while n < self.newton_maxiter:
```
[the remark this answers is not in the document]
lisawim: Ok, will change that!

pySDC/implementations/problem_classes/DiscontinuousTestODE.py
lisawim: Ok, will also change that!

lisawim:
Many thanks for your comments, I appreciate that! I'll update it, but anyway I has to to that due to the failed test.

lisawim:
I'm a tiny bit confused.. I saw that there is an `AssertionError` due to the hardcoded solutions does not match with the solution. But locally the test passed.. I also updated all my packages.. it's still passing.

pySDC/projects/PinTSimE/discontinuous_test_ODE.py
```diff
@@ -0,0 +1,344 @@
...
+    msg = 'Newton tolerance should be set as small as possible to get best possible resolution of solution'
+    assert newton_tol <= 1e-8, msg
+
+    assert t0 >= 1.0, 'Problem is only defined for t >= 1!'
+    assert Tend >= np.log(5), f'To investigate event, please set Tend larger than {np.log(5):.5f}'
+
+    num_nodes = description['sweeper_params']['num_nodes']
+    for M in [2, 3, 4]:
+        if num_nodes not in [2, 3, 4]:
+            assert num_nodes == M, f'Hardcoded solutions are only for M={M}!'
```
brownbaerchen, line 331: Why the loop over `M`?

pySDC/projects/PinTSimE/estimation_check.py
```diff
@@ -297,17 +324,15 @@ def differences_around_switch(
         times_adapt = [me[0] for me in get_sorted(stats_adapt, type='u', recomputed=False)]
         times_SE_adapt = [me[0] for me in get_sorted(stats_SE_adapt, type='u', recomputed=False)]
 
-        diffs_true_at.append(
-            [diff_SE[m] for m in range(len(times_SE)) if np.isclose(times_SE[m], t_switch, atol=1e-15)][0]
-        )
+        diffs_true_at.append([diff_SE[m] for m in range(len(times_SE)) if abs(times_SE[m] - t_switch) <= 1e-17][0])
```
brownbaerchen, line 327: testing if the difference between two floats is less than machine precision seems fishy too me.

pySDC/projects/PinTSimE/estimation_check.py
```diff
@@ -297,17 +324,15 @@ def differences_around_switch(
...
 
         diffs_false_before.append(
             [diff[m - 1] for m in range(1, len(times)) if times[m - 1] <= t_switch <= times[m]][0]
         )
         diffs_false_after.append([diff[m] for m in range(1, len(times)) if times[m - 1] <= t_switch <= times[m]][0])
 
         for m in range(len(times_SE_adapt)):
-            if np.isclose(times_SE_adapt[m], t_switch_SE_adapt, atol=1e-10):
+            if abs(times_SE_adapt[m] - t_switch_SE_adapt) <= 1e-16:
```
brownbaerchen, line 335: This is also below machine precision.

brownbaerchen (commented):
I just glanced over the code and looks good for the most part. You did a very thorough job on documentation, which is really good!

Of course I was nitpicking a little bit. There are mainly three small things I noticed:
 - I am wondering about the parameters you choose for resolution in time. You perform 12 iterations with only up to 4 nodes. Why 12? This goes beyond the order achievable by the collocation problem and if you want to solve it exactly, maybe use a residual tolerance instead of fixed number of iterations?
 - For counting events we have the `WorkCounters` now. The advantage of these is that they can be passed as a callback to `numpy` functions as well as being used wherever and they provide a common interface.
 - We have a new implementations folder with generic hooks now. I see you have hooks for logging the solution and the error, but it is favourable to use the generic implementations to streamline the interface across projects as much as possible and to reduce copy-pasting code.

I did not look into the tests in too much detail, but I believe you mainly test that the results of simulations stay the same in the future. I did that too, but recently I started to test also individual functionality of things that I was implementing and it turns out I missed many bugs. By testing that the result of simulations using this stuff stayed the same, I made sure the implementation would not change, but I did not make sure it was correct. 
Do you have a test that checks that the switch estimator is doing exactly what you expect it to irrespective of any particular simulation? If not, I think that would be really great. But that's also independent of this PR, I guess.

pySDC/implementations/problem_classes/DiscontinuousTestODE.py
```diff
@@ -0,0 +1,217 @@
...
+            # form function g with g(u) = 0
+            if h >= 0 or t >= t_switch:
+                g = u - dt * (4 / self.t_switch_exact) - rhs
```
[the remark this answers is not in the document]
lisawim: No, it's not cheating! ;) This results from the construction of the problem. The problem knows the exact event time, yes. But you can still see not using the switch estimator that the error grows after this time.

pySDC/projects/PinTSimE/battery_2capacitors_model.py
```diff
@@ -33,19 +66,43 @@ def run():
 
     problem_classes = [battery_n_capacitors]
     sweeper_classes = [imex_1st_order]
+    num_nodes = 4
+    restol = -1
+    maxiter = 12
```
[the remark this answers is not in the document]
lisawim: I'm also very confused here. I wanted to use a residual tolerance to reach a stop either the residual drops below the tolerance or after reaching `maxiter`. But setting e.g. `restol = 1e-15` leads to no iterations, i.e., nothing is done because `L.status.residual=0.0` is set somewhere.. Do I have to set something using this `skip_residual_computation`? But this does not make any sense, because the discontinuous test ODE problem works with setting an residual tolerance.. So confusing.. it feels like pySDC and everything else is hating me since yesterday..

pySDC/projects/PinTSimE/discontinuous_test_ODE.py
```diff
@@ -0,0 +1,344 @@
...
+            time=L.time + L.dt,
+            level=L.level_index,
+            iter=0,
+            sweep=L.status.sweep,
+            type='state_function',
+            value=L.uend[0] - 5,
+        )
+
+
+class LogSolution(hooks):
```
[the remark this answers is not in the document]
lisawim: Ah yes, sure! That's true.. thanks for the hint!

pySDC/projects/PinTSimE/estimation_check.py
lisawim: I agree, it's fishy.. it also works with a larger tolerance.

lisawim:
Yes, it's weird. Sometimes the tests passes sometimes not, also locally now! I also think that it depends on the roundings..

pySDC/projects/PinTSimE/battery_2capacitors_model.py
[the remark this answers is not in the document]
brownbaerchen: Hmmmm. Is this for an RK sweeper or something like that? The residual computations should not be skipped by default, but if you do skip it, it will be initialised to 0. Sounds like this is indeed the issue.
I hope pySDC and you can make up soon 🥴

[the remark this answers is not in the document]
lisawim: No, I don't want to skip that! This is why I am confused I want to use the residual tolerance. But it does not work. Probably the problem is sitting in front of the laptop.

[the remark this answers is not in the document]
brownbaerchen: The problem is probably sitting in front of my laptop. What's the output when you decrease the logger level? What do you get for `skip_residual_computation` in the sweeper params and what does the code say about the residual if it is computing any?

brownbaerchen:
> 

Wait a minute. You do the same test locally multiple times and sometimes it passes, sometimes not? But you don't have any random stuff in your code, right? What the hell is going on? Did you anger the machine god somehow? :D

pySDC/projects/PinTSimE/battery_2capacitors_model.py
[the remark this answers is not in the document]
lisawim: I set `logger_level=15` so it should output something. But there is not output. The initial output after initialization of the controller states `skip_residual_computation = ()`. I can see no computation. I was printing out `L.status.residual` in check_convergence.py and it is set to 0.0. Hence, no computation because we are "converged".

lisawim:
I don't use random stuff. Otherwise this would be an explanation for that. But no, only "exactly" stuff that should be going on.

lisawim:
Ooops..

pySDC/projects/PinTSimE/battery_2capacitors_model.py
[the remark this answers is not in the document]
lisawim: The residual for both battery models is zero already before start of computation. The reason why nothing is happening.. Now I remember why I'm using `sweeper_params['initial_guess'] = 'zero'` when I started to implement the battery models one year ago. :D But this not really explain why there is no logging output, or maybe yes? ...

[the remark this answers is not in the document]
brownbaerchen: I started a PR on your fork with a fix for this. The issue is not the initial guess, but how the `A` matrix is initialised. Did you not get an email about the PR?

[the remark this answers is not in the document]
lisawim: No email there.. Thanks for the PR!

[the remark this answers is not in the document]
brownbaerchen: By the way: The built in python debugger proved to be quite useful in debugging this. @kyrillh was making fun of me until I finally gave it a try and it's actually pretty good! You can just write 
```
breakpoint()
```
in the code and when you execute it, the code will stop there. Now you see the output generated before and have the option to manipulate everything. You move on by doing `c` for continue. This is very handy when the code does many iterations of the same thing. If I just print, I have to scroll around like crazy in the end. If I do `raise` after the print as I used to, I can't see what happens in the next iteration. You can exit the debugger with `ctrl-D`
In this case, I printed the `A` matrix in `eval_f` and had a breakpoint in `compute_residual` then I noticed the matrix is 0 in the first iteration and non-zero later.
Maybe you know about this already, but I believe the built-in debugger is under-appreciated. I believed myself that debugging is reserved for fancy IDEs until very recently. But luckily there is no need to give in to IDEs ;)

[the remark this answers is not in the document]
lisawim: Hey, that's very cool! I just used that and it should be very helpful! Since now I used many print lines for debugging, very annoying.. But now I can do my life a bit easier with using an editor. :D

lisawim:
Lets see if the tests passes now..

pySDC/projects/PinTSimE/discontinuous_test_ODE.py
```diff
@@ -11,6 +11,8 @@
 from pySDC.projects.PinTSimE.battery_model import get_recomputed, generate_description
 import pySDC.helpers.plot_helper as plt_helper
 from pySDC.core.Hooks import hooks
+from pySDC.implementations.hooks.log_errors import LogLocalErrorPostStep
```
brownbaerchen, line 14: Ok, I see that this is poorly implemented on my part. The local error requires an exact solution starting from the beginning of the step, not from t=0. That's why some of the `u_exact` functions have the `u_init` and `t_init` variables. However, if they are not used, the `LogLocalErrorPostStep` function logs the global error without telling you. Maybe we should put in warnings whenever this happens...

pySDC/projects/PinTSimE/estimation_check.py
```diff
@@ -69,7 +72,7 @@ def run(cwd='./'):
         for dt_item in dt_list:
             for use_SE in use_switch_estimator:
                 for use_A in use_adaptivity:
-                    tol_event = 1e-10
+                    tol_event = 1e-10 if problem.__name__ == 'generic_implicit' else 1e-17
```
brownbaerchen, line 75: Is this meant to be `sweeper.__name__`?

pancetta:
LGTM now. @brownbaerchen?

brownbaerchen (commented):
The tolerances look much better now, I think! Nice! I just have some minor comments regarding the hooks.

How did you manage that the tests now pass? Do you know what happened? Are you sure they are not flaky anymore?

pySDC/implementations/problem_classes/DiscontinuousTestODE.py
```diff
@@ -0,0 +1,221 @@
...
+
+        self.newton_ncalls += 1
+        self.newton_itercount += n
+
+        me = self.dtype_u(self.init)
+        me[:] = u[:]
+
+        return me
+
+    def u_exact(self, t, u_init=None, t_init=None):
```
brownbaerchen, line 153: Maybe you can get a logger for the problem and raise a warning when `u_init` or `t_init` are not None. I should add something like this to other problems with analytical exact solution because all of them will return the global error when using the local error hook without telling you about it.

pySDC/projects/PinTSimE/battery_model.py
```diff
@@ -7,47 +7,33 @@
...
+from pySDC.implementations.hooks.log_solution import LogSolution
+from pySDC.implementations.hooks.log_step_size import LogStepSize
 
 from pySDC.projects.PinTSimE.switch_estimator import SwitchEstimator
 from pySDC.implementations.convergence_controller_classes.adaptivity import Adaptivity
+from pySDC.implementations.convergence_controller_classes.basic_restarting import BasicRestartingNonMPI
 
 
-class LogData(hooks):
+class LogErrorEmbeddedEstimate(hooks):
```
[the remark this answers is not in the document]
lisawim: Oh, for some unknown reason I didn't found that yet. Thanks, I'll add this!

pySDC/projects/PinTSimE/estimation_check.py
lisawim: Ah, yes sure!

pySDC/implementations/problem_classes/DiscontinuousTestODE.py
lisawim: Will add that!

lisawim:
As you can see the SE have many different tolerances. I've set almost all tolerances larger in order to get away from rounding errors. But I didn't find out yet why the machine precision god wants to annoy me..

pySDC/projects/PinTSimE/discontinuous_test_ODE.py
lisawim: Hm, I would add a warning now in the discontinuous test problem class, but in this case a warning does make sense for all. How would you add the warning?

lisawim: I can also put a warning this in the other problem classes where t_init and u_init is set the parameters.

brownbaerchen: I added it to the advection, heat and nonlinear Schrödinger equations, but I am waiting for the tests to pass before I start the PR.
[here](https://github.com/brownbaerchen/pySDC/blob/fa2fa12962989a6b4a1a4c5857e2cd0888b861e6/pySDC/implementations/problem_classes/AdvectionEquation_ND_FD.py#L100) is the warning I added. It doesn't need to be too fancy, I guess. Note that in the function definition, are only the variables that are used in the function and then `**kwargs` to allow for additional keyword arguments. I don't know if this is the pythonic way, but I think this is nice because you can more easily tell by the function definition what arguments are allowed.
If you pass on the keyword arguments to another function, a good editor will show the allowed arguments for the both functions when you look up the documentation for the "outer" function. Although I have to admit my editor struggles with this sometimes. ;)

lisawim: Of course the `kwargs**` argument is very useful! I would add a warning for `AllenCahn_2D_FFT`, `DiscontinuousTestODE`, `Lorenz`, `Piline`, `Quench`, `TestEquation_0D` and `Van_der_Pol_implicit`, which I already did but will update the warning.

brownbaerchen: Are you sure non of them can compute the local error? I looked up the test equation and even though it's analytic, it can compute the local error. All other problems you mentioned rely on numerical exact solutions, I believe, which should support the local error.

pySDC/implementations/problem_classes/Piline.py
```diff
@@ -137,6 +137,9 @@ def u_exact(self, t, u_init=None, t_init=None):
             The reference solution.
         """
 
+        if t_init is not None and u_init is not None:
```
brownbaerchen, line 140: This problem does support computing the local error.

brownbaerchen (changes requested):

brownbaerchen (approved):

brownbaerchen:
@pancetta Do you want to try squashing while merging here? :D

brownbaerchen:
Wait a minute! You are still trying to record the local error with a problem that does not support this. Are you sure you want to do this? If you want to use this also for other problems and you manage to keep track of this, that may be ok, I guess

lisawim:
I'm sorry I'm still confused why the `DiscontinuousTestODE` does not support the possibility of computing the local error. The exact solution is available. And taking a look into the `LogLocalErrorPostStep` hook class states that the value for the error is

```
err = abs(L.prob.u_exact(t=L.time + L.dt, u_init=L.u[0] * 1.0, t_init=L.time) - L.uend)
```

Do I miss something?

brownbaerchen:
> I'm sorry I'm still confused why the `DiscontinuousTestODE` does not support the possibility of computing the local error. The exact solution is available. And taking a look into the `LogLocalErrorPostStep` hook class states that the value for the error is
> 
> ```
> err = abs(L.prob.u_exact(t=L.time + L.dt, u_init=L.u[0] * 1.0, t_init=L.time) - L.uend)
> ```
> 
> Do I miss something?

You can compute the global error there, but not the local error. For the local error, you have to use the initial conditions of the step as the initial conditions for the exact solution. In this problem, you have only the exact solution starting from $t=0$ which will give the global error. Right? You add something to the exact solution that can start from later than $t=0$?

lisawim:
No, only the battery models have an exact solution for $t=0$, but `Discontinuous TestODE` also provides an exact solution for $t\geq 0$.

brownbaerchen:
The global error is what you get when you compare the numerical solution at time $t$ to the exact solution when taking the initial conditions from $t=0$ at time $t$. This you can compute using your function.
The local error is what you get when you compare the numerical solution at time $t$ to the exact solution when taking the same initial conditions as the time stepping scheme for the particular step at time $t$. These are not the exact solution at $t-\Delta t$, but the numerical solution at $t-\Delta t$. That is what you put into the `u_exact` function in the local error hook.
But your test problem can only give the exact solution at $t$ starting from the initial conditions at $t=0$. It cannot generate an exact solution at some later time from any numerical intermediate solution it does not use the `u_init` or `t_init` arguments.
Therefore, you can only compute the global error for this problem, not the local one. Hence the warning that if you try to do this, you get something different that what you asked for. That's why you should use the hook for logging the global error instead of the one for logging the local one.

lisawim:
Ok, many thanks for the detailed explanation! I would like to keep the one with the global error.

pancetta:
@brownbaerchen @lisawim Is there anything left to change here?

brownbaerchen:
> Ok, many thanks for the detailed explanation! I would like to keep the one with the global error.

Then please replace the hook that logs the local error with the one that logs the global error in line 14 of [pySDC/projects/PinTSimE/discontinuous_test_ODE.py](https://github.com/Parallel-in-Time/pySDC/pull/338/files/d6055c092452f22934ef9ffcf1ddbdbc7f9d8e71..fabaa90b06ceb0e10e397b8e85b4fa5844d015da#diff-f3c869734a8077c33f37a270d4a4ca659eeeffbc9bc363bdb9e36465984ab076)

PR #339, opened 2023-07-27
Added warnings when trying to compute local error with analytic exact solution

brownbaerchen:
There is a hook in the implementations that records the local error by generating exact solutions starting from the initial conditions and initial time of the step. However, many problems have an interface that is compatible with the hook, but always compute the exact solution from $t=0$, for instance when an analytic exact solution is available.

I added some warnings when this happens. It's not perfect, but it's better than nothing. If I forgot any problems, feel free to add the warnings yourself or tell me about it and I can do it.


brownbaerchen:
@pancetta Do you want to try squashing while merging?

PR #342, opened 2023-08-07
Fully implicit nonlinear Schrödinger

brownbaerchen:
Simple fully implicit version of nonlinear Schrödinger based on a `scipy` Jacobian-free Newton-Krylov solver. There is a test that this version does more or less the same as the IMEX version.

There is also a small change to the sweeper class which allows to use an explicit preconditioner for `QI`. This will warn you that you are doing something which is sort of stupid. But I find it useful for quick testing. That is to say when I want to use an explicit preconditioner, I can continue to use the implicit sweepers. If you don't like it, we can remove it.

### DIFF
--- a/pySDC/core/Sweeper.py
+++ b/pySDC/core/Sweeper.py
@@ -296,7 +296,13 @@ class sweeper(object):
             self.parallelizable = True
 
         else:
-            raise NotImplementedError(f'qd_type implicit "{qd_type}" not implemented')
+            # see if an explicit preconditioner with this name is available
+            try:
+                QDmat = self.get_Qdelta_explicit(coll, qd_type)
+                self.logger.warn(f'Using explicit preconditioner \"{qd_type}\" on the left hand side!')
+            except NotImplementedError:
+                raise NotImplementedError(f'qd_type implicit "{qd_type}" not implemented')
+
         # check if we got not more than a lower triangular matrix
         np.testing.assert_array_equal(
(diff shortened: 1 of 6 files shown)


pySDC/core/Sweeper.py
```diff
@@ -296,7 +296,13 @@ def func(coeffs):
             self.parallelizable = True
 
         else:
-            raise NotImplementedError(f'qd_type implicit "{qd_type}" not implemented')
+            # see if an explicit preconditioner with this name is available
+            try:
+                QDmat = self.get_Qdelta_explicit(coll, qd_type)
+                self.logger.warn(f'Using explicit preconditioner \"{qd_type}\" on the left hand side!')
+            except NotImplementedError:
+                raise NotImplementedError(f'qd_type implicit "{qd_type}" not implemented')
```
pancetta, line 304: How can you make sure that the sweeper respects this? There is no `QE` in the generic implicit sweeper, so this will go unnoticed, right?

brownbaerchen: This is the function for getting `QI` and if you request for instance `EE` as `QI`, that is weird because `EE` is explicit, of course. But the `generic_implicit` sweeper will still work with an explicit preconditioner, it will just solve the systems like `(1-0)u=rhs`, which gives the correct explicit result, as far as I know. Does that answer your question?

pancetta: It does, indeed.

PR #343, opened 2023-08-08
Upper case mpi4py communication in convergence controllers

brownbaerchen:
I am starting to merge back some of the changes I have in a development branch of mine, so expect some PRs in the upcoming days with random small things. This is one of them.

I ran into issues with the pipeline on GitHub behaving differently from my local machine and going to upper case communication in mpi4py solved that. It is a few more lines, but also should be faster because it does not pickle everything before sending because you have to specify a buffer yourself. There is still the option to do the lower case stuff and I believe it is still used in some places as well.
The tests for which this was relevant are not part of this PR and I don't actually know if the issue persists or has been solved by some updates. Anyways, I guess forgoing the pickling is enough to merit doing this.

The second small change is a different initial guess for RK methods. Instead of the initial conditions, they now use the last stage as recommended by Daniel.


pancetta:
Could you please encapsulate the additional lines of code in the `send` and `recv` functions? I just cleaned up the controller a bit to make it more readable (hahahaha)..

brownbaerchen:
I don't think I can because different convergence controllers will need different buffers. And the same convergence controller will need different buffers when communicating different things. The controller, I believe, only communicates the solution so there everything is the same, but here this is not the case.
I could overload the functions in some convergence controllers to streamline it a bit, but honestly I am afraid that would be even more mess.
Maybe I misunderstood and you can elaborate?

pancetta:
No, no, I understand. Since you're not touching the controller directly, I'm ok with this.

brownbaerchen:
I am excited to see the new clean controller by the way! :D

pancetta:
There is none besides the one that is already in the master branch. The "cleanup" was done many weeks ago and didn't help much. So, no excitement here.

PR #348, opened 2023-08-23
Cleanup, mainly concerning sweepers

brownbaerchen:
I cleaned up a few things:
 - `generic_LU` is replaced by `generic_implicit` and is now removed from all files that are not deprecated
 - Said sweeper is deleted
 - Small typos
 - Moved `generic_implicit_MPI` to implementations folder
 - Removed unnecessary documentation as recommended by [PEP8](https://peps.python.org/pep-0008/#inline-comments) in a few cases
 - Minor refinements to the implementation extrapolation within Q error estimate that don't change the functionality


pySDC/playgrounds/EnergyGrids/playground.py
```diff
@@ -7,7 +7,6 @@
 from pySDC.implementations.problem_classes.Piline import piline
 from pySDC.implementations.sweeper_classes.imex_1st_order import imex_1st_order
 
-# from pySDC.implementations.sweeper_classes.generic_LU import generic_LU
```
pancetta, line 10: Isn't there a `sweeper_params['QI'] = 'LU'` missing?

pancetta (approved):

PR #352, opened 2023-09-11
Error estimate based on interpolation inside the collocation problem

brownbaerchen:
The solution to the collocation problem is an order $M$ polynomial that we can evaluate wherever we want by means of interpolation. This is specific to SDC and hence I have to find it cool. 
If we use $M-1$ collocation nodes of a converged collocation problem with $M$ nodes, we get an order $M-1$ polynomial. The result is not the same as the solution to a colocation problem with $M-1$ nodes, but both are valid polynomial approximations of the solution of order $M-1$, only using different quadrature points. Equidistant nodes work, Radau points work, whatever. They just give different properties regarding stability and and accuracy of the end point. If we choose any $M-1$ points other than the end point, if possible, we give up the super convergence. But we don't need that.
We can instead interpolate to the node we did not use and get a order $M-1$ approximation to that node, which we can compare to the order $M$ approximation given by the solution of the converged colocation problem to obtain an error estimate of the lower order polynomial, which is order $M$.
Sound familiar? We have been doing "embedded" SDC by doing $2M-1$ iterations with Gauß-Radau-Right quadrature and used the increment of the last iteration to obtain an error estimate of the $2M-2$ order second to last sweep. (Order 2M-1).
And now we can do adaptivity. Preliminary numerical results suggest that this works well. The major advantage is that we can do whatever we want to arrive at a converged collocation problem. Inexactness and what not. Diagonal preconditioners and the likes. And this could beat RK methods.
One disadvantage is that the lower order method has a lower order than with the increment based error estimate. If we consider 3 nodes for instance, we get an order 5(4) scheme with the increment, but a 5(2) scheme with this method. Now how bad is that? I thought because we are considering local errors only, this just means the same tolerance on the lower order method will give a tighter error of higher order method. But maybe I am wrong.
Crucially, we use solutions to converged collocation problems to advance to the next step. And if we choose a sensible criterion for what convergence means numerically, we make sure that the order of the solution to the step is what we think it is. Better even than with the increment method, I guess. As far as I understand, order reduction is not observed in converged collocation problems.

The included tests compute the order of accuracy of the error estimate with respect to the exact solution of the polynomial test problem, which is a polynomial of sufficiently high order.

Below is a quick and dirty performance plot for Allen-Cahn. The blue line is using adaptivity based on this error estimate using inexactness in both the Newton and linear iterations. The orange line is increment based adaptivity and the pink one is ESDIRK5(3). Don't worry about the yellow line. It is using a different estimate of the local error based on converged collocation problems based on a Taylor expansion of the solution at the end point of the interval, also of order $M-1$.
Figure_1

### DIFF
--- a/pySDC/implementations/convergence_controller_classes/estimate_interpolation_error.py
+++ b/pySDC/implementations/convergence_controller_classes/estimate_interpolation_error.py
@@ -1,0 +1,153 @@
+import numpy as np
+
+from pySDC.core.Lagrange import LagrangeApproximation
+from pySDC.core.ConvergenceController import ConvergenceController, Status
+from pySDC.core.Collocation import CollBase
+
+
+class EstimateInterpolationError(ConvergenceController):
+    """
+    Estimate the local error by using all but one collocation node in a polynomial interpolation to that node.
+    While the converged collocation problem with M nodes gives a order M approximation to this point, the interpolation
+    gives only an order M-1 approximation. Hence, we have two solutions with different order, and we know their order.
+    That is to say this gives an error estimate that is order M. Keep in mind that the collocation problem should be
+    converged for this and has order up to 2M. Still, the lower order method can be used for time step selection, for
+    instance.
+    By default, we interpolate to the second to last node.
+    """
+
+    def setup(self, controller, params, description, **kwargs):
+        """
+        Args:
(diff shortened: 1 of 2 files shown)


pancetta:
Man, I'm happy that we have these convergence controllers..

PR #359, opened 2023-09-22
Addition to #358

lisawim:
Forgot two classes .. now, everything should be fine with this PR. I'm sorry for the circumstances!

### DIFF
--- a/pySDC/implementations/problem_classes/AllenCahn_MPIFFT.py
+++ b/pySDC/implementations/problem_classes/AllenCahn_MPIFFT.py
@@ -30,7 +30,7 @@ class allencahn_imex(ptype):
     Parameters
     ----------
     nvars : List of int tuples, optional
-        Number of unknowns in the problem, e.g. ``nvars=[(128, 128), (128, 128)]``.
+        Number of unknowns in the problem, e.g. ``nvars=(128, 128)``.
     eps : float, optional
         Scaling parameter :math:`\varepsilon`.
     radius : float, optional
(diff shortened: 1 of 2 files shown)


pySDC/implementations/problem_classes/AllenCahn_MPIFFT.py
```diff
@@ -23,14 +23,14 @@ class allencahn_imex(ptype):
...
+    for :math:`i, j=0,..,N-1`, where :math:`N` is the number of spatial grid points. For time-stepping, the problem is treated
     *semi-implicitly*, i.e., the linear part is solved with Fast-Fourier Tranform (FFT) and the nonlinear part in the right-hand
-    side will be treated explicitly using mpi4py-fft [1]_ to solve them.
+    side will be treated explicitly using ``mpi4py-fft`` [1]_ to solve them.
 
     Parameters
     ----------
     nvars : List of int tuples, optional
-        Number of unknowns in the problem, e.g. [(128, 128), (128, 128)].
+        Number of unknowns in the problem, e.g. ``nvars=[(128, 128), (128, 128)]``.
```
brownbaerchen, line 33: Are you sure this works? In line 90, it seems that the length of `nvars` is used as the dimension. But I believe this indicates two grids of different resolution. I have never used this feature in pySDC, but I would expect the problem class to only be aware of a single grid..? Did you check if the default value works?

pancetta: Indeed, the problem class does not know about more than one grid. Using a list in the frontends indicates that some multi-level magic is about to happen. Only the transfer classes know about this (from a user's perspective), though.

lisawim:
I don't know what went wrong in the tests (in both PR's), because in both PR's Tutorial step 7 raises ``AssertionError``'s. Suddenly also something went wrong with the class(es) using PETSc.. Weird..

pancetta:
Please merge master, I fixed two dependency issues there.

PR #361, opened 2023-09-23
Small test for `nonlinear_ODE_1`

lisawim:
Starting with improving the test coverage of `pySDC` I added a small test for the `nonlinear_ODE_1` class. For this, I replaced the `ProblemError` in the `solve_system` method with a warning s.t. I can prove the number of Newton iterations.

### DIFF
--- a/pySDC/implementations/problem_classes/nonlinear_ODE_1.py
+++ b/pySDC/implementations/problem_classes/nonlinear_ODE_1.py
@@ -13,7 +13,7 @@ class nonlinear_ODE_1(ptype):
     given by
 
     .. math::
-        \frac{du(t)}{dt} = \sqrt(1 - u(t))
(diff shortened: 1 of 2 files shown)


pySDC/implementations/problem_classes/nonlinear_ODE_1.py
```diff
@@ -42,6 +49,9 @@ def __init__(self, u0=0.0, newton_maxiter=200, newton_tol=5e-11, stop_at_nan=Tru
             'u0', 'newton_maxiter', 'newton_tol', 'stop_at_nan', localVars=locals(), readOnly=True
         )
 
+        self.newton_itercount = 0
```
brownbaerchen, line 52: Please use the work counter interface in the future so we can have a common interface across different problems. You can add an arbitrary number of them and record of them in a single hook.

PR #370, opened 2023-10-27
Test for `AllenCahn_1D_FD.py`

lisawim:
I added a test for the one-dimensional AllenCahn equation. The test is more or less based on `test_NonlinearSchroedinger_MPIFFT.py`, testing things like:

- Do right-hand side evaluations match in semi-implicit and fully-implicit case?,
- check errors for each problem classes,
- do number of function calls match in semi-implicit and fully-implicit case?,
- is `ProblemError` raised correctly when `nan` arises?,
- and are `logger.warning` also called correctly in case of `nan`

Such basic things. In my opinion there could be also tests for some problem specific things. But at this moment I don't know yet which things this could be. For example, one of the AC-classes uses *Finel's trick*, which is only a rewriting of the right-hand side. Here could be also a specific test added, but I don't have any idea what test would make sense.

Let me know if anyone has some idea what I can add.

### DIFF
--- a/pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
+++ b/pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
@@ -183,7 +183,7 @@ class allencahn_front_fullyimplicit(ptype):
             raise ProblemError('Newton got nan after %i iterations, aborting...' % n)
         elif np.isnan(res):
             self.logger.warning('Newton got nan after %i iterations...' % n)
-            print(f'{type(self).__name__}')
+
         if n == self.newton_maxiter:
             self.logger.warning('Newton did not converge after %i iterations, error is %s' % (n, res))
 
--- a/pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
+++ b/pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
@@ -3,7 +3,7 @@ import scipy.sparse as sp
 from scipy.sparse.linalg import spsolve
 
 from pySDC.core.Errors import ProblemError
-from pySDC.core.Problem import ptype
+from pySDC.core.Problem import ptype, WorkCounter
 from pySDC.helpers import problem_helper
 from pySDC.implementations.datatype_classes.mesh import mesh, imex_mesh, comp2_mesh
 
@@ -53,14 +53,9 @@ class allencahn_front_fullyimplicit(ptype):
         Spatial grid values.
     uext : dtype_u
         Contains additionally the external values of the boundary.
-    newton_itercount : int
-        Counter for iterations in Newton solver.
-    lin_itercount : int
-        Counter for iterations in linear solver.
-    newton_ncalls : int
-        Number of calls of Newton solver.
-    lin_ncalls : int
-        Number of calls of linear solver.
+    work_counters : WorkCounter
+        Counter for statistics. Here, number of Newton calls and number of evaluations
+        of right-hand side are counted.
     """
 
     dtype_u = mesh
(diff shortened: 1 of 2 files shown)


pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
```diff
@@ -184,18 +177,16 @@ def solve_system(self, rhs, factor, u0, t):
...
             # increase iteration count
             n += 1
+            self.work_counters['newton']()
 
         if np.isnan(res) and self.stop_at_nan:
             raise ProblemError('Newton got nan after %i iterations, aborting...' % n)
         elif np.isnan(res):
             self.logger.warning('Newton got nan after %i iterations...' % n)
-
+            print(f'{type(self).__name__}')
```
brownbaerchen, line 186: Did you forget to remove this?

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
```diff
@@ -735,7 +717,7 @@ def __init__(
         stop_at_nan=True,
     ):
         super().__init__(nvars, dw, eps, newton_maxiter, newton_tol, interval, radius, stop_at_nan)
-        self.A -= sp.eye(self.init) * 0.0 / self.eps**2
+        self.A -= sp.eye(self.nvars) * 0.0 / self.eps**2
```
brownbaerchen, line 720: Does this line do anything? It just subtracts zero. At least with your fix this does not raise an error, but still. Is there a reason for this line @pancetta?

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
```diff
@@ -822,7 +805,7 @@ def __init__(
         stop_at_nan=True,
     ):
         super().__init__(nvars, dw, eps, newton_maxiter, newton_tol, interval, radius, stop_at_nan)
-        self.A -= sp.eye(self.init) * 0.0 / self.eps**2
+        self.A -= sp.eye(nvars) * 0.0 / self.eps**2
```
brownbaerchen, line 808: Again, what does this do?

brownbaerchen (commented):
Thanks a lot! Looks like you fixed quite a lot of things!

Generally, a test that I would find useful for problems with a Newton solver is to compare the solver to [Newton Krylov from scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.newton_krylov.html) like implemented [here](https://github.com/Parallel-in-Time/pySDC/blob/714271351d7cdd6132d7a78094faaa8c455cfb54/pySDC/implementations/problem_classes/NonlinearSchroedinger_MPIFFT.py#L275).

It should be possible to write a generic function that takes as input a problem class and some parameters and then solves using the solver in the problem class and scipy and compares. It is some effort, though. So don't feel like you need to do this. But if you are eager to write some more tests, this could be useful.

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
lisawim: Yes, did forget that.

lisawim:
Thanks for your comment! That sounds meaningful!

I also noticed that I forgot several lines to test. I'll add some more testing for `allencahn_periodic_multiimplicit`.

brownbaerchen:
I advocate the use of git hooks to automate black as described [here](https://github.com/Parallel-in-Time/pySDC/blob/master/docs/contrib/02_continuous_integration.md).

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
pancetta: There was, but not anymore. I don't know why this is still in there. Can be removed!

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
pancetta: see above

PR #371, opened 2023-10-31
Added `solve_system` to all DAE problem classes

lisawim:
Currently, the `fully_implicit_DAE` sweeper solves all DAE problems in the same way, i.e., using `scipy.optimize.root(f, x0, method='hybr')`. Since it could be useful to choose the solver problem dependent, a `solve_system` method is added to all DAE classes as this is already the case for all implemented ODE/PDE classes in `pySDC`. The implicit system is defined in the sweeper anyway, because this nonlinear system does not change (since it is the collocation problem to be solved).

So no magic here in this PR, I only moved these two line
```
opt = root(impl_sys, u0, method='hybr', tol=self.newton_tol)
me[:] = opt.x
```
Now, the problem classes still uses the same solver, but for the future, also different solver could be implemented and no one has to take care to change the sweeper itself every time.
to the problem classes.

Now, also `WorkCounters` are used to count the work which is also useful to get some statistics about the calls of the right-hand side and the calls of the right-hand side during the solve (because for `hybr` no callback is available and also no iterations needed for the solve can be found..).

### DIFF
--- a/pySDC/projects/DAE/problems/simple_DAE.py
+++ b/pySDC/projects/DAE/problems/simple_DAE.py
@@ -1,7 +1,6 @@
 import warnings
 import numpy as np
 from scipy.interpolate import interp1d
-from scipy.optimize import root
 
 from pySDC.projects.DAE.misc.ProblemDAE import ptype_dae
 from pySDC.core.Problem import WorkCounter
@@ -38,8 +37,7 @@ class pendulum_2d(ptype_dae):
     ----------
     work_counters : WorkCounter
         Counts the work, i.e., number of function calls of right-hand side is called and stored in
-        ``work_counters['rhs']``, the number of function calls of the Newton-like solver is stored in
-        ``work_counters['newton']``.
+        ``work_counters['rhs']``.
 
     References
     ----------
@@ -57,7 +55,6 @@ class pendulum_2d(ptype_dae):
         # solution = data[:, 1:]
         # self.u_ref = interp1d(t, solution, kind='cubic', axis=0, fill_value='extrapolate')
         self.t_end = 0.0
-        self.work_counters['newton'] = WorkCounter()
         self.work_counters['rhs'] = WorkCounter()
 
     def eval_f(self, u, du, t):
@@ -85,36 +82,6 @@ class pendulum_2d(ptype_dae):
         f[:] = (du[0] - u[2], du[1] - u[3], du[2] + u[4] * u[0], du[3] + u[4] * u[1] + g, u[0] ** 2 + u[1] ** 2 - 1)
         self.work_counters['rhs']()
         return f
-
-    def solve_system(self, impl_sys, u0, t):
-        r"""
-        Solver for nonlinear implicit system (defined in sweeper).
-
-        Parameters
-        ----------
-        impl_sys : callable
-            Implicit system to be solved.
-        u0 : dtype_u
-            Initial guess for solver.
-        t : float
-            Current time :math:`t`.
-
-        Returns
-        -------
-        me : dtype_u
-            Numerical solution.
-        """
-
-        me = self.dtype_u(self.init)
-        opt = root(
-            impl_sys,
-            u0,
-            method='hybr',
-            tol=self.newton_tol,
-        )
-        me[:] = opt.x
-        self.work_counters['newton'].niter += opt.nfev
-        return me
 
     def u_exact(self, t):
         """
@@ -177,8 +144,7 @@ class simple_dae_1(ptype_dae):
     ----------
     work_counters : WorkCounter
         Counts the work, i.e., number of function calls of right-hand side is called and stored in
-        ``work_counters['rhs']``, the number of function calls of the Newton-like solver is stored in
-        ``work_counters['newton']``.
+        ``work_counters['rhs']``.
 
     References
     ----------
@@ -190,7 +156,6 @@ class simple_dae_1(ptype_dae):
         """Initialization routine"""
         super().__init__(nvars, newton_tol)
 
-        self.work_counters['newton'] = WorkCounter()
         self.work_counters['rhs'] = WorkCounter()
 
     def eval_f(self, u, du, t):
@@ -221,36 +186,6 @@ class simple_dae_1(ptype_dae):
         )
         self.work_counters['rhs']()
         return f
-
-    def solve_system(self, impl_sys, u0, t):
-        r"""
-        Solver for nonlinear implicit system (defined in sweeper).
-
-        Parameters
-        ----------
-        impl_sys : callable
-            Implicit system to be solved.
-        u0 : dtype_u
-            Initial guess for solver.
-        t : float
-            Current time :math:`t`.
-
-        Returns
-        -------
-        me : dtype_u
-            Numerical solution.
-        """
-
-        me = self.dtype_u(self.init)
-        opt = root(
-            impl_sys,
-            u0,
-            method='hybr',
-            tol=self.newton_tol,
-        )
-        me[:] = opt.x
-        self.work_counters['newton'].niter += opt.nfev
-        return me
 
     def u_exact(self, t):
         """
@@ -297,8 +232,7 @@ class problematic_f(ptype_dae):
         Specific parameter of the problem.
     work_counters : WorkCounter
         Counts the work, i.e., number of function calls of right-hand side is called and stored in
-        ``work_counters['rhs']``, the number of function calls of the Newton-like solver is stored in
-        ``work_counters['newton']``.
+        ``work_counters['rhs']``
 
     References
     ----------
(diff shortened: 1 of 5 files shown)


pancetta:
Very nice! To me there seems to be quite a bit of boilerplate code (since the `solve_system` routine is there for many different problem classes. Would it make sense to have this as a generic/default solver routine somewhere? If there is even more boilerplate code like this (didn't check), you could inherit from one of the problems classes when designing the other ones?

lisawim:
Of course it does make sense to have a class with the `solve_system` method and all problem classes could inherit from! I was thinking of the `pySDC.projects.DAE.misc.ProblemDAE` class where the method is added, because all DAE classes inherits from it anyway. Would this be okay? Since another solver should be implemented, `solve_system` can then be overwritten in the corresponding class.

pancetta:
Sounds good! You can add more generic functions there, I assume?

lisawim:
Yes, this should be possible.

pySDC/projects/DAE/problems/simple_DAE.py
```diff
@@ -49,6 +55,7 @@ def __init__(self, nvars, newton_tol):
         # solution = data[:, 1:]
         # self.u_ref = interp1d(t, solution, kind='cubic', axis=0, fill_value='extrapolate')
         self.t_end = 0.0
+        self.work_counters['rhs'] = WorkCounter()
```
pancetta, line 58: Looks like at least this one could go to the generic parent class, too?

lisawim: True! I added this for all problem classes. I'll do that.

lisawim: Looks better and more convenient (like before), thanks a lot!

pySDC/projects/DAE/sweepers/fully_implicit_DAE.py
```diff
@@ -141,12 +183,16 @@ def impl_fn(params):
         return None
 
     def predict(self):
-        """
-        Predictor to fill values at nodes before first sweep
+        r"""
+        Predictor to fill values at nodes before first sweep. It can decides whether the
```
brownbaerchen, line 187: `decides` -> `decide`

brownbaerchen (commented):
To me it seems like you could indeed save a lot of lines by inheriting from `generic_implicit` in the `fully_implicit_DAE` sweeper. You can also consider doing multiple inheritance if you plan to do IMEX at some point.
I have not checked very thoroughly, but I believe the `integrate`, `compute_residual`,  `predict` and `compute_end_point` functions are all equivalent to the ones in `generic_implicit`.
If you move the `implSystem` function to the problem class, would it be possible to just use the `generic_implicit` sweeper? I think there was some issue with this. Also, `u_approx` is not usually passed to `solve_system`, I think..? It probably doesn't work, but if it did, that would be nice.

lisawim:
> To me it seems like you could indeed save a lot of lines by inheriting from `generic_implicit` in the `fully_implicit_DAE` sweeper. You can also consider doing multiple inheritance if you plan to do IMEX at some point. I have not checked very thoroughly, but I believe the `integrate`, `compute_residual`, `predict` and `compute_end_point` functions are all equivalent to the ones in `generic_implicit`. If you move the `implSystem` function to the problem class, would it be possible to just use the `generic_implicit` sweeper? I think there was some issue with this. Also, `u_approx` is not usually passed to `solve_system`, I think..? It probably doesn't work, but if it did, that would be nice.

Yes, IMEX is indeed planned (and already done but not pushed yet), and I agree that multiple inheritance can be done here. At least `integrate`, `compute_end_point` and `predict` could be used from `generic_implicit`. For `compute_residual` I have to think about that. `eval_f` is used here anyway, but in case of DAEs we also have the `du` argument here. This issue could probably be solved by setting `du` as a arbitrary argument (i.e., by setting `du=None` somewhere)..?
Moving the implSystem function to the problem class - I'll check if this would work.

lisawim:
Hm, I also got a message that you @brownbaerchen wrote somewhere (I cannot find it) that the `scipy.optimize.root` routine has a callback where I can call `work_counters['newton']`. This is true, but for `method='hybr'` this callback function is not available.

brownbaerchen:
> Hm, I also got a message that you @brownbaerchen wrote somewhere (I cannot find it) that the `scipy.optimize.root` routine has a callback where I can call `work_counters['newton']`. This is true, but for `method='hybr'` this callback function is not available.

Yes, I saw what you wrote after I made the comment, so I marked it as resolved.

brownbaerchen:
> > To me it seems like you could indeed save a lot of lines by inheriting from `generic_implicit` in the `fully_implicit_DAE` sweeper. You can also consider doing multiple inheritance if you plan to do IMEX at some point. I have not checked very thoroughly, but I believe the `integrate`, `compute_residual`, `predict` and `compute_end_point` functions are all equivalent to the ones in `generic_implicit`. If you move the `implSystem` function to the problem class, would it be possible to just use the `generic_implicit` sweeper? I think there was some issue with this. Also, `u_approx` is not usually passed to `solve_system`, I think..? It probably doesn't work, but if it did, that would be nice.
> 
> Yes, IMEX is indeed planned (and already done but not pushed yet), and I agree that multiple inheritance can be done here. At least `integrate`, `compute_end_point` and `predict` could be used from `generic_implicit`. For `compute_residual` I have to think about that. `eval_f` is used here anyway, but in case of DAEs we also have the `du` argument here. This issue could probably be solved by setting `du` as a arbitrary argument (i.e., by setting `du=None` somewhere)..? Moving the `implSystem` function to the problem class - I'll check if this would work.

In this case, I vote to just overload the function and copy paste a few lines rather than adding default arguments to the other stuff. Multiple inheritance is tricky business in my opinion. For me, it requires a bit of trial and error as well as some praying. But it is kind of working [here](https://github.com/Parallel-in-Time/pySDC/blob/73cb8e30428b01f82728136e6cbc095124cb586b/pySDC/implementations/sweeper_classes/generic_implicit_MPI.py#L133) and I think it's quite elegant once it works.

lisawim:
Indeed, I could removed lines by removing `integrate` and `compute_end_point` in the sweeper. For `predict`, `compute_residual` I didn't see yet how I can use the one from the parent class. For `predict` using `spread` `eval_f` is used here to spread the values. For `compute_residual` also `eval_f` is used to define the residual.

brownbaerchen:
I see. In `predict`, I think the only thing that's different is `spread`. So you could call the parent's spread method if the initial guess is not `spread` and otherwise do what you have here. But that's optional :D

As for the residual, I am a bit confused that the residual is based on the magnitude of the right hand side evaluations rather than the integral. But I have no idea about the DAE implementation. Feel free to ignore this. You don't have to abstract as much as you can. Using code that is already there is better than repeating code, but bending over backwards to somehow make the code that is already there do what you want in a different place is also no good.

lisawim:
I tried that but in the `predict` method from `sweeper` it is `L.f[0] = P.eval_f(L.u[0], L.time)` used to compute the function value at initial time, and here I also have the struggle by using `eval_f` which needs `du` to be computed.. 

Regarding the residual: At first I was also a bit confused because using the right-hand side for computing the residual was a bit cumbersome. Let me explain you shortly (as this makes sense for me at least): How do you compute the residual for a system of equations? In case of having $Au=b$, the residual can be computed by $r=Au-b$. In case of nonlinearity, you can define a system like $F(u)=0$, then $r=F(u)$ is the residual. When you think of that a DAE is given in the general nonlinear formulation $$F\left(\frac{d u}{dt}, u, t\right) = 0,$$ then you can also compute the residual by evaluating the right-hand side using your current numerical solutions `L.u` and `L.f` here. So, for the current numerical solution it is checked how well it does satisfy the DAE formulation or not. Does that answer your question?

But now I'm thinking whether we also can compute the residual by using the integral formulation inside the DAE formulation.. Hm..

lisawim:
Ah, I see. For $F\left(u, \frac{du}{dt}, t\right)=0$, you have the (node-wise) collocation formulation $$F\left(u_0 + \sum_{j=1}^M q_{m, j} U_j, U_m, \tau_m\right)=0, \qquad m=1,..,M$$ for $u_m\approx u(\tau_m)$, $U_m \approx u'(\tau_m)$ and collocation nodes $t_0 \leq \tau_1 <..<\tau_M \leq t_1$.
This would also be the formula to compute the (node-wise) residual in case of DAEs. The integral is already computed in line `148-149` in `fully_implicit_DAE`, so you can use `L.u[m + 1]` here, insert into `eval_f` and this is the residual at $\tau_m$. Does that make sense?

brownbaerchen:
Honestly, don't worry about abstraction. I was missing some small details that do make a difference and I think there is little to gain from further abstraction. Sorry about that :D

pancetta:
Can we merge this? Issue https://github.com/lisawim/pySDC/issues/8 seems to be a topic for another PR anyway?

lisawim:
Yes, can be merged.

PR #374, opened 2023-11-07
MPI version of IMEX sweeper

brownbaerchen:
Added MPI version of `imex_1st_order`. Accompanied by a simple test that runs a quick problem and checks if the result is the same, both in terms of `level.uend` as well as the residual. It's not the most elegant test, but the best compromise between kinda meaningful and quick to implement that I could come up with this morning.

### DIFF
--- a/pySDC/tests/test_sweepers/test_MPI_sweeper.py
+++ b/pySDC/tests/test_sweepers/test_MPI_sweeper.py
@@ -97,8 +97,8 @@ def test_sweeper(num_nodes, quad_type, residual_type, imex, launch=True):
             use_MPI=False, num_nodes=int(num_nodes), quad_type=quad_type, residual_type=residual_type, imex=imex
         )
 
-        assert np.allclose(MPI.uend, nonMPI.uend), 'Got different solutions at end point!'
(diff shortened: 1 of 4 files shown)


pySDC/tests/test_sweepers/test_MPI_sweeper.py
```diff
@@ -0,0 +1,108 @@
...
+        import numpy as np
+
+        imex = False if imex == 'False' else True
+        MPI = run(use_MPI=True, num_nodes=int(num_nodes), quad_type=quad_type, residual_type=residual_type, imex=imex)
+        nonMPI = run(
+            use_MPI=False, num_nodes=int(num_nodes), quad_type=quad_type, residual_type=residual_type, imex=imex
+        )
+
+        assert np.allclose(MPI.uend, nonMPI.uend), 'Got different solutions at end point!'
+        assert np.allclose(MPI.status.residual, nonMPI.status.residual), 'Got different residuals!'
```
pancetta, line 101: Could you specify tight tolerances here, please? The default parameters are rather high, it seems.

PR #376, opened 2023-11-08
Inexactness and some tests

brownbaerchen:
This PR includes slight changes to the Newton inexactness convergence controller to make it a bit more general as well as the option for linear inexactness in Allen-Cahn and Quench.
Also, the test for the extrapolation error estimate within the collocation method is refined to be closer to a unit test than an integration test.
Efficient sweepers allow to save a few operations when computing the residual only on the last rank.


pancetta:
Would it make sense to test this with IMEX, too?

brownbaerchen:
What exactly are you referring to? The interpolation between restarts actually is not implemented with IMEX, if I remember correctly. The extrapolation inside the collocation problem is something we can test in IMEX. We would have to add an IMEX version of the polynomial test problem. But then again, this feature is not too important.

pySDC/implementations/convergence_controller_classes/estimate_extrapolation_error.py
```diff
@@ -485,18 +487,21 @@ def post_iteration_processing(self, controller, S, **kwargs):
...
+        if None in self.coeff.u or self.params.recompute_coefficients:
+            self.get_extrapolation_coefficients(nodes, dts, t_eval)
 
         # compute the extrapolated solution
+        if lvl.f[0] is None:
+            lvl.f[0] = lvl.prob.eval_f(lvl.u[0], lvl.time)
+
         if type(lvl.f[0]) == imex_mesh:
-            f = [me.impl + me.expl for me in lvl.f]
+            f = [lvl.f[i].impl + lvl.f[i].expl if self.coeff.f[i] and lvl.f[i] else 0.0 for i in range(len(lvl.f) - 1)]
```
pancetta, line 502: I guess this is what I'm referring to

PR #380, opened 2023-11-16
XPU implementation of Dahlquist problem

brownbaerchen:
Being an abstraction junky, I yearn for GPU implementations that are not copy-paste versions of the CPU version. Starting out with GPUs, I was immediately annoyed that the GPU version of Allen-Cahn has not been updated together with the CPU version. So I made an XPU version of that analogous to the Dahlquist thing that is in this PR and it seems to work reasonably well.
I did the XPU version of Dahlquist because I needed it to fix a bug in the RK sweeper. (The bug was of the shallow vs. deep copy kind in line 274 which I am angry at myself for still making...)

Now, I doubt this is the best XPU implementation possible. I specifically did this, sort of, minimal example so you can complain and we can find a good solution together.

Are there any major objections to this abstraction? Since the classes are distinct, we maintain the option to use completely different methods on GPU by simple overloading, which may become important when doing multi-GPU stuff or sth. like that later on. Of course, `self.xp` instead of `np` is not great. I recon most people are so used to seeing `np` that this can cause some confusion. I personally prefer this over excessive copy-pasting, though.


pySDC/implementations/problem_classes/TestEquation_0D_GPU.py
pancetta, line 1: I think you should put that into the `TestEquation_0D.py` file, right? We don't need backward compatibility here.

pySDC/implementations/problem_classes/TestEquation_0D.py
```diff
@@ -30,15 +27,38 @@ class testequation0d(ptype):
...
+        pySDC.Problem
+            A problem class implementing the desired implementation
+        """
+        if version == 'CPU':
+            return testequation0d
+        elif version == 'GPU':
+            from pySDC.implementations.problem_classes.TestEquation_0D_GPU import testequation0dGPU
+
+            return testequation0dGPU
+        else:
```
pancetta, line 51: Can we make the CPU version the default? What do you think?

pySDC/implementations/problem_classes/TestEquation_0D.py
```diff
@@ -30,15 +27,38 @@ class testequation0d(ptype):
...
+
+            raise ParameterError(f'Don\'t know version {version}! Please choose \'CPU\' or \'GPU\'!')
 
     def __init__(self, lambdas=None, u0=0.0):
         """Initialization routine"""
         if lambdas is None:
-            re = np.linspace(-30, 19, 50)
-            im = np.linspace(-50, 49, 50)
-            lambdas = np.array([[complex(re[i], im[j]) for i in range(len(re))] for j in range(len(im))]).reshape(
+            re = self.xp.linspace(-30, 19, 50)
```
pancetta, line 59: fine by me

pySDC/implementations/sweeper_classes/Runge_Kutta.py
```diff
@@ -272,7 +271,7 @@ def update_nodes(self):
 
         for m in range(0, M):
             # build rhs, consisting of the known values from above and new values from previous nodes (at k+1)
-            rhs = lvl.u[0]
+            rhs = prob.dtype_u(lvl.u[0])
```
pancetta, line 274: Been there, done that..

pySDC/implementations/problem_classes/TestEquation_0D_GPU.py
brownbaerchen: I started out doing this, but apparently it tries importing the modules when reading this file, regardless of whether this class is imported. So we have to put it in a different file. We could consider having a single file with GPU implementations of all problems if we apply this to multiple problems.

pancetta: Sounds like a similar problem as with the tests and MPI. What if you do the import within the function, not on top of the file?

brownbaerchen: The import happens immediately before we set the class attribute `dtype_u`. Actually, I started out with having this inside `__init__`, but then I was afraid of unexpected behaviour when you access the class attribute without instantiation.

brownbaerchen:
Do we want to merge this if there are no further complaints?

pancetta:
I guess if we merge this we'd need to port the other GPU-ready problems classes to this new way of abstraction, too?

pancetta:
I wonder if we should keep the GPU stuff in a separate branch, also because there are no automated tests etc. to make sure everything works.

brownbaerchen:
I can port the other problems, that's no problem. Although these changes don't break the API, so we can do that in a separate PR. Whatever you want.

I personally suggest merging in the master branch so as not to cause divergence. In particular, the abstraction means that if the CPU version works, the GPU version will work as well, unless the cupy version is showing unexpected behaviour. So there is not 100% confidence that tests passing on CPU will pass on GPU as well, but I guess it is likely. Since the cupy functions of the same name are meant to replicate the numpy behaviour, it would be a bug on their part.
Of course there are some missing functions. For instance, the `scipy.integrate` package is missing in cupy. So adding a new function to the CPU implementation is not guaranteed to work anymore.
Neither is particularly great. But compared to the current implementation of existing problems, I think we increase both confidence and coverage in GPU implementations.

PR #384, opened 2023-11-30
Datatype `DAEMesh` for DAEs

lisawim:
Together with @brownbaerchen I implemented a datatype for DAEs that allows to distinguish between the differential part and the algebraic part of the problem. `fully_implicit_DAE` does not change, but `solve_system` needs to be adapted, since it also has to be distinguish between these different parts. Since the scipy root solver don't know such a mesh type, a separate function `implSysAsNumpy` is defined to convert the current mesh into a numpy array of appropriate shape. The DAE hook class also have a new hook logging the error in the algebraic variable, which is more clean with the new datatype now.

I guess a pretty nice advantage is that the IMEX-SDC-DAE sweeper (that will come in another PR sometime) can also be used for the same problems as `fully_implicit_DAE` can be used. Thus, no IMEX DAE versions of the problems are necessary here.

For this, all files and all tests of the DAE projects needs to be adapted. Yesterday, we noticed that the tests for the problems and also for the sweeper could be way more strict, however.. The PinTSimE project needs to be adapted as well since it contains this folder for the paper including more DAE stuff. But this is moved to future, where more time is available to do that.

If you have some further comments, improvements, etc. I'm happy when you let me know that.


pySDC/projects/DAE/problems/transistor_amplifier.py
```diff
@@ -61,8 +61,8 @@ class one_transistor_amplifier(ptype_dae):
         Springer (2009).
     """
 
-    def __init__(self, nvars, newton_tol):
-        super().__init__(nvars, newton_tol)
+    def __init__(self, newton_tol):
+        super().__init__(nvars=(5, 0), newton_tol=newton_tol)
```
brownbaerchen, line 65: Why is this class a DAE if there is no algebraic part? Does it make sense to switch to `mesh` for this class?

pySDC/projects/DAE/problems/transistor_amplifier.py
```diff
@@ -195,8 +198,8 @@ class two_transistor_amplifier(ptype_dae):
         Lect. Notes Math. (1989).
     """
 
-    def __init__(self, nvars, newton_tol):
-        super().__init__(nvars, newton_tol)
+    def __init__(self, newton_tol):
+        super().__init__(nvars=(8, 0), newton_tol=newton_tol)
```
brownbaerchen, line 202: This class also doesn't have an algebraic part.

pySDC/tests/test_projects/test_DAE/test_misc.py
```diff
@@ -8,7 +8,7 @@ def test_problem_class_main():
     from pySDC.projects.DAE.problems.simple_DAE import simple_dae_1
 
     # initialize problem parameters
-    problem_params = dict()
+    problem_params = {'nvars': 3}
```
brownbaerchen, line 11: Why do you need to pass `nvars` here? Didn't you remove it everywhere? Does this class support different number of variables?

lisawim:
Corrections to be continued.

pySDC/projects/DAE/problems/transistor_amplifier.py
lisawim: There are also fully-implicit DAEs, i.e., you don't have any algebraic equations here, also no algebraic variables. The thing what it makes the equation to a DAE is its singular Jacobian matrix. So, a DAE mesh does not make any sense here.

pySDC/tests/test_projects/test_DAE/test_misc.py
lisawim: Maybe this whole test does not make any sense, so I'll remove that. Complaints?

brownbaerchen: Fine by me. I think we don't need to overdo the tests. Testing math is more important than testing API.

lisawim:
Corrections are finished now. Feel free to merge (or to comment).

brownbaerchen:
We certainly added some mess while reducing the mess. Since it's all contained within the DAE project, I am fine with merging this and cleaning it up later. Whichever you prefer @lisawim, @pancetta.

pySDC/projects/DAE/misc/HookClass_DAE.py
```diff
@@ -41,24 +44,27 @@ def post_step(self, step, level_number):
...
+    each time step.
     """
 
     def __init__(self):
-        """
-        Initialization routine for the custom hook
-        """
-        super(error_hook, self).__init__()
+        """Initialization routine"""
+        super().__init__()
```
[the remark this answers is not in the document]
pancetta: Remove this, please.

pySDC/projects/DAE/problems/simple_DAE.py
```diff
@@ -139,6 +154,11 @@ class simple_dae_1(ptype_dae):
         equations. Society for Industrial and Applied Mathematics (1998).
     """
 
+    def __init__(self, newton_tol=1e-10):
+        """Initialization routine"""
+        super().__init__(nvars=(2, 1), newton_tol=newton_tol)
+        self._makeAttributeAndRegister('newton_tol', localVars=locals())
```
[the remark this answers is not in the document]
pancetta: @lisawim ?

pySDC/projects/DAE/misc/HookClass_DAE.py
```diff
@@ -41,24 +44,27 @@ def post_step(self, step, level_number):
 
 class error_hook(hooks):
     """
-    Hook class to add the approximate solution to the output generated by the sweeper after each time step
+    Hook class to log the error to the output generated by the sweeper after
```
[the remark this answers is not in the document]
pancetta: @lisawim ?

lisawim:
Using fully-implicit DAEs with `mesh` type and semi-explicit DAEs with `DAEMesh` type with FI-SDC still causes some inconvenient handlings that I have fixed but its a kind of some mess as @brownbaerchen mentioned. At this time, I had no reasonable idea to make it as clean as possible. But I've arranged to meet @brownbaerchen in Bruges to clean it up. I'm also fine with both. But the SISDC sweeper will follow anyway. :smiling_imp:

pySDC/projects/DAE/misc/HookClass_DAE.py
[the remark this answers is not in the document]
lisawim: Ah, forgot that.. I'll change that.

pySDC/projects/DAE/problems/simple_DAE.py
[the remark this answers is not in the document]
lisawim: I'll also fix that!

pySDC/projects/DAE/misc/HookClass_DAE.py
```diff
@@ -1,84 +1,86 @@
...
         L = step.levels[level_number]
+        P = L.prob
 
         # TODO: is it really necessary to recompute the end point? Hasn't this been done already?
         L.sweep.compute_end_point()
 
+        # compute and save errors
+        # Note that the component from which the error is measured is specified here
+        upde = P.u_exact(step.time + step.dt)
+        err = abs(upde.diff - L.uend.diff)
```
lisawim, line 38: But here @pancetta and @brownbaerchen the error is computed over the differential variables. When I would use the generic error hook classes I got the maximum error over both the differential and the algebraic variables. So, I still would like to keep this class at least for the semi-explicit DAEs.  For the fully-implicit DAEs the generic error hook classes does make sense to use!

pancetta:
So, do you want to leave it open until Bruges or should I merge?

lisawim:
I would like to leave it open until Bruges. When the new data type will be merged into `pySDC`, then it should be implemented in a clean way.

lisawim:
Using the suggested ``MultiComponentMesh`` datatype for the DAE mesh I updated all the stuff in this PR. It is probably a better solution than before! Before pushing I'd like to write test(s) for the new mesh. Comments, improvements, anything else is welcome!

pancetta:
Looks like the test is too slow, please correct. Otherwise, I really like this solution!

brownbaerchen:
Can you extend the test of DAE mesh different shapes, i.e. PDAEs? Also, you only add a number to the entire array, but we should check what happens when you do ufuncs with the components of the mesh as well.

lisawim:
Ah, cool! I did not see your commit.. Indeed, it can deal with that. I adapted the first test for the mesh for initialization. Adaption of the other test will follow!

pancetta:
Are we good to go now?

pySDC/projects/DAE/misc/ProblemDAE.py
```diff
@@ -54,14 +54,30 @@ def solve_system(self, impl_sys, u0, t):
...
-        )
-        me[:] = opt.x
+
+        def implSysAsNumpy(unknowns, **kwargs):
+            me.diff[:] = unknowns[: np.size(me.diff)].reshape(me.diff.shape)
+            me.alg[:] = unknowns[np.size(me.diff) :].reshape(me.alg.shape)
+            sys = impl_sys(me, **kwargs)
+            return np.append(sys.diff.flatten(), sys.alg.flatten())  # TODO: more efficient way?
+
+        if type(me) == DAEMesh:
```
brownbaerchen, line 65: We don't need this anymore, right? We should be able to plug this into scipy solvers directly.

lisawim:
Yes, Mr. @pancetta!

brownbaerchen (changes requested):

pySDC/projects/DAE/sweepers/fully_implicit_DAE.py
```diff
@@ -118,11 +107,14 @@ def implSystem(params):
                     System to be solved as implicit function.
                 """
 
-                params_mesh = P.dtype_f(P.init)
-                params_mesh[:] = params
+                if type(params) == DAEMesh:
```
brownbaerchen, line 110: I think we also don't need this anymore...? Maybe we need to do sth like `params_mesh[...] = params`?

brownbaerchen (approved):

PR #386, opened 2023-12-15
WIP: Use HPC for CI

jakob-fritz:


.gitlab-ci.yml
```diff
@@ -1,8 +1,76 @@
...
+    - module load Stages/2024
+    - module load GCC
+    - module load OpenMPI
+    - module load FFTW
+    - module load mpi4py
+    - module load SciPy-Stack
+    - module load CuPy
+    - pwd
+    - ls -lah
+    - pip install -e .
```
brownbaerchen, line 60: Do you need to `pip install` in the CI? User installation with pip is persistent, afaik. Is the gitlab not associated with any user?
I would prefer to put the `module load` commands in the job scripts. But if you need to repeat the `pip install`s every time, it's better here.

.gitlab-ci.yml
```diff
@@ -1,8 +1,76 @@
...
+    - jacamar
+    - juwels
+    - login
+    - shell
+  artifacts:
+    when: always
+    paths:
+      - benchmarks
+      - sbatch.err
+      - sbatch.out
```
brownbaerchen, line 46: Could we also have a directory where all job scripts can post their output? It would be neat to allow multiple job scripts. Not really needed, though.

jakob-fritz: Yes sure we can have a directory for the outputs. Each job in Gitlab spawns in a separate directory. Therefore, multiple jobs (from Gitlab) won't put files in the same directory. If multiple tasks from slurm are spawned from a single Job in Gitlab, having this directory might make sense.

.gitlab-ci.yml
jakob-fritz: I'm not sure, if we need `pip install` each time. The jobs are run as a predefined user (the one who triggers the CI in Gitlab). As mirroring is done with a personal access token, that user is impersonated and the CI will always run as this specific user.

Regarding `module load`: We can put them in the script. I find it easier/nicer if the scripts are as short as possible. Therefore, I moved the `module load` into the YAML-file. Furthermore, these steps (as module load or pip install) are executed on a login-node. The content of the sh-file is executed on a compute-node. In term of quota, it is "cheaper" if we don't spend compute-time for module load (although it does not take too long).

 If you want to, feel free to move the pip install and module load into the script. If I shall do that, feel free to ping me!

jakob-fritz:
For the job `mirror to gitlab` to work, the information needs to be provided in the Github-Repo.
For the required variables (and where to get them) see [the readme of the used action](https://github.com/jakob-fritz/github2lab_action?tab=readme-ov-file#how-to-set-up-a-gitlab-ci-with-a-github-repo-as-a-maintainer)

brownbaerchen:
How do we continue? If I understand correctly, we need to add a personal access token from GitLab as a secret to this repository. Does it make sense that @pancetta sets up the GitLab repo now? He has to add the secret anyhow.

jakob-fritz:
Yes, probably it is best, if @pancetta does this. I can also create the repo (or use an existing one from me), but pancetta needs to add the secret in the end anyway.

As said: The steps are listed above. If questions occur, I am willing to help. In person only in the next year, as this year, I'm out of office

pancetta:
OK, the repository is `https://gitlab.jsc.fz-juelich.de/atml-ati/pysdc` and the secret is under `GITLAB_SECRET`.

pancetta:
Note that the gitlab repository is empty, so I could not allow force-push. I tried with wildcard `*`, but I don't know if that works.

jakob-fritz:
@brownbaerchen , I assume to be done with the changes. If you're fine with them, feel free to merge. Once this is done, we can double-check and merge that branch into the main one at PiT/pySDC.

If you find issues or want things to be changed, feel free to contact me.

brownbaerchen:
Thanks a lot for your work! Can you please explain what your code is actually doing? I can see there is a job that mirrors to Gitlab, but what it does is not clear to me. I assume it runs the GPU environment tests and returns the result to GitHub somehow? But as far as I can tell, the coverage report of the GPU tests is not used in the post processing step in the original pipeline. This is a crucial feature that we still need.

PR #391, opened 2024-01-10
Removed seperate file for GPU Dahlquist implementation

### LINKED ISSUE #390: Parallel modifications of the `testequation0d` class due to GPU implementations
Just realized it when merging the current [master branch](https://github.com/Parallel-in-Time/pySDC/tree/master) into the [development branch for diagonal SDC](https://github.com/caklovicka/pySDC/tree/diagonal) ...

Concerning the `TestQuation_0D.py` problem class module : 

- modifications by @brownbaerchen for CPU/GPU capabilities (https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/implementations/problem_classes/TestEquation_0D.py)
- my own modifications to avoid some issues when generating stability graph, and removal of any scipy.sparse dependencies (because why the hell do we need it :raised_eyebrow:) (https://github.com/caklovicka/pySDC/blob/diagonal/pySDC/implementations/problem_classes/TestEquation_0D.py)

Currently this generates a conflict, and there is two options :

1. do @brownbaerchen need additional modifications later on this file ? In that case I'll wait for this last modifications to merge master into my diagonal development branch, and then solve the conflicts on my side by taking into account the simplified implementation for the `testequation0d`.
2. negative answer to the previous question, then I'll solve the conflict now while ensuring that it does not brakes the changes induced for the GPU implementation

brownbaerchen:
I am still not sure that this is the best implementation of GPU and CPU in the same file, but I don't have a better idea at the moment. Please feel free to suggest something if you have any ideas!

Closes #390


pancetta:
Did you test the GPU part?

tlunet:
Probably with the usage of a metaclass on the base problem level

brownbaerchen:
> Did you test the GPU part?

Yes

brownbaerchen:
> Probably with the usage of a metaclass on the mesh data level ...
> 
> Or more simply : define a xp in a pySDC module that determines if it uses np or cp on its own level, an then all the other data classes using np should use pySDC.xp instead ... this goes well with the (unfinished) discussion about reworking the base data class design

I don't understand entirely what you mean. We can discuss in Hamburg maybe. Anyways, there should not be major merge conflicts from now on. But feel free to merge your modifications to the Dahlquist problem before the rest of the paper.

tlunet:
Thanks @brownbaerchen ! And indeed, we can discuss this more in detail when you'll come to HH :wink:

PR #397, opened 2024-02-07
Parallel SDC (Reloaded) project

tlunet:
Here comes the major push with all the changes, figures etc ... for the upcoming paper on optimized diagonal SDC. With some additional (main) stuff :

- debug of the autonomous part of the RungeKutta sweepers
- addition of many new problems classes
- implementation of the `pySDC.helpers.testing.DataChecker` class, that simplifies data test regression for scripts

Incoming in some next pull request :

1. uniformisation and simplification of the fixed-point based sweepers (no IMEX), transparent from the user side but simplifying developer side
2. addition of a second order multi-step sweeper, based on previous work in my PhD (that was never added into pySDC and compared to classical sweepers)


pySDC/core/Sweeper.py
```diff
@@ -15,7 +16,7 @@
 class _Pars(FrozenClass):
     def __init__(self, pars):
         self.do_coll_update = False
-        self.initial_guess = 'spread'
+        self.initial_guess = 'spread'  # oh God, why is the default value here !?!?!?!?
```
pancetta, line 19: Yeah, no. Please remove.

pySDC/core/Sweeper.py
```diff
@@ -58,13 +59,13 @@ def __init__(self, params):
             params['collocation_class'] = CollBase
 
         # prepare random generator for initial guess
-        if params.get('initial_guess', 'spread') == 'random':
+        if params.get('initial_guess', 'spread') == 'random':  # aaaahhh ...
```
pancetta, line 62: See above.

pySDC/core/Sweeper.py
```diff
@@ -270,40 +280,102 @@ def rho(x):
...
+                """Function fitting given coefficients to a power law"""
+
+                def law(ab):
+                    a, b = ab
+                    return np.linalg.norm(a * nodes**b - coeffs)
+
+                sol = sp.optimize.minimize(law, [1.0, 1.0], method="nelder-mead")
+                return sol.x
+
+            # TODO : put this somewhere as a parameter ...
```
pancetta, line 339: So?

pySDC/core/Sweeper.py
tlunet: See above

pySDC/core/Sweeper.py
tlunet: Removed it (eventually put it back later, but that's just for experiments)

pySDC/core/Sweeper.py
```diff
@@ -270,40 +280,96 @@ def rho(x):
...
+            # Main function to compute coefficients
+            def computeCoeffs(m, a=None, b=None):
+                coll = CollBase(num_nodes=m, node_type=nodeType, quad_type=quadType)
+
+                Q = coll.Qmat[1:, 1:]
+                nodes = coll.nodes
+
+                nCoeffs = len(nodes)
+                if quadType in ['LOBATTO', 'RADAU-LEFT']:
+                    nCoeffs -= 1
```
brownbaerchen, line 297: if you define `nCoeffs` after removing the first node, you don't need to do this again.

pySDC/core/Sweeper.py
```diff
@@ -270,40 +280,96 @@ def rho(x):
...
         elif qd_type == "MIN-SR-S":
+
             M = QDmat.shape[0] - 1
-            Q = coll.Qmat[1:, 1:]
-            nodes = coll.nodes
+            quadType = coll.quad_type
+            nodeType = coll.node_type
+
+            # Main function to compute coefficients
+            def computeCoeffs(m, a=None, b=None):
```
brownbaerchen, line 289: the function signature is very nondescript. Without understanding the entire thing, I have no indication what `a` and `b` might be. In fact, as I go from the top, I don't get it.

pySDC/core/Sweeper.py
```diff
@@ -270,40 +280,96 @@ def rho(x):
...
+                coll = CollBase(num_nodes=m, node_type=nodeType, quad_type=quadType)
+
+                Q = coll.Qmat[1:, 1:]
+                nodes = coll.nodes
```
brownbaerchen, line 293: same comment as with `coll`, I think. Not super sure, honestly.

pySDC/core/Sweeper.py
```diff
@@ -270,40 +280,96 @@ def rho(x):
...
 
-            def func(coeffs):
-                coeffs = np.asarray(coeffs)
-                kMats = [(1 - z) * np.eye(nCoeffs) + z * np.diag(1 / coeffs) @ Q for z in nodes]
-                vals = [np.linalg.det(K) - 1 for K in kMats]
-                return np.array(vals)
+                # Handle first node equal to zero
+                if quadType in ['LOBATTO', 'RADAU-LEFT']:
+                    coeffs = np.asarray([0.0] + list(coeffs))
+                    nodes = coll.nodes
```
brownbaerchen, line 325: Are you adding back the first node here? It's a bit confusing.

pySDC/core/Sweeper.py
```diff
@@ -270,40 +280,96 @@ def rho(x):
...
+                """Function fitting given coefficients to a power law"""
+
+                def law(ab):
+                    a, b = ab
+                    return np.linalg.norm(a * nodes**b - coeffs)
+
+                sol = sp.optimize.minimize(law, [1.0, 1.0], method="nelder-mead")
+                return sol.x
+
+            # Compute incrementaly coefficients
```
brownbaerchen, line 339: typo: `Compute coefficients incrementally`

pySDC/core/Sweeper.py
```diff
@@ -270,40 +280,96 @@ def rho(x):
...
+            a, b = None, None
+            m0 = 2 if quadType in ['LOBATTO', 'RADAU-LEFT'] else 1
+            for m in range(m0, M + 1):
+                coeffs, nodes = computeCoeffs(m, a, b)
+                if m > 1:
+                    a, b = fit(coeffs * m, nodes)
```
brownbaerchen, line 345: What are you doing with these `a` and `b` coefficients after you compute them?

pySDC/core/Sweeper.py
```diff
@@ -270,40 +280,96 @@ def rho(x):
             self.parallelizable = True
 
         elif qd_type == "MIN-SR-S":
+
```
brownbaerchen, line 283: There is quite a lot of code in this `if`-`else` chain. I would advocate for splitting this up into functions and only calling them here.

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
```diff
@@ -663,21 +672,24 @@ def eval_f(self, u, t):
         self.work_counters['rhs']()
         return f
 
-    def u_exact(self, t):
+    def u_exact(self, t, u_init=None, t_init=None):
```
brownbaerchen, line 675: Actually, I don't think this function approximates anything with scipy... Please either remove the added comments and replace `u_init` and `t_init` with `**kwargs` or implement a call to the function generating scipy reference solutions in the core problem class.

pySDC/implementations/problem_classes/odeScalar.py
```diff
@@ -0,0 +1,230 @@
...
+        self._makeAttributeAndRegister(
+            'epsilon', 'nonLinear', 'newton_maxiter', 'newton_tol', 'stop_at_nan', localVars=locals(), readOnly=True
+        )
+        self.work_counters['newton'] = WorkCounter()
+        self.work_counters['rhs'] = WorkCounter()
+
+    # -------------------------------------------------------------------------
+    # g function (analytical solution), and its first and second derivative
+    # -------------------------------------------------------------------------
+    def g(self, t):
```
brownbaerchen, line 100: maybe you can make this and the following methods static. But I guess it's not really needed.

pySDC/implementations/problem_classes/odeSystem.py
```diff
@@ -0,0 +1,925 @@
...
+        self._makeAttributeAndRegister(
+            'epsilon', 'nonLinear', 'newton_maxiter', 'newton_tol', 'stop_at_nan', localVars=locals(), readOnly=True
+        )
+        self.work_counters['newton'] = WorkCounter()
+        self.work_counters['rhs'] = WorkCounter()
+
+    # -------------------------------------------------------------------------
+    # g function (analytical solution), and its first and second derivative
+    # -------------------------------------------------------------------------
+    def g(self, t):
```
brownbaerchen, line 93: Again, could be static, but who cares

pySDC/implementations/problem_classes/odeSystem.py
```diff
@@ -0,0 +1,925 @@
...
+        e = self.epsilon
+        g, g1, g2 = self.g(t), self.dg(t), self.dg2(t)
+        return np.array(
+            [[1 / (3 * dt * u**2 / e + 1), (dt * g2 + 3 * dt * g**2 * g1 / e) / (3 * dt * u**2 / e + 1)], [0, 1]]
+        )
+
+    # -------------------------------------------------------------------------
+    # pySDC required methods
+    # -------------------------------------------------------------------------
+    def u_exact(self, t, u_init=None, t_init=None):
```
brownbaerchen, line 132: Again, the documentation is false and the signature is misleading. There is no call to scipy.

pySDC/implementations/problem_classes/odeSystem.py
```diff
@@ -0,0 +1,925 @@
...
+        nvars = 2
+        super().__init__((nvars, None, np.dtype('float64')))
+
+        self._makeAttributeAndRegister(
+            'epsilon', 'newton_maxiter', 'newton_tol', 'stop_at_nan', localVars=locals(), readOnly=True
+        )
+        self.work_counters['newton'] = WorkCounter()
+        self.work_counters['rhs'] = WorkCounter()
+
+    def u_exact(self, t, u_init=None, t_init=None):
```
brownbaerchen, line 279: Again, no scipy here.

pySDC/implementations/problem_classes/odeSystem.py
```diff
@@ -0,0 +1,925 @@
...
+        """
+        # create new mesh object from u0 and set initial values for iteration
+        u = self.dtype_u(u0)
+
+        # start newton iteration
+        n, res = 0, np.inf
+        while n < self.newton_maxiter:
+
+            # form the function g with g(u) = 0
+            g = u - dt * self.eval_f(u, t) - rhs
```
brownbaerchen, line 207: If you call the right hand side function in the Newton solver, you increment the associated work counter. Maybe this is desired by you. I prefer to untangle it and decrement the RHS counter in the Newton solver. If you think it makes more sense to add it cost twice in a way, we can do that, but we should do it consistently.

I suspect this comment applies to most problems you implemented. And the behaviour is likely inconsistent with Lorenz and this point. So I deem this important for your paper.

pySDC/implementations/problem_classes/odeSystem.py
```diff
@@ -0,0 +1,925 @@
...
+        # create new mesh object from u0 and set initial values for iteration
+        u = self.dtype_u(u0)
+        eps = self.epsilon
+
+        # start newton iteration
+        n, res = 0, np.inf
+        while n < self.newton_maxiter:
+
+            x, y = u
+            f = np.array([-(2 + 1 / eps) * x + y**2 / eps, x - y * (1 + y)])
```
brownbaerchen, line 354: Why are you no calling `eval_f` here? Notice that there is no additional increment in the counter of RHS evaluations using this implementation

pySDC/implementations/problem_classes/odeSystem.py
```diff
@@ -0,0 +1,925 @@
...
+        """
+        # create new mesh object from u0 and set initial values for iteration
+        u = self.dtype_u(u0)
+
+        # start newton iteration
+        n, res = 0, np.inf
+        while n < self.newton_maxiter:
+
+            c1, c2, c3 = u
+            f = -np.array([0.013 * c1 + 1000 * c3 * c1, 2500 * c3 * c2, 0.013 * c1 + 1000 * c1 * c3 + 2500 * c2 * c3])
```
brownbaerchen, line 527: Again, why not call `eval_f`?

pySDC/implementations/problem_classes/odeSystem.py
```diff
@@ -0,0 +1,925 @@
...
+        """
+        # create new mesh object from u0 and set initial values for iteration
+        u = self.dtype_u(u0)
+
+        # start newton iteration
+        n, res = 0, np.inf
+        while n < self.newton_maxiter:
+
+            u1, u2, u3 = u
+            f = np.array([u2 * u3, -u1 * u3, -0.51 * u1 * u2])
```
brownbaerchen, line 876: Same, same.

pySDC/implementations/sweeper_classes/Runge_Kutta.py
```diff
@@ -655,6 +655,71 @@ def get_update_order(cls):
...
+    s2 = 2**0.5
+
+    nodes = np.array([0, 1 / 2, (2 - 2**0.5) / 4, 5 / 8, 26 / 25, 1.0])
+    matrix = np.zeros((6, 6))
+    matrix[1, :2] = [1 / 4, 1 / 4]  # OK
+    matrix[2, :3] = [
+        (1 - 2**0.5) / 8,
+        (1 - 2**0.5) / 8,
+        1 / 4,
+    ]  # OK
```
brownbaerchen, line 673: what's with the `OK`?

pySDC/implementations/sweeper_classes/generic_implicit.py
```diff
@@ -89,9 +98,11 @@ def update_nodes(self):
                 rhs += L.dt * self.QI[m + 1, j] * L.f[j]
 
             # implicit solve with prefactor stemming from the diagonal of Qd
-            L.u[m + 1] = P.solve_system(
-                rhs, L.dt * self.QI[m + 1, m + 1], L.u[m + 1], L.time + L.dt * self.coll.nodes[m]
-            )
+            alpha = L.dt * self.QI[m + 1, m + 1]
+            if alpha == 0:
```
brownbaerchen, line 102: nice

pySDC/implementations/sweeper_classes/generic_implicit_MPI.py
```diff
@@ -127,12 +127,20 @@ def predict(self):
...
+
         if self.params.initial_guess == 'spread':
-            L.u[self.rank + 1] = P.dtype_u(L.u[0])
-            L.f[self.rank + 1] = P.eval_f(L.u[self.rank + 1], L.time + L.dt * self.coll.nodes[self.rank])
+            # copy u[0] to all collocation nodes, evaluate RHS
+            L.u[m + 1] = P.dtype_u(L.u[0])
+            L.f[m + 1] = P.eval_f(L.u[m + 1], L.time + L.dt * self.coll.nodes[m])
+        elif self.params.initial_guess == 'copy':
+            # copy u[0] and RHS evaluation to all collocation nodes
+            L.u[m] = P.dtype_u(L.u[0])
```
brownbaerchen, line 138: Why is it `m+1` in the other cases but `m` here? Did you test that this works? I would guess not.

pySDC/projects/parallelSDC_reloaded/allenCahn_accuracy.py
```diff
@@ -0,0 +1,120 @@
...
+
+def getError(uNum, uRef):
+    if uNum is None:
+        return np.inf
+    return np.linalg.norm(uRef[-1, :] - uNum[-1, :], ord=2)
+
+
+def getCost(counters):
+    nNewton, nRHS, tComp = counters
+    return 2 * nNewton + nRHS
```
brownbaerchen, line 37: Did you check that right hand side evaluations are not counted additionally in the Newton solver?

brownbaerchen (changes requested):
I didn't really look through the plotting scripts.
Looks good to me for the most part, I just have a few points.
 - There may be inconsistencies in how the RHS evaluations are counted in combination with Newton iterations
 - The documentation for some of the `u_exact` functions is wrong and does not do anything to alert the user that what comes back is not what it is expected.

pySDC/core/Sweeper.py
tlunet: Corrected

pySDC/core/Sweeper.py
tlunet: Thanks, changed

pySDC/core/Sweeper.py
tlunet: More documentation have been added

pySDC/core/Sweeper.py
tlunet: Switched to `QM` notation

pySDC/core/Sweeper.py
tlunet: Yes ... changed to an explicit (although less efficient) solution.

pySDC/core/Sweeper.py
tlunet: See documentation in the `computeCoefficients` function

pySDC/core/Sweeper.py
tlunet: Indeed, but this will come with an other PR later, including a simplification of the QDelta matrix generation ...

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
tlunet: Docstring has been corrected

pySDC/implementations/problem_classes/odeSystem.py
tlunet: Not so sure ... this is related to the problem instance anyway (in particular if one wants to add a parameter). More in the spirit of "who cares" then

pySDC/implementations/problem_classes/odeSystem.py
tlunet: Corrected

pySDC/implementations/problem_classes/odeScalar.py
tlunet: indeed, see previous comment

pySDC/implementations/problem_classes/odeSystem.py
tlunet: Corrected

pySDC/implementations/problem_classes/odeSystem.py
tlunet: Indeed, but this is the only place were it is used. In fact, the RHS is already evaluated before, but not passed to the `solve_system` function. So it make sense not to use the `eval_f` function (as it is done in all the other classes).
This has been corrected now, it only calls the `f` method, which does not increment the RHS counter.

Furthermore,  passing the RHS evaluation in `system_solve` will probably be something to add to the problem rework, see #255

pySDC/implementations/problem_classes/odeSystem.py
tlunet: See previous comment

pySDC/implementations/problem_classes/odeSystem.py
tlunet: Same same

pySDC/implementations/sweeper_classes/Runge_Kutta.py
tlunet: Some previous consistency checks, that have been removed.

pySDC/implementations/sweeper_classes/generic_implicit.py
tlunet: thanks

pySDC/implementations/sweeper_classes/generic_implicit_MPI.py
tlunet: Just took the previous implementation (which was not tested, btw ...). Due to this ghost point in the Q matrix, that is not in the nodes ....

pySDC/projects/parallelSDC_reloaded/allenCahn_accuracy.py
tlunet: They are not (cf previous discussion)

pySDC/implementations/problem_classes/odeSystem.py
tlunet: See previous comment

pySDC/implementations/problem_classes/odeSystem.py
tlunet: Hum, actually I was wrong : the initial evaluation of the RHS can only be used the first time, but then it should be evaluated at each newton iteration. But this does not change the paper results, as this problem class is not used in the numerical experiments.

Also, for most problem class, it does not change the fact that the Newton iteration has almost the same cost as the RHS evaluation ...

tlunet:
Data for figures are tested with [56cb776](https://github.com/Parallel-in-Time/pySDC/pull/397/commits/56cb7768861aa67b73cd18b934f287d63503adbf) (and the few commits that came later).

One thing tho : Nan and Infinite numpy values are not read correctly in JSON file for python < 3.8, to one test (figures for Allen-Cahn) is not executed for python=3.7

brownbaerchen (approved):

pySDC/implementations/sweeper_classes/generic_implicit_MPI.py
brownbaerchen: But there was no previous implementation of your new feature :D I am 72% sure that this will fail. Please test it!

tlunet:
@brownbaerchen @pancetta I think I'm finish. Let me know if you see any improvement that can be done, or else I guess you can merge it

tlunet:
nvm

tlunet:
Should be good now

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
```diff
@@ -224,21 +230,24 @@ def eval_f(self, u, t):
...
```
brownbaerchen, line 233: Sorry for nitpicking, but putting `u_init` and `t_init` in the signature implies that you can use this function to compute a local error, when really you can't. The function will ignore these variables and give the analytical exact solution starting from t=0 regardless. So I suggest replacing them with `**kwargs` to avoid confusion. Also there is a typo in the doctoring regarding `u_init` below.

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
brownbaerchen: Should have been more precise about the role of `u_init` and `t_init` earlier. Sorry for that!

pySDC/implementations/problem_classes/odeScalar.py
```diff
@@ -0,0 +1,213 @@
...
+    def jac_LIN(self, u, t):
+        return -self.epsilon ** (-1)
+
+    def jac_NONLIN(self, u, t):
+        return -self.epsilon ** (-1) * 3 * u**2
+
+    # -------------------------------------------------------------------------
+    # pySDC required methods
+    # -------------------------------------------------------------------------
+    def u_exact(self, t, u_init=None, t_init=None):
```
brownbaerchen, line 117: Same thing with `u_init` and `t_init`. Maybe there is a VS code functionality that lets you do all these at once? :P

pySDC/implementations/problem_classes/odeSystem.py
brownbaerchen: I think you forgot to switch `eval_f` for `f` in assembling `g`? `f` is not used, I think. You can also do what I did [here](https://github.com/Parallel-in-Time/pySDC/blob/a8a04311a0ce594fed24050c640866809ba6e870/pySDC/implementations/problem_classes/Quench.py#L327), to decrement the work counter for RHS if you prefer.

pySDC/implementations/sweeper_classes/generic_implicit.py
```diff
@@ -65,6 +65,15 @@ def update_nodes(self):
         # get number of collocation nodes for easier access
         M = self.coll.num_nodes
 
+        # update the MIN-SR-FLEX preconditioner
```
brownbaerchen, line 68: How specific is this to the generic implicit sweeper? Would it make sense to put this in a function `updatePreconditioner(self, k)` or whatever in the core Sweeper module that can be called in `update_nodes` of all sweepers? This irks me a bit because it is not really related to updating the nodes. We could consider putting this functionality in a convergence controller. There you wouldn't need the hack with the strings as well.
Having a preconditioner be a convergence controller is not great. If we can find a way of automatically adding this when the flex preconditioner is used in the sweeper parameters it might be cleaner than this, though. Again, sorry for not raising this point earlier. What do you think?

brownbaerchen (commented):
I read a bit more carefully this time. I think you should change the function signatures because it's quick and may confuse people otherwise. Regarding the implementation of the flex preconditioner, that's up to you. I don't see a perfect implementation for this. I do think a convergence controller would be cleaner than implementing this in `update_nodes` in every sweeper. But do as you see fit.

pySDC/implementations/sweeper_classes/generic_implicit.py
tlunet: This is indeed not generic at all, but I would leave it as it is for now : I will make it more generic when merging the explicit and implicit sweeper and regrouping the QDelta generation in a specific module out of the sweeper (because, there is no reason for the sweeper to be defined with QDelta matrices, one can also use a node-to-node formulation to define a sweeper, or an other approach that do not involve QDelta matrices).

This is just a quick solution which also makes sense for now : the FLEX preconditioning has been developed in an implicit only settings, so it makes sense to being able to use it with the implicit only sweeper.

pySDC/implementations/problem_classes/AllenCahn_1D_FD.py
tlunet: Don't really like that the signature of `u_exact` can be different depending on the problem, it breaks genericity. 
Also, I'm not sure that the signature should indicate what the method does, when comparing two methods that have the same name.

Actually, I thought that those `u_init` and `t_init` argument could be used to reduce computation time, when computing reference solution with scipy on many time points, which have no impact on `u_exact` method using an analytical reference solution but then is kept for consistency.

I would prefer indicating in the docstring that those `u_init` and `t_init` are not used. Modifying this with **kwargs add more ambiguity than everything else. But if you really disagree with keeping those, then I'll simply remove it and adapt the project code using the `u_exact` generically with a `try` and `catch` block ...

tlunet: Did not receive any comment from your side, so I'll just comply and remove the `u_init` and `t_init` from the signature (I spend quite some time on this PR, this should be merged at some point to submit the paper).

But this just delay some serious discussion we should have at some point about uniformizing the problem classes ...

pancetta:
Are we good now? I have no strong feelings concerning the function signature problem. I agree with @tlunet concerning the flex preconditioning as part of the project implicit sweeper only. Some of these decisions can we taken back easily and indeed, some will be changed when we have „the serious discussion“ (and time for the consequences).

tlunet:
lgtm ... I can still propose some other small PR for finer details later

pancetta:
@brownbaerchen ?

PR #401, opened 2024-02-19
2D Brusselator problem

brownbaerchen:
This is a reaction diffusion equation with time-dependent source term. Taken from [here](https://link.springer.com/book/10.1007/978-3-642-05221-7). Thanks @grosilho for the suggestion!
The solution strategy is IMEX splitting: Integrate the Laplace operator implicitly using mpi4py-fft and the rest explicitly.

Because I had to verify the implementation visually, I added some framework for plotting. Let me know if you want this removed.

The test is based on a suggestion from @tlunet. The idea is to estimate the error by doing an implicit (IMEX in this case) Euler step forward and then do an explicit Euler step backward. Because it's backward, it's stable and we should return to the initial conditions somewhat. I check that the order of the thus derived local error estimate is 2. How do you like this? It's not uber meaningful for this IMEX problem, I guess. It makes more sense for Newton solvers. But I didn't want zero coverage :D

### DIFF
--- a/pySDC/core/Problem.py
+++ b/pySDC/core/Problem.py
@@ -135,3 +135,32 @@ class ptype(RegisterParams):
 
         u_shape = u_init.shape
         return solve_ivp(eval_rhs, (t_init, t), u_init.flatten(), **kwargs).y[:, -1].reshape(u_shape)
+
+    def get_fig(self):
+        """
+        Get a figure suitable to plot the solution of this problem
+
+        Returns
+        -------
+        self.fig : matplotlib.pyplot.figure.Figure
+        """
+        raise NotImplementedError
+
+    def plot(self, u, t=None, fig=None):
+        r"""
+        Plot the solution. Please supply a figure with the same structure as returned by ``self.get_fig``.
+
+        Parameters
+        ----------
+        u : dtype_u
+            Solution to be plotted
+        t : float
+            Time to display at the top of the figure
+        fig : matplotlib.pyplot.figure.Figure
+            Figure with the correct structure
+
+        Returns
+        -------
+        None
+        """
+        raise NotImplementedError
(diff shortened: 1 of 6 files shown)


pySDC/core/Problem.py
```diff
@@ -135,3 +135,32 @@ def generate_scipy_reference_solution(self, eval_rhs, t, u_init=None, t_init=Non
 
         u_shape = u_init.shape
         return solve_ivp(eval_rhs, (t_init, t), u_init.flatten(), **kwargs).y[:, -1].reshape(u_shape)
+
```
pancetta, line 138: Why do these functions have to be part of the core problems class?

pySDC/tests/test_problems/test_Brusselator.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+
+
+def test_Brusselator():
```
pancetta, line 4: Missing the `pytest` marks, test did not run.

pySDC/core/Problem.py
brownbaerchen: I think it's a nice feature wrt prototyping to just do `prob.plot(u); plt.show()` and get a nice plot of the solution, complete with labels etc. In this case, for instance, you get plots of the two components side by side. I would like this to be implemented in other problems in the future as well. But I can remove it if you don't want it! 
Another advantage is that you can have hooks such as the one included in this PR, which you can just give to the controller and it plots the solution, as intended by the author of the problem class, in regular intervals.

pancetta:
Please merge `master` again, the CI does strange things at the moment.

pancetta:
No testing of the `Brusselator.py`?

brownbaerchen:
> No testing of the `Brusselator.py`?

Well, this file is not really doing anything at the moment. It's just running an arbitrary example. I wouldn't know how to test it. I can remove the file and add it only later when I actually do sth with the Brusselator if you want.

pancetta:
Keeping it will add untested stuff to the project, which is something we try to avoid. So, it would be good if you either test it or remove it for now.

PR #403, opened 2024-02-27
Fix for RK sweeper - Node changed for `BackwardEuler`

lisawim:
One of the biggest PRs ever seen in ``pySDC``. Node for implicit Euler is changed. Since it is an implicit method it needs to be changed to 1.


brownbaerchen:
I have 400 complaints. We are going to need at least 6 months before we can merge this.

lisawim:
Hm, I think I should add lot of documentation.

PR #404, opened 2024-03-05
TIME-X Test Hackathon @ TUD: Test for `SwitchEstimator`

lisawim:
Since the TIME-X project comes to the end, it is a good time to push the results from the [test hackathon](https://time-x.eu/events/). In Darmstadt, I learnt interesting things about testing! So, I started writing a test for the ``SwitchEstimator`` to test all the components of it. Here, for most of the tests regarding the ODE case I created a dummy problem called ``ExactDiscontinuousTestODE`` inheriting from ``DiscontinuousTestODE`` that returns the exact dynamics of the parent class, i.e.,

- ``solve_system`` returns the exact solution,

- ``eval_f`` returns the (exact) derivative.

I also tested this dummy problem separately (which is probably not necessary, but tested stuff is better than untested stuff..)

Now, there are the following tests:

- ``testExactDummyProblem``: as mentioned, here the exact dummy problem is tested if it provides the dynamics that it should supposed to do.

- ``testAdaptInterpolationInfo``: In the SE, the interpolation x-axis and corresponding values of state function needs to be adapted either ``'LOBATTO'`` or ``'RADAU-RIGHT'`` is used. When ``'LOBATTO'`` the first value of the state function is removed since it is double (since $t_0 = \tau_1$). When ``'RADAU-RIGHT'`` we can also use the value of the state function at $t_0$ for the interpolation, i.e., $t_0$ is added to the interpolation x-axis. Then, the tests check if the x- and y- values for interpolation do have the same length and also if the stuff is satisfied that is explained above.

- ``testDetectionBoundary``: In this test, the step size is set s.t. the event does occur on the boundary. In this case the SE should not executed a restart which is tested there.

- ``testDetectionODE``: Now, this test is a fully application of the SE to ``DiscontinuousTestODE``, i.e., an integration test, where tolerances are checked here. 

- ``testDetectionDAE``: Since the SE should also be applied to the DAE case, I also provided a test where the SE is applied to the ``DiscontinuousTestDAE``. Here, tolerances are checked for the error and the event time error.

I found out that a backward finite difference to approximate the derivative of the state function does results in better results, i.e., having a better accuracy here. 

[Here](https://github.com/lisawim/pySDC/issues/7), more information can be found if I forgot something..


pySDC/projects/PinTSimE/estimation_check.py
```diff
@@ -239,15 +242,19 @@ def plotStateFunctionAroundEvent(u_num, prob_cls_name, M_fix):  # pragma: no cov
...
 
+                        t_switch = [t_event[i] for t_event in t_switches]
                         ax[0, ind].plot(
                             dt_list,
                             [
                                 h_item[m]
                                 for (t_item, h_item, t_switch_item) in zip(t, h, t_switch)
                                 for m in range(len(t_item))
-                                if abs(t_item[m] - t_switch_item) <= 1e-14
+                                if abs(t_item[m] - t_switch_item) <= 2.7961188919789493e-11
```
brownbaerchen, line 257: How did you arrive at this number? :D You could use something like the 10 times the `eps` attribute of float (see [here](https://numpy.org/doc/stable/reference/generated/numpy.finfo.html)) if that makes sense?

pySDC/tests/test_projects/test_DAE/test_problems.py
```diff
@@ -443,17 +443,15 @@ def test_DiscontinuousTestDAE_SDC_detection(M):
     from pySDC.implementations.convergence_controller_classes.basic_restarting import BasicRestartingNonMPI
 
     err_tol = {
-        2: 5.3952e-9,
-        3: 2.6741e-9,
-        4: 1.9163e-8,
-        5: 2.4791e-8,
+        3: 5.97e-13,
```
brownbaerchen, line 446: With tolerances so low, I am worried that this test might not work on certain architectures. I believe you are not guaranteed the same machine error below machine precision on different operating systems or CPUs.

pySDC/tests/test_projects/test_pintsime/test_SwitchEstimator.py
```diff
@@ -0,0 +1,464 @@
...
+        controller_params=controller_params,
+        controller=controller,
+        t0=t0,
+        Tend=Tend,
+        exact_event_time_avail=exact_event_time_avail,
+    )
+
+    # in this specific example only one event has to be found
+    switches = [me[1] for me in get_sorted(stats, type='switch', sortby='time', recomputed=False)]
+    assert len(switches) >= 1, f'{problem.__name__}: No events found for tol={tol}!'
```
brownbaerchen, line 363: In the Jülich chat someone recently posted that you can do string formatting like `{tol=}` and it will show as "toll=3.14" or whatever. Of course, you don't need to change this minor detail. But I use it all the time now and want to pass this on :)

pySDC/tests/test_projects/test_pintsime/test_SwitchEstimator.py
```diff
@@ -0,0 +1,464 @@
...
+    dt = Tend - t0
+    tol = 1e-10
+
+    sweeper = fully_implicit_DAE
+    QI = 'LU'
+    quad_type = 'RADAU-RIGHT'
+
+    _, _, _, _, useA, useSE, exact_event_time_avail = getParamsRun()
+
+    restol = 1e-13
```
brownbaerchen, line 414: A residual tolerance this low will probably enforce maximum iterations on some architectures. Maybe this test could be a bit faster with a slightly larger tolerance or a lower number of maximum iterations?

pySDC/tests/test_projects/test_pintsime/test_SwitchEstimator.py
```diff
@@ -0,0 +1,464 @@
...
+    assert len(switches) >= 1, f'{problem.__name__}: No events found for tol={tol} and M={num_nodes}!'
+
+    t_switch = switches[-1]
+    event_err = abs(t_switch - t_switch_exact)
+    assert np.isclose(event_err, 0, atol=2.2e-6), f'Event time error {event_err} is not small enough!'
+
+    h = np.array([val[1] for val in get_sorted(stats, type='state_function', sortby='time', recomputed=False)])
+    if h[-1] < 0:
+        assert abs(h[-1]) < 4.7e-10, f"State function has large negative value -> SE does switch too early! Got {h[-1]}"
+    assert np.isclose(abs(h[-1]), 0.0, atol=5e-10), f'State function is not close to zero; value is {h[-1]}'
```
brownbaerchen, line 459: Is this not the same as the above statement?

pySDC/projects/PinTSimE/estimation_check.py
lisawim: I tried that out.. but for the current implementation of the SE I don't know if such an statement makes sense anymore. Since you use the $\alpha\leq 1$ (and probably $\alpha<1$) the event time ``t_switch_item`` here lies more far away from the logged time(?)

pySDC/tests/test_projects/test_DAE/test_problems.py
lisawim: Mhm, yes. I agree. Maybe the tolerances should be set larger to guarantee that, I'll change that!

pySDC/tests/test_projects/test_pintsime/test_SwitchEstimator.py
lisawim: Ok, I'll try that out!

pySDC/tests/test_projects/test_pintsime/test_SwitchEstimator.py
lisawim: Hm. Ah, yes. Since the state function here only depends on the differential variable, we could use a larger residual tolerance to speed up the test a bit. Thanks for the hint!

pySDC/tests/test_projects/test_pintsime/test_SwitchEstimator.py
```diff
@@ -0,0 +1,462 @@
+import numpy as np
+import pytest
+
+from pySDC.implementations.problem_classes.DiscontinuousTestODE import DiscontinuousTestODE
+from pySDC.projects.DAE.problems.DiscontinuousTestDAE import DiscontinuousTestDAE
```
pancetta, line 5: Please do not import these things in a global way, otherwise other tests fail due to an MPI problem (here: fenics). Put these imports into the test functions.

pancetta:
Sorry for being late here: are we ready to merge? @lisawim @brownbaerchen

pySDC/projects/PinTSimE/estimation_check.py
```diff
@@ -239,15 +242,19 @@ def plotStateFunctionAroundEvent(u_num, prob_cls_name, M_fix):  # pragma: no cov
 
                     if use_SE:
                         t_switches = [u_num[dt][M_fix][use_SE][use_A]['t_switches'] for dt in dt_list]
-                        t_switch = [t_event[i] for t_event in t_switches]
+                        for t_switch_item in t_switches:
+                            for m in range(1, len(t_switch_item)):
```
brownbaerchen, line 246: how many iterations does this typically have? If it's a lot, you may want to speed this up using masks with sth. like
```
mask = np.append([True], np.abs(t_swtich_item[1:] - t_switch_item[:-1] > 1e-10)
t_switch_item = t_switch_item[mask]
```
If the loop is entered only a couple of times, don't bother.

brownbaerchen:
Let's merge!

PR #406, opened 2024-03-08
`pySDC`-build in `LagrangeApproximation` class in `SwitchEstimator`

lisawim:
Since we already have a Lagrange interpolation class I removed the afterwards implemented Lagrange class in `SwitchEstimator` and replaced the Lagrange interpolation there by @tlunet's `LagrangeApproximation` class. For finding the root a callable function is necessary, thus the`__call__` method is added as well.


tlunet:
Looks good to me ! :grinning:

PR #407, opened 2024-03-20
Monodomain project

grosilho:
A pySDC project where we solve the monodomain equation in cardiac electrophysiology using the exponential SDC method.


tlunet:
Hi @grosilho, thanks for this PR ! Just so you know, I've activated the CI runs for this PR, so every time you push on the `monodomain_skinny` branch, it will re-trigger all the tests. You should see all the results, coverage report etc ... soon automatically added to the PR.

brownbaerchen (commented):
Pretty cool stuff!

This is quite a large PR and I don't really know how the exponential stuff works. So I did not try to confirm that the code does what it says. I trust you and your tests.

I marked a lot of little typos. Doesn't matter too much if you fix them or not, I guess.

The one big thing that I noticed is the datatypes which seem a bit cumbersome. At the end of the day, the code does what you want and it is entirely contained within your project. So I understand if you don't want to refactor, and I guess it's not really necessary.
That being said, I do think that a datatype derived from `mesh` similar to the `MultiComponentMesh` datatype would allow to shorten the code and make it faster because you can use all the efficient numpy implementations. But this is only true if the length of the vectors is the same in each vector of vectors.
Let me know if you are interested in this and need help. We only implemented the `MultiComponentMesh` datatype on the train ride back from Bruges, actually.

pySDC/projects/Monodomain/transfer_classes/TransferVectorOfDCTVectors.py
```diff
@@ -0,0 +1,48 @@
+from pySDC.core.SpaceTransfer import space_transfer
+from pySDC.projects.Monodomain.transfer_classes.Transfer_DCT_Vector import DCT_to_DCT
+from pySDC.projects.Monodomain.datatype_classes.VectorOfVectors import VectorOfVectors, IMEXEXP_VectorOfVectors
+
+
+class TransferVectorOfDCTVectors(space_transfer):
+    """
+    This implementation can restrict and prolong between super vectors
```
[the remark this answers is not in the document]
grosilho: yeah, I meant vector of vectors. Fixed.

pySDC/tests/test_projects/test_monodomain/test_monodomain_iterations.py
```diff
@@ -0,0 +1,145 @@
...
+    read_init_val = False
+    init_time = 0.0
+    end_time = 2.0
+    write_as_reference_solution = False
+    write_all_variables = False
+    output_file_name = "monodomain"
+    ref_sol = ""
+
+    err, rel_err, avg_niters, times, niters, residuals = setup_and_run(
+        integrator,
```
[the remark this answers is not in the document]
grosilho: yes indeed it looks better now. Fixed.

pySDC/tests/test_projects/test_monodomain/test_monodomain_iterations_parallel.py
```diff
@@ -0,0 +1,320 @@
...
+    print(f"Got average number of iterations {iters_info['avg_niters']}, expected was {expected_avg_niters}")
+
+    assert iters_info['avg_niters'] == pytest.approx(
+        expected_avg_niters, rel=0.1
+    ), f"Average number of iterations {iters_info['avg_niters']} too different from the expected {expected_avg_niters}"
+
+    return iters_info
+
+
+def move_coverage_folders():
```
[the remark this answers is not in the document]
grosilho: Indeed it is not needed anymore. Removed

pySDC/projects/Monodomain/sweeper_classes/exponential_runge_kutta/imexexp_1st_order.py
```diff
@@ -0,0 +1,393 @@
...
+        # check if Mth node is equal to right point and do_coll_update is false, perform a simple copy
+        if self.coll.right_is_node and not self.params.do_coll_update:
+            # a copy is sufficient
+            L.uend = P.dtype_u(L.u[-1])
+        else:
+            raise CollocationError("This option is not implemented yet.")
+
+        return None
+
+    def predict(self):
```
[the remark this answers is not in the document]
grosilho: I needed to overload to update the status of the lambda and yinf variables. But it it not needed anymore. Removed.

pySDC/projects/Monodomain/sweeper_classes/exponential_runge_kutta/imexexp_1st_order.py
```diff
@@ -0,0 +1,393 @@
...
+        L.status.unlocked = True
+        L.status.updated = True
+
+    #     self.update_lmbda_yinf_status(outdated=True)
+
+    # def update_lmbda_yinf_status(self, outdated):
+    #     if not self.level.prob.constant_lambda_and_phi and outdated:
+    #         self.lambda_and_phi_outdated = True
+
+    def compute_residual(self, stage=''):
```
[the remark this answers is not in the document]
grosilho: It's not crucial but without this norm the residual might be underestimated.

pySDC/projects/Monodomain/sweeper_classes/runge_kutta/imexexp_1st_order.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class imexexp_1st_order(sweeper):
+    """
+    Custom sweeper class, implements Sweeper.py
+
+    First-order IMEXEXP sweeper using implicit/explicit/exponential Euler as base integrator
+    In the cardiac electrphysiology community this is known as Rush-Larsen scheme.
+
+    The underlying intergrator is collocation method, leading to standard SDC.
```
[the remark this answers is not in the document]
grosilho: It was to point out the difference with respect to the other sweeper that uses exponential Runge-Kutta. I removed it

pySDC/projects/Monodomain/problem_classes/MonodomainODE.py
```diff
@@ -0,0 +1,412 @@
...
+    def write_reference_solution(self, uh, all=False):
+        # write solution to file, only the potential V=uh[0] or all variables if all=True
+        self.parabolic.write_reference_solution(uh, list(range(uh.size)) if all else [0])
+
+    def read_reference_solution(self, uh, ref_file_name, all=False):
+        # read solution from file, only the potential V=uh[0] or all variables if all=True
+        # returns true if read was successful, false else
+        return self.parabolic.read_reference_solution(uh, list(range(uh.size)) if all else [0], ref_file_name)
+
+    def initial_value(self):
```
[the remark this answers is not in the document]
grosilho: Ok, I can easily change the function name, but then should I implement the reading of the reference solution too? I have the reference solution at the final time only and depending on the experiment the final time changes.

grosilho:
Hey @brownbaerchen , I went through most of your comments. It remains the MultiComponentMesh stuff that I have to look into. I will get back to you asap :)

pySDC/projects/Monodomain/problem_classes/MonodomainODE.py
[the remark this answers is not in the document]
brownbaerchen: I see. In that case, it would probably be misleading to load the reference solution in the `u_exact` function.

pySDC/projects/Monodomain/sweeper_classes/exponential_runge_kutta/imexexp_1st_order.py
```diff
@@ -0,0 +1,368 @@
...
+        b = self.phi_coll.weights
+
+        # iterate only over the indeces of the problem having an exponential term
+        # hence we update only the subvectors phi_k(factor*lambda)_i in the vector of vectors phi_k(factor*lambda)
+        for i in P.rhs_exp_indeces:
+            # iterate over all the factors
+            for n in range(N_fac):
+                factor = factors[n]
+                # compute e^((1-c_j)*factor*lmbda) for nodes c_j on the quadrature rule
+                exp_terms = [np.exp(((1.0 - c[j]) * factor) * lmbda[i]) for j in range(self.phi_coll.num_nodes)]
```
grosilho, line 93: This line for instance is slow, with lmbda a mesh instance of shape (19,512), for instance. With 19 the number of model unknowns and 512 the mesh size

grosilho:
Hi @brownbaerchen, I replaced my datatype with `mesh` and indeed I can save some lines of code, but it is much slower (~0.5x). I profiled and it seems that the overloading of __ufunc__ in mesh is slow, is that correct? 
See the above comment for an example of where it is slower than before :)

pySDC/projects/Monodomain/sweeper_classes/exponential_runge_kutta/imexexp_1st_order.py
pancetta: I think we need a list comprehension here, so that Numpy can do it's magic. Do-loops are evil in Python.

pySDC/projects/Monodomain/datatype_classes/my_mesh.py
```diff
@@ -0,0 +1,9 @@
+from pySDC.implementations.datatype_classes.mesh import MultiComponentMesh
+
+
+class imexexp_mesh(MultiComponentMesh):
+    components = ['impl', 'expl', 'exp']
+
+
+def get_np_list(vec):
```
brownbaerchen, line 8: What is the purpose of this function? I assume `vec` should be of time `imexexp_mesh`, so so you can directly interface the components from that instead of splitting it apart into a list first.

pySDC/projects/Monodomain/problem_classes/TestODE.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # set lambdas, if not provided by user
+        if not hasattr(self, 'lmbda_laplacian'):
+            self.lmbda_laplacian = -5.0
+        if not hasattr(self, 'lmbda_gating'):
+            self.lmbda_gating = -10.0
+        if not hasattr(self, 'lmbda_others'):
+            self.lmbda_others = -1.0
+
+        self.dtype_u = mesh
```
brownbaerchen, line 47: In other problem classes these are class attributes in pySDC. Previously, there was some instance-specific logic, but if there isn't, it's better to make this a class attribute for constituency.

pySDC/projects/Monodomain/sweeper_classes/exponential_runge_kutta/imexexp_1st_order.py
brownbaerchen: If I understand correctly, the problem is the outer loop over `P.rhs_exp_indeces`. You are solving different components with different methods. So `dtype_u` should not be `mesh` but some sort of `MultiComponentMesh` as well. I guess, specific to the monodomain equation, you would have
```
class MonodomainMesh(MultiComponentMesh):
    components = ['V', 'z'] 
```
or something like that.
By using masks, you should be much faster than with an explicit for loop, that is to say the code in the sweeper should be 
```
for n in range(N_fac):
    factor = factors[n]
    # compute e^((1-c_j)*factor*lmbda) for nodes c_j on the quadrature rule
    exp_terms = [np.exp(((1.0 - c[j]) * factor) * lmbda.z) for j in range(self.phi_coll.num_nodes)]
    ...
    phi[m][n].z = np.exp(factor * lambda.z)
    ...
```
At this point, you will run into the issue that you want, in general, different shapes for V and z, I assume. But this can surely be dealt with somehow.

Now that I wrote all of this, I am not sure that is what is slow, though. After all, the for loop is run only over few indices. I find this very confusing, to be honest. Can you check if `lambda[i]` owns its memory by printing `lambda[I].flags`? We don't want it to, as we want a view on the original data. Can you send me the profile? Did you use cProfile?

brownbaerchen (commented):
:shrug:

pySDC/projects/Monodomain/sweeper_classes/exponential_runge_kutta/imexexp_1st_order.py
pancetta: But this is O(indices * num_nodes * N_fac). The hope would be to get this into one big numpy expression with lots of `[:]` to get this to full speed (with threading).

pySDC/projects/Monodomain/datatype_classes/my_mesh.py
grosilho: yes, it was a temporary trick for compatibility with the c++ interface. Now it's fixed.

pySDC/projects/Monodomain/problem_classes/TestODE.py
grosilho: agree, fixed

pySDC/projects/Monodomain/sweeper_classes/exponential_runge_kutta/imexexp_1st_order.py
grosilho: I did like Robert suggested :)

pancetta: Great job!!

pancetta:
Looks great! Any objections, @brownbaerchen?

brownbaerchen:
Nope, let's merge!

PR #414, opened 2024-04-21
SDC-DAE sweeper for semi-explicit DAEs

lisawim:
This PR contains the new SDC-DAE sweeper especially for DAEs of the semi-explicit form $$y'=f(y, z),$$ $$0 = g(y,z),$$ as promised. Tests are also added, where I orientated on the tests of ``fully_implicit_DAE`` (btw the test of this sweeper is also renamed in this PR). Hm, what else?.. I also added a test that checks the order of accuracy for a DAE of index $1$ and a DAE of index $2$. Note that for index- $2$ DAEs there is order reduction in the algebraic variable, which is satisfied here.


pySDC/projects/DAE/sweepers/SemiImplicitDAE.py
```diff
@@ -0,0 +1,190 @@
...
+        """Initialization routine"""
+
+        if 'QI' not in params:
+            params['QI'] = 'IE'
+
+        # call parent's initialization routine
+        super().__init__(params)
+
+        msg = f"Quadrature type {self.params.quad_type} is not implemented yet. Use 'RADAU-RIGHT' instead!"
+        if self.coll.left_is_node:
```
brownbaerchen, line 96: Shouldn't this be `if not self.coll.right_is_node:`? I assume you don't want the collocation update..?

pySDC/projects/DAE/sweepers/SemiImplicitDAE.py
```diff
@@ -0,0 +1,190 @@
...
+        P = L.prob
+
+        # only if the level has been touched before
+        assert L.status.unlocked
+        M = self.coll.num_nodes
+
+        integral = self.integrate()
+        # build the rest of the known solution u_0 + del_t(Q - Q_del)U_k
+        for m in range(1, M + 1):
+            for j in range(1, M + 1):
```
brownbaerchen, line 141: this range should only go to $m+1$. All other entries in the preconditioner are zero. And we don't need to multiply many solution size objects by zero and then add that. It's also done like this in the generic implicit sweeper, but I think we should rather change it in both. The rest of this function depends on the lower triangular structure of `QI`, such that this does not reduce generality, if I am not mistaken.

pySDC/tests/test_projects/test_DAE/test_SemiImplicitDAE.py
```diff
@@ -0,0 +1,354 @@
...
+    S = step(description=description)
+    L = S.levels[0]
+    P = L.prob
+
+    assert isinstance(L.sweep, SemiImplicitDAE), "Sweeper cannot instantiate an object of type SemiImplicitDAE!"
+
+    L.status.time = 1.0
+    L.u[0] = P.u_exact(L.time)
+
+    assert isinstance(L.f[0], type(None)), "Initial condition f0 is not of type None!"
```
brownbaerchen, line 84: Why do you test this? This seems to be a design choice irrespective or your sweeper.

brownbaerchen (commented):
Good job! :)

pySDC/projects/DAE/sweepers/SemiImplicitDAE.py
lisawim: We also wouldn't like to use ``'LOBATTO'`` since it is not tested yet and for ``'LOBATTO'`` an initial condition for ``du`` is needed.

pySDC/tests/test_projects/test_DAE/test_SemiImplicitDAE.py
lisawim: True. I'll remove that.

brownbaerchen:
LGTM

Issue #415, opened 2024-04-22
Couple with Gitlab-Repo


jakob-fritz:
Mirror repository with Gitlab-Repo so that CI can use HPC-Resources.
This is helpful e.g. for running GPU-Tests and/or for running Benchmarks.

Tasks that should be done:

- [x] Mirror with Gitlab for new pushes
- [x] Mirror also for updated Pull-Requests (only if certain tags are set and people with appropriate permissions triggered CI)
- [x] Run tests on HPC (e.g. GPU-Tests)
- [x] Get back Reports (e.g. coverage) from Gitlab and include in Github-Statistics

jakob-fritz:
#386 Works on this

PR #416, opened 2024-04-22
Include gitlab-sync in master branch

jakob-fritz:
Currently, development took place in fork and separate branch. The syncing to gitlab and using HPC for CI could now be included in master


brownbaerchen:
LGTM. A short summary of (what I think) is going on:

When somebody opens a PR, nothing new happens unless they attach the label `gitlab-mirror`. This will trigger the pipeline on the JSC gitlab, which runs on JUWELS. Code will only be actually run on JUWELS if the person who triggered the workflow has write access to this repository. The gitlab pipeline is always triggered on pushes to this repository.
For us it is then as easy as attaching the label whenever we want to run the tests. For others, the action will fail and someone from us has to retrigger the workflow, such that the required permissions are there. I don't think we can do without this degree of manual action.

The main thing that is still missing is integration of the coverage report from the two pipelines. The report is generated and transferred back to GitHub. But it is apparently more difficult to share reports between actions than between repositories. Still, as this actually runs the test, I think we gain a lot by merging this now and doing the report in a separate step.

@pancetta: Are you happy with JUWELS or should we rather use JUSUF? JUSUF is pretty empty, but the nodes are usually in an energy saving state which requires wait times of up to 5 mins for them to boot up properly. I find it difficult to gauge how much compute time these actions will take. So far, the actions complete rather fast.

brownbaerchen:
Wait a minute! There is now an extra workflow triggered which tests everything on push. Also, it tests with python 3.7, which @pancetta removed recently. Adding this back in is not marked in the diff here. Very confusing. I honestly don't know why any of this is happening, but we should fix it before merging!

jakob-fritz:
Works on #415

PR #424, opened 2024-04-29
Removed cupy-tests until they are fixed

jakob-fritz:
As I cannot push to master (which makes sense), this PR removes the cupy-test in gitlab, so that CI now succeeds again. Then failures can be detected easier.

Once, the cupy-tests are fixed, the updated code shall be merged in master and also the tests will be added again.

### DIFF
--- a/pySDC/tests/test_sweepers/test_MPI_sweeper.py
+++ b/pySDC/tests/test_sweepers/test_MPI_sweeper.py
@@ -152,7 +152,8 @@ def test_sweeper(num_nodes, quad_type, residual_type, imex, initGuess, launch=Tr
     individual_test(num_nodes, quad_type, residual_type, imex, initGuess, useNCCL=False, launch=launch)
 
 
-# @pytest.mark.cupy
+@pytest.mark.cupy
+@pytest.mark.skip(reason="We haven\'t figured out how to run tests on the cluster with multiple processes yet.")
 @pytest.mark.parametrize("num_nodes", [2])
 @pytest.mark.parametrize("quad_type", ['GAUSS', 'RADAU-RIGHT'])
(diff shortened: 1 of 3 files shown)


jakob-fritz:
Updated the CI-file, so that the gitlab-ci is also triggered, when a new label is added (e.g. the label to start Gitlab-CI)

Combined in this PR, as both work on CI in general and gitlab-CI in particular

brownbaerchen:
Actually, I was planning to port some new things to GPU and make use of the cupy tests today. I would rather we skip the single test that runs in parallel and run the other tests for now.

jakob-fritz:
@brownbaerchen we can also do that. I just readded the general tests and commented out the one test, that got stuck. So now it (hopefully) succeeds.

pySDC/tests/test_sweepers/test_MPI_sweeper.py
```diff
@@ -152,7 +152,7 @@ def test_sweeper(num_nodes, quad_type, residual_type, imex, initGuess, launch=Tr
     individual_test(num_nodes, quad_type, residual_type, imex, initGuess, useNCCL=False, launch=launch)
 
 
-@pytest.mark.cupy
+# @pytest.mark.cupy
```
brownbaerchen, line 155: I think the cleanest way to implement this is to add a [skip decorator]. 
```
@pytest.mark.cupy
@pytest.mark.skip(reason="We haven\'t figured out how to run tests on the cluster with multiple processes yet.")
...
```

jakob-fritz: Yes, correct. I just forgot it. 🤦‍♂️

brownbaerchen:
LGTM

PR #426, opened 2024-04-29
Split project tests

jakob-fritz:
Splitting tests from base-environment into separate jobs to increase parallelism and speed


jakob-fritz:
@brownbaerchen is it on purpose, that the tests for monodomain are marked with "monodomain" instead of e.g. "base"? See all test-files in directory. E.g. [this one](https://github.com/Parallel-in-Time/pySDC/blob/376ab368848940ecd1a50da42ba584b083dd90f1/pySDC/tests/test_projects/test_monodomain/test_monodomain_convergence.py#L99)

I was surprised and assume, that those tests won't run, because currently only tests for base or other keywords will run.

However, we can also "invert" the decision and I run all tests in projects, that are not marked with any of the keywords, that need another environment-file

jakob-fritz:
Ignore question above. I saw, that monodomain needs packages, that are not included in "base" environment. Therefore, I created a separate env and moved the monodomain tests, so that they don't run with the base-env (and fail)

brownbaerchen:
The libpressio and monodomain tests need compiled code. That's why they're treated separately already and are not part of the matrix. You can just leave them as they are.

.github/workflows/ci_pipeline.yml
```diff
@@ -65,7 +65,69 @@ jobs:
...
+        shell: bash -l {0}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Conda environment with Micromamba
+        uses: mamba-org/setup-micromamba@v1
+        with:
+          environment-file: "etc/environment-base.yml"
```
brownbaerchen, line 121: This runs all of the project tests in the base environment except those with mpi4py, fenics or petsc, yes? I think what @pancetta wanted is to allow for project specific environments.
What about using [`conda env update`](https://docs.conda.io/projects/conda/en/latest/commands/update.html) with `--file` option? This way we could maybe define a base environment and build project environments on top which only need to specify what is different to the base environment.

brownbaerchen (changes requested):

.github/workflows/ci_pipeline.yml
jakob-fritz: Added this now. It uses `micromamba install` because the `update` really only updates packages that are already installed.

So all projects now got an additional file `env_update.yml` where pytest is installed. Pytest also is included in the base-env, but the files cannot be empty. As an alterantive another package could also be used if preferred.

.github/workflows/ci_pipeline.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     strategy:
       matrix:
         python: ['3.8', '3.9', '3.10']
-        env: ['base', 'fenics', 'mpi4py', 'petsc']
+        env: ['fenics', 'mpi4py', 'petsc']
```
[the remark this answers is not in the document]
jakob-fritz: Added `base` environment back to matrix, so that all base-tests from outside projects are now also tested again

jakob-fritz:
@brownbaerchen do you know, why this test on [SDC_showdown](https://github.com/Parallel-in-Time/pySDC/actions/runs/8892081318/job/24415404605?pr=426) always fails?

I tries to convert a float into a matrix, but I don't see what i changed that triggered this error.

brownbaerchen:
Cool! Can you add a short description of how to add a new project in the [contribution guide](https://github.com/Parallel-in-Time/pySDC/blob/master/CONTRIBUTING.md).
The steps are:
 - Add a folder in pySDC/projects
 - Add a non-empty environment update file called `env_update.yml` in the project directory
 - Add project name to the GitHub pipeline
 - Add tests to a directory called `tests` inside the project folder

That's it, right? I am wondering if we can forgo the need by the user to add the project name to the pipeline. Can we just run the tests for all directories in pySDC/projects except ones we specifically exclude?

brownbaerchen:
> @brownbaerchen do you know, why this test on [SDC_showdown](https://github.com/Parallel-in-Time/pySDC/actions/runs/8892081318/job/24415404605?pr=426) always fails?
> 
> I tries to convert a float into a matrix, but I don't see what i changed that triggered this error.

Hm. I think @pancetta wrote this test and probably knows best. The petsc environment limits the version to `petsc4py<3.20`. Have you tried that?

jakob-fritz:
> The petsc environment limits the version to `petsc4py<3.20`. Have you tried that?

Thank you. That fixed it.

jakob-fritz:
> Cool! Can you add a short description of how to add a new project in the [contribution guide](https://github.com/Parallel-in-Time/pySDC/blob/master/CONTRIBUTING.md). The steps are:
> 
> * Add a folder in pySDC/projects
> * Add a non-empty environment update file called `env_update.yml` in the project directory
> * Add project name to the GitHub pipeline
> * Add tests to a directory called `tests` inside the project folder
> 
> That's it, right? I am wondering if we can forgo the need by the user to add the project name to the pipeline. Can we just run the tests for all directories in pySDC/projects except ones we specifically exclude?

That is right, with a one annotations. The environment-file must not only be non-empty, but also with the correct structure and in YAML. See any environment-file in the other projects.

I don't know a way how the list of projects could be done automatically in CI __before__ starting the jobs. Therefore, it is necessary to add the (new) own project to the list in the CI-file manually.

I read through the contribution guide and couldn't figure out, where these pieces of information fit best, i.e. where they are most likely to be found by the relevant people, when a new project is created. What is your view on the preferred location?

pancetta:
I think I don't understand why we need the (empty or dummy) `env_update.yml` file. This is unclear to people writing projects and seems redundant? If we require an env-file for each project anyway, why not the full one? Then it's at least understandable what needs to be in there.

brownbaerchen:
> I think I don't understand why we need the (empty or dummy) `env_update.yml` file. This is unclear to people writing projects and seems redundant? If we require an env-file for each project anyway, why not the full one? Then it's at least understandable what needs to be in there.

I guess the main reason we went with this is that there is a lot of pytest stuff that is annoying. How about doing both? We can specify all modules we need in the projects env files and use that to update the environment from the base environment in the pipeline such that we don’t need four obscure pytest installations in each project.

jakob-fritz:
> I think I don't understand why we need the (empty or dummy) `env_update.yml` file. This is unclear to people writing projects and seems redundant? If we require an env-file for each project anyway, why not the full one? Then it's at least understandable what needs to be in there.

Of course, we can also go with 'only' the environment-file per project. The reason why I chose this approach is to avoid (or at least reduce) duplication of depedencies. If we go for a 'complete' environment-file per project, many dependencies will duplicate leading to an increased risk of diverging over time, when only a few environment-files are updated. Therefore, the idea was to have a common basis and putting the requirements that are specific for a project into this second file.

In addition to what @brownbaerchen said, we can also put the requirements that are only needed for testing into a separate env-file. Then they are installed in CI, but not when using pySDC locally.

Getting rid of the base-environment-file completely will be difficult, as we need some file for tests of the code, that is not part of any project.

pancetta:
Why do you see the problem of divergence? Unless we don't pinpoint versions, there is no divergence, right? And if we had global requirements, these should go into the base file. And if a test fails to run, someone has to fix the environment (if that's the problem). Maybe I'm just not understanding the problem here?

brownbaerchen:
> > I think I don't understand why we need the (empty or dummy) `env_update.yml` file. This is unclear to people writing projects and seems redundant? If we require an env-file for each project anyway, why not the full one? Then it's at least understandable what needs to be in there.
> 
> Of course, we can also go with 'only' the environment-file per project. The reason why I chose this approach is to avoid (or at least reduce) duplication of depedencies. If we go for a 'complete' environment-file per project, many dependencies will duplicate leading to an increased risk of diverging over time, when only a few environment-files are updated. Therefore, the idea was to have a common basis and putting the requirements that are specific for a project into this second file.
> 
> In addition to what @brownbaerchen said, we can also put the requirements that are only needed for testing into a separate env-file. Then they are installed in CI, but not when using pySDC locally.
> 
> Getting rid of the base-environment-file completely will be difficult, as we need some file for tests of the code, that is not part of any project.

I think it actually makes sense to decouple the base environment from the project environments because it is totally fine if they diverge. The reason we test the projects in the first place is that we want results in papers to be reproducible. If this requires limiting the version number of some dependency then that is totally fine. If the rest of the code runs with the most up-to-date version of the modules, then we should offer that in the base environment.

Let's do an environment for testing only and remove testing stuff from all environments and add it in by updating the environments. Are you fine with that @pancetta?
Then we can put everything in one matrix job again by using pytest tags for each project.
Can we somehow let the pipeline test all tags that were added to pytest instead of the user having to add the tag to the pipeline manually?

jakob-fritz:
I splitted/decoupled the environments of the projects from the base-environment. Now there is one environment for each project, where only the needed dependencies are included. Because of this change, I also changed the name to `environment.yml`, which is also the default name for conda/mamba. Another environment-file now contains all the pytest-related dependencies, that are only needed for testing (so not for productive use).
The base-environment still exists and is used for the tests that are not part of any project.

Using the project-environments to simply update the base-environment is the way, the files were designed before my changes today. There, the base-env included all dependencies (incl. pytest-stuff) for the majority of the code and the env-files in the projects contained additional dependencies if needed. If no additional dependencies were needed, I included a dummy-dependency (I chose pytest to make clear it's no dependency for the actual pySDC-code), as the files needs to exist and follow a valid structure. But again, I changed this structure today to self-sustained environment-files for the projects. Otherwise, I can also revert those changes.

Combining the project-tests and the other tests into a single matrix might be difficult, as the paths follow different patterns. Therefore, conditionals need to be included in the job, when the matrix is combined. This is doable, but I wanted to do those steps after each other.

jakob-fritz:
There is still the issue, that the tests in parallelSDC (in the file [test_preconditioner_MPI](<https://github.com/jakob-fritz/pySDC/blob/split_project_tests/pySDC/projects/parallelSDC/tests/test_preconditioner_MPI.py>)) still take very long (over 1h) to complete. I don't know why, it takes that long. Do you, @brownbaerchen @pancetta have any idea how this can be fixed, or if these jobs should be omitted?

pancetta:
> There is still the issue, that the tests in parallelSDC (in the file [test_preconditioner_MPI](https://github.com/jakob-fritz/pySDC/blob/split_project_tests/pySDC/projects/parallelSDC/tests/test_preconditioner_MPI.py)) still take very long (over 1h) to complete. I don't know why, it takes that long. Do you, @brownbaerchen @pancetta have any idea how this can be fixed, or if these jobs should be omitted?

But they didn't take that long before, right?

.github/workflows/ci_pipeline.yml
```diff
@@ -42,8 +42,9 @@ jobs:
 
     strategy:
       matrix:
-        python: ['3.8', '3.9', '3.10']
         env: ['base', 'fenics', 'mpi4py', 'petsc']
+        # python: ['3.8', '3.9', '3.10']
+        python: ['3.10']
```
brownbaerchen, line 47: Don't forget to revert this once we're ready to merge.

.github/workflows/ci_pipeline.yml
```diff
@@ -82,6 +87,73 @@ jobs:
...
+          - parallelSDC_reloaded
+          - PinTSimE
+          - RDC
+          - Resilience
+          - SDC_showdown
+          - Second_orderSDC
+          - soft_failure
+          - TOMS
+        # python: ['3.8', '3.9', '3.10']
+        python: ['3.10']
```
brownbaerchen, line 114: @pancetta also wants all python versions tested here. Btw: Did you test 3.11 and 3.12? Especially with the non-project tests?

.github/workflows/ci_pipeline.yml
```diff
@@ -82,6 +87,73 @@ jobs:
...
+      - name: Install additional packages as needed
+        run: |
+          micromamba install -y --file etc/environment-tests.yml --freeze-installed
+
+      - name: Run pytest for CPU stuff
+        run: |
+          echo "print('Loading sitecustomize.py...')
+          import coverage
+          coverage.process_startup() " > sitecustomize.py
+          coverage run -m pytest --continue-on-collection-errors -v --durations=0 pySDC/projects/${{ matrix.env }}/tests
```
brownbaerchen, line 140: Did you check that the website looks good? The plots are put into a relative path, which is likely wrong when we move the tests. Sorry that I only remember this now. Maybe we better put the tests back to where they were. Then we can use the same matrix for everything. But the matrix we can do either way, I think, when we use the pytest markers and just test everything in `pySDC/tests` and `pySDC/tests/projects`.

.github/workflows/ci_pipeline.yml
```diff
@@ -108,7 +180,7 @@ jobs:
           source /opt/spack/share/spack/setup-env.sh
           spack load libpressio
            
-          coverage run -m pytest --continue-on-collection-errors -v --durations=0 pySDC/tests -m libpressio
+          coverage run -m pytest --continue-on-collection-errors -v --durations=0 pySDC/projects/compression/tests
```
brownbaerchen, line 183: Let's not remove the pytest marker

.github/workflows/ci_pipeline.yml
```diff
@@ -156,7 +228,7 @@ jobs:
           echo "print('Loading sitecustomize.py...')
           import coverage
           coverage.process_startup() " > sitecustomize.py
-          coverage run -m pytest --continue-on-collection-errors -v --durations=0 pySDC/tests -m monodomain
+          coverage run -m pytest --continue-on-collection-errors -v --durations=0 pySDC/projects/Monodomain/tests
```
brownbaerchen, line 231: Also here I'd like to keep the marker

.github/workflows/ci_pipeline.yml
```diff
@@ -230,7 +303,7 @@ jobs:
       - name: Install Conda environment with Micromamba
         uses: mamba-org/setup-micromamba@v1
         with:
-          environment-file: "etc/environment-base.yml"
+          environment-file: "etc/environment-tests.yml"
```
brownbaerchen, line 306: Are we sure this is enough? Can you try running this job locally and see if the website looks good?

docs/contrib/06_new_project.md
```diff
@@ -0,0 +1,40 @@
+# Adding a project to pySDC (and automatic testing)
+
+When adding a new project to pySDC, it needs to be tested.
+In order to run those tests as fast as possible, a few steps are needed, that are given below.
```
brownbaerchen, line 4: When you do a project with pySDC we appreciate it if you merge it back to the main repository. We are committed to keeping your work reproducible and prevent it from fading into oblivion.
To that end, please write extensive tests for your code and add them to the project. See the [contribution guide](../../../CONTRIBUTING.md) for general advice on testing etc. This guide will detail only how to add a project to pySDC.

docs/contrib/06_new_project.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Add a directory in 'pySDC/projects'
+
+First, create a new directory in `pySDC/projects` with the name of your project.
+The code of the new project should go into that newly created directory.
+
+## Add an environment-file
+
+In addition to the codde of your projects, an environment-file is needed.
```
brownbaerchen, line 13: The testing pipeline uses [micromamba](https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html) and requires an environment file for setup.

docs/contrib/06_new_project.md
```diff
@@ -0,0 +1,40 @@
...
+This includes the dependencies that are required for your project to run.
+The file needs to be named `env_update.yml` and needs to follow the structure shown below:
```
brownbaerchen, line 15: `env_update.yml` is outdated

docs/contrib/06_new_project.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+The list of dependencies can be extended as needed.
+The name should stay `pySDC`. The channels cover most of the usual packages.
+If a package is needed that cannot be found in those channels by conda (or mamba),
+please add the correct channel to the list.
+
+## Add tests to the project
+
+In order to automatically find the tests of your project, please add the tests to a subdirectory called `tests` in the directory of your project. Furthermore, the files should start with `test_` and the functions should also start with that. For more information see the documentation of pytest on [test discovery](<https://docs.pytest.org/en/8.2.x/explanation/goodpractices.html#tests-as-part-of-application-code>).
```
brownbaerchen, line 33: We need to remember to put whatever we settle on in the end here.

docs/contrib/06_new_project.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Add the project in the CI-file
```
brownbaerchen, line 35: ## Add the project to the continuous integration pipeline

docs/contrib/06_new_project.md
```diff
@@ -0,0 +1,40 @@
...
+
+To run the tests of all projects in parallel, the projects are explicitly mentioned in the CI-file.
+In order to run the tests of your project, please add the name of your project **as your directory is named**
+in the [CI-File](./../../.github/workflows/ci_pipeline.yml)
+in the job `project_cpu_tests_linux` in the list `strategy/matrix/env`.
```
brownbaerchen, line 40: Again, we need to make sure that this is correct in the end. If we decide to use the. pytest markers (I vote we do) we need to detail here how to add you own.

brownbaerchen, line 40: @pancetta, do you want to add a few lines on how to obtain a release of pySDC to go along with a publication?

etc/environment-mpi4py.yml
```diff
@@ -10,6 +10,3 @@ dependencies:
   - mpich
   - mpi4py-fft>=2.0.2
   - mpi4py>=3.0.0
-  - pytest
-  - pytest-cov
-  - pytest-timeout
```
brownbaerchen, line 15: @pancetta why was `pytest-order` missing here? Are we even using that?

etc/environment-tests.yml
```diff
@@ -0,0 +1,11 @@
...
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - pytest
+  - pytest-benchmark
+  - pytest-timeout
+  - pytest-order
+  - coverage[toml]
+  - sphinx
```
brownbaerchen, line 11: Does it make sene to do a special environment for post processing such we don't need to install `sphinx` in every environment? I don't know how long this takes. But if we can shave a few seconds off every single pipeline, that might be worth doing.

pySDC/projects/parallelSDC/environment.yml
```diff
@@ -0,0 +1,10 @@
+name: pySDC
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - mpi4py
+  - numpy
+  - matplotlib>=3.0
+  - dill>=0.2.6
+  - scipy>=0.17.1
```
brownbaerchen, line 10: Let's put exactly the base environment here and see if the test times change.

brownbaerchen (changes requested):

docs/contrib/06_new_project.md
pancetta: Something like:
```
If you project is published and you need a dedicated pySDC version with a DOI, please get in touch with us and/or open a new issue. We will help you with this as soon as possible. Note that a final DOI is usually only necessary once a paper is accepted and the final proofs are due. We strongly encourage to describe and cite the current version of pySDC already during initial submission, though.
```

etc/environment-mpi4py.yml
pancetta: I.... don't know.

.github/workflows/ci_pipeline.yml
brownbaerchen: I forgot what the issue was. That's why I wanted to test everything with everything and see if the base stuff at least works with the latest python. That would be nice because I would like to transition to a newer version on my laptop. On the cluster I have been using 3.11 with no issues. Actually, the HPC tests are probably all run with 3.11, maybe..?

etc/environment-mpi4py.yml
jakob-fritz: We use it only in one project (TOMS). As the tests there are marked with "base", there is no need for pytest-order there. For simplicity I included it in environment-tests anyway, as one job needs it.

.github/workflows/ci_pipeline.yml
jakob-fritz: No, I did not test more python versions. This reduction on python-versions is also only for current development, as the CI is faster with fewer jobs and finds most of the errors also with only a single python-version.

It is not meant to stay like that for merging.

.github/workflows/ci_pipeline.yml
jakob-fritz: We can also use the markers. I chose this change to make sure to include all tests there; also ones, where the marker may be missing. But We can also include the requirement for a marker there, if you prefer that.

.github/workflows/ci_pipeline.yml
jakob-fritz: Same as above. I chose to remove the marker, as the path limits the tests already thematically. Therefore, the marker was an additional potential point of failure (skipping tests instead of running them). But also here: We/I can also include that marker again.

docs/contrib/06_new_project.md
jakob-fritz: Changes will be included in the next commit.

docs/contrib/06_new_project.md
jakob-fritz: Will also be included in the next commit

docs/contrib/06_new_project.md
jakob-fritz: Name of the file is updated in the next commit.

docs/contrib/06_new_project.md
jakob-fritz: Headline updated in the next commit

pySDC/projects/parallelSDC/environment.yml
jakob-fritz: I just gave it a try with an environment that is similar (not the same) to the mpi4py-environment. The local results looked promising. Let's see how CI will behave.

.github/workflows/ci_pipeline.yml
brownbaerchen: Yes please. The marker marks tests that should be run in the docker container. Let's use it like it's intended.

etc/environment-mpi4py.yml
pancetta: Not sure why I did this [here](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/tests/test_projects/test_TOMS/test_AllenCahn_contracting_circle.py), but I guess by merging the two tests we could get rid of it.

brownbaerchen: Actually, what you did there is quite nice! I suggest we keep this module and use it more often for plotting for the website. Personally, I really like splitting the tests as much as possible so you get fine-grained feedback, but then you cannot plot all of the results in a single plot without this module.

pancetta: Fine by me!

pySDC/projects/parallelSDC/environment.yml
jakob-fritz: As the jobs for parallelSDC need only 10-20 minutes (the first one and the reloaded one respectively), I assume, we can close this conversation. If you think otherwise, please reopen it or open a new one.

docs/contrib/06_new_project.md
brownbaerchen: I like the markers because they assign the tests to the respective environments in a very explicit way. But I can see that we disagree on this issue. @pancetta, please note your preference and we'll go with that.

.github/workflows/ci_pipeline.yml
jakob-fritz: Reverted in eebbdd3b89a71f3c3063efccfab926abde748fea

.github/workflows/ci_pipeline.yml
jakob-fritz: Reverted in eebbdd3b89a71f3c3063efccfab926abde748fea and also added more python-versions for testing (3.11 & 3.12)

.github/workflows/ci_pipeline.yml
jakob-fritz: Readded the marker in b858c33c7fdec93b049845f5232f5b817c127c27

docs/contrib/06_new_project.md
pancetta: The markers were used originally because we wanted to filter the tests by a certain property. Adding new ones is pretty clumsy and forgetting to set a marker was also easily possible. So, if we don't need markers, we should get rid of them. But if we need to do other filter magic instead, we should keep them. So, what would happen if we drop markers?

.github/workflows/ci_pipeline.yml
jakob-fritz: Readded the marker in b858c33c7fdec93b049845f5232f5b817c127c27

docs/contrib/06_new_project.md
pancetta: In any case, we should not mix markers and non-markers. If we decide to use markers, we should only use markers. If we decide to get rid of markers, we should get fully rid of them.

jakob-fritz: Added the text proposed by @pancetta in commit 1a4b58f591c1e0a0f1fe1e22922c3f16a81f046c and also added a headline as it is not directly related to the CI-Pipeline

brownbaerchen: Jakob currently set things up so that the project tests are in the project folders and the pipeline just runs all tests in a project folder with the environment file in the same folder. The remaining tests are still in the separate tests directory and use the markers `base`, `mpi4py`, ... 
So then you don't need to add markers when adding a new project. Fine. I agree that it's a bit simpler ;)

docs/contrib/06_new_project.md
```diff
@@ -0,0 +1,60 @@
+# Adding a project to pySDC (and automatic testing)
+
+When you do a project with pySDC we appreciate it if you merge it back to the main repository.
+We are committed to keeping your work reproducible and prevent it from fading into oblivion.
+To that end, please write extensive tests for your code and add them to the project.
+See the [contribution guide](./../../CONTRIBUTING.md) for general advice on testing etc.
+This guide will detail only how to add a project to pySDC.
+
+When adding a new project to pySDC, it needs to be tested.
+In order to run those tests as fast as possible, a few steps are needed, that are given below.
```
brownbaerchen, line 10: I think we don't need the above two lines anymore. My goal was to elaborate on these points with the above text.

jakob-fritz: Fine. Will be removed in next commit

jakob-fritz:
One job fails for python 3.12. The issue arises for higher number of Threads (5 and 8) but does not arise for older versions of python <=3.11 (or lower numbers of threads).
See log [here](<https://github.com/Parallel-in-Time/pySDC/actions/runs/8936665061/job/24547469281?pr=426>)
This is not a one-time-issue, but occured at least twice for two pushes.

The issue seems to come from using a different MPI. When comparing the installation logs, in Python 3.11 impi (sometimes listed as impi_rt) is used. In Python 3.12 OpenMPI is installed instead. Maybe these two packages have different default options (maybe for oversubscription or the likes?)

Do you @brownbaerchen want to have a look there, or shall I dig in, but that will take more time, as I haven't worked with the pySDC code itself until now.

pancetta:
Looks like a oversubscription issue which is indeed fixed by either specifying a parameter during `mpirun` or by using another MPI version (I think MPICH does not have this problem).

jakob-fritz:
> Looks like a oversubscription issue which is indeed fixed by either specifying a parameter during `mpirun` or by using another MPI version (I think MPICH does not have this problem).

Yes, adding mpich in the requirements solved it. Thank you @pancetta for the pointer!

brownbaerchen:
Looking good! Testing the past 5 versions is perhaps a bit excessive. Which ones do you want to keep @pancetta?

pancetta:
Looking [here](https://devguide.python.org/versions/), v3.8 to v3.12 would be great to have. I'd be fine dropping the oldest one, though.

pancetta:
So, runtime of the tests went up to 1h.. is that intended?

jakob-fritz:
No it is not. This is because of matplotlib. When I limit the version, the duration goes down to 10-20 minutes. But fails for Python 3.11 and Python 3.12., because those older versions of matplotlib are not compatible with the later python versions.

I am currently at it and try to find where this issue with matplotlib comes from

pancetta:
Ahhh, yes, I recall that I did not pursue 3.12 because Matplotlib was causing trouble... right.

brownbaerchen:
Hm. That's annoying. What about testing projects only up to 3.10 and the rest with everything? Since only the projects do plots (right? Tutorials..?) we don't need to worry about matplotlib for the rest.

jakob-fritz:
I added a statement to skip the combination of parallelSDC and Python 3.11 or Python 3.12. So all other projects are tested with 5 versions of python and the project parallelSDC is tested with "only" 3 versions of python. By this, we can cover as much as possible for the other projects, while keeping the version-limits for matplotlib

pySDC/projects/parallelSDC/tests/test_preconditioner_MPI.py
brownbaerchen, line 1: Why is this file marked as new instead of moved in the diff? Did you check in the "removed" version in the original directory? Probably some GitHub bug...

brownbaerchen:
By chance, the relative paths for the plots still work out. So the website and coverage report look like before. Time is down to ~30 mins from ~45. More Python versions including the latest 3.12 are tested.

To get time down further, we would have to rewrite some tests, mainly in the DAE, resilience and the parallelSDC_reloaded projects. But as they generate some plots, this may be difficult and definitely shouldn't be part of this PR.

I am happy now. Any more requests, @pancetta?

brownbaerchen (approved):

docs/contrib/06_new_project.md
jakob-fritz: Just to avoid confusion: We can get rid of all markers within the tests for projects. But currently we still use pytest-markers for all tests outside of projects. I wanted to explicitly state that as it slightly contradicts the statement of @pancetta about using markers everywhere or nowhere.
Is that fine with you, that we use them nowhere within the project-tests and everywhere outside of project-tests, @pancetta ?
I don't see a good way to get rid of the markers for the tests outside of projects, but don't think we need them within project-tests.

pancetta:
Ok, let me try to understand this: each project now has a separate environment file, yes? And then there are environment files for the tutorials (and other stuff?) in the `env` folder? Does it make sense to keep those separate? Why not do this consistently?

docs/contrib/06_new_project.md
pancetta: Yeah, we did introduce inconsistencies here, too (see also my other comment below).

jakob-fritz:
Yes, you got this right, @pancetta. The environment-file for each project is located in the directory of that project. Furthermore, there are more general environment-files (e.g. for tutorial, the tests outside projects and with testing-dependencies in general) in the `etc` directory. We could move the env-files of the projects also there, but then the projects are less self-contained. If we keep the environment-files in the project-dirs, the environment-files are in two places (conceptually; of course, many more, since the files are in the project-dirs).

pancetta:
Alright, there is probably no single great solution to this. I really appreciate that the projects are now more self-contained and that the tests are faster/more modular. Any famous last words here, @brownbaerchen @jakob-fritz ?

jakob-fritz:
> Alright, there is probably no single great solution to this. I really appreciate that the projects are now more self-contained and that the tests are faster/more modular. Any famous last words here, @brownbaerchen @jakob-fritz ?

Not last words: Please don't merge yet. I am adding dependencies for the documentation (did not raise errors, but only warnings in the logs, I saw this morning).

jakob-fritz:
Now, the documentation has all needed dependencies installed.
I just pushed a minor change, where I removed print-statements in the CI, that I added earlier today for debugging.

Once this is done, I am fine with squash and merge

.github/workflows/ci_pipeline.yml
jakob-fritz: Now the environment for documentation has all needed dependencies

docs/contrib/06_new_project.md
jakob-fritz: Should work this way, as we don't need pytest-markers and only put the files in the correct spot.

PR #427, opened 2024-04-30
Implemented MPI-parallel multilevel diagonal SDC

brownbaerchen:
Turns out the base transfer class is not compatible with the diagonal sweepers. This is required for PFASSTER.
I don't really know the underlying mathematics, so I just test that the result is the same between the new `base_transfer_MPI` and `base_transfer` classes. Let me know if you want anything else tested. There is also a test that a single sweep of multilevel SDC gives the same result with diagonal sweeper and regular sweeper.

I tried `pytest-easyMPI` and it didn't work for me. Instead, I use `subprocess` in combination with `argparse`. I didn't use `argparse` so far, but will use it a lot from now on. I think it's a very useful library generally, including tests with MPI. Have a look if you're interested in using command line arguments.


tlunet:
Yep, `argparse` is really nice indeed (been using it quite a lot already ...).

However, I'm not sure to fully understand why the base transfer class is not compatible with diagonal sweepers ... this should be totally decoupled normally (at least I thought so)

pySDC/core/BaseTransfer.py
```diff
@@ -45,7 +45,6 @@ def __init__(self, fine_level, coarse_level, base_transfer_params, space_transfe
         # set up logger
         self.logger = logging.getLogger('transfer')
 
-        # just copy by object
```
tlunet, line 48: It's not a copy, but rather the assignment of `self.fine` to the variable referenced by `fine_level`. Python never use copy except if you tell him that explicitly, or create new objects before assigning

pySDC/tests/test_sweepers/test_MPI_sweeper.py
```diff
@@ -158,8 +176,8 @@ def test_sweeper(num_nodes, quad_type, residual_type, imex, initGuess, launch=Tr
 @pytest.mark.parametrize("quad_type", ['GAUSS', 'RADAU-RIGHT'])
 @pytest.mark.parametrize("residual_type", ['last_abs', 'full_rel'])
 @pytest.mark.parametrize("imex", [False])
-@pytest.mark.parametrize("initGuess", ['spread', 'copy', 'zero'])
-def test_sweeper_NCCL(num_nodes, quad_type, residual_type, imex, initGuess, launch=True):
+@pytest.mark.parametrize("init_guess", ['spread', 'copy', 'zero'])
```
tlunet, line 179: sensing some resistance about camelCase here ... :innocent:

pySDC/tests/test_sweepers/test_MPI_sweeper.py
```diff
@@ -171,15 +189,36 @@ def test_sweeper_NCCL(num_nodes, quad_type, residual_type, imex, initGuess, laun
...
+        type=str,
+        help='Way of computing the residual',
+        choices=['full_rel', 'last_abs', 'full_abs', 'last_rel'],
+    )
+    parser.add_argument('--imex', type=str_to_bool, help='Toggle for IMEX', choices=[True, False])
+    parser.add_argument('--useNCCL', type=str_to_bool, help='Toggle for NCCL communicator', choices=[True, False])
+    parser.add_argument('--init_guess', type=str, help='Initial guess', choices=['spread', 'copy', 'zero'])
+    args = parser.parse_args()
+
+    kwargs = {me[0]: me[1] for me in args._get_kwargs()}
```
tlunet, line 223: Simpler like this :

```python
kwargs = vars(args)
```
which you even more simplify in the following line :

```python
individual_test(**vars(args))
```

tlunet: Hope you didn't spend too much time finding this other solution :laughing: 

image

pancetta:
> Yep, `argparse` is really nice indeed (been using it quite a lot already ...).
> 
> However, I'm not sure to fully understand why the base transfer class is not compatible with diagonal sweepers ... this should be totally decoupled normally (at least I thought so)

The transfer classes need to build the tau correction, which in turn needs to calculate `QF`. And that is distributed in case of diagonal/parallel sweepers. Yes, this is a design flaw, but hey, 10 years ago..

pySDC/tests/test_sweepers/test_MPI_sweeper.py
brownbaerchen: It just looked very inconsistent 🙈

pySDC/tests/test_sweepers/test_MPI_sweeper.py
brownbaerchen: Thanks, that's indeed nicer! I was just going from the argparse docs which didn't take too long. Maybe I should use chatGPT more...

brownbaerchen:
Bug was in the tau correction which was not computed with only two levels. Test for MPI sweeper now goes up to 3 levels.

pancetta:
What is the status here?

brownbaerchen:
Either somebody review it again or merge it, I guess :D

pancetta:
I kept thinking "Didn't I do this at some point??" and yes, I did: https://github.com/Parallel-in-Time/pySDC/blob/a5498546005050b284b77a4f6bac5b1396f4e181/pySDC/projects/parallelSDC/BaseTransfer_MPI.py#L20

brownbaerchen:
face palm... Looks rather similar :D But are you sure your [`prolong f`](https://github.com/Parallel-in-Time/pySDC/blob/a5498546005050b284b77a4f6bac5b1396f4e181/pySDC/projects/parallelSDC/BaseTransfer_MPI.py#L194) works? There's no communication at all in there...
How do you want to proceed? We should have this feature not hidden in a project in the end, but we can use your version if you want.

pancetta:
As you know, I'm a very good programmer, so, sure, it works 😉 ... that said, can you please remove my code and replace it with yours, making it indeed accessible for "everyone"?

brownbaerchen:
Using the lovely diff feature of vim, I can see there are some differences after all. Why do you do stuff to the initial conditions in prolong? Eg [here](https://github.com/Parallel-in-Time/pySDC/blob/a5498546005050b284b77a4f6bac5b1396f4e181/pySDC/projects/parallelSDC/BaseTransfer_MPI.py#L183). I cannot find something similar in the non MPI version. So is this needed or not? We should have it consistent anyways.

tlunet:
> Using the lovely diff feature of vim, I can see there are some differences after all. Why do you do stuff to the initial conditions in prolong? Eg [here](https://github.com/Parallel-in-Time/pySDC/blob/a5498546005050b284b77a4f6bac5b1396f4e181/pySDC/projects/parallelSDC/BaseTransfer_MPI.py#L183). I cannot find something similar in the non MPI version. So is this needed or not? We should have it consistent anyways.

"it might have changed in PFASST" ... wait, what does the `might` mean ? :scream:

brownbaerchen:
@pancetta, what the hell where you going from when you implemented this MPI version? The stuff in question was removed by you in the non MPI version in 2018, see [here](https://github.com/Parallel-in-Time/pySDC/commit/2bcb321b2b26eec46e9c0d2d5e07d392475a0fb7) :D

pancetta:
Relax. "might" means that these things are also called during MLSDC sweeps, but there the `u0` is not changed. PFASST however changes the `u0` during its iterations and that's why we need to take care of it here. It was not removed by the commit you mention, just done differently.

brownbaerchen:
> Relax. "might" means that these things are also called during MLSDC sweeps, but there the `u0` is not changed. PFASST however changes the `u0` during its iterations and that's why we need to take care of it here. It was not removed by the commit you mention, just done differently.

Ah, sorry! Was it removed [here](https://github.com/Parallel-in-Time/pySDC/commit/b3f63f49771f3072e418535ca9d3d76fa2e2776b)?

pancetta:
Hm. Maybe so. In any case, it is currently not there. Tbh, this is all a bit blurry in my head, long time ago. I would assume the base transfer class works as expected and we should go from there.

brownbaerchen:
I don't really know what's going on. I just took the current version and MPI-ied it. The tests check that MPI and non-MPI versions produce the same results. Let me know if more tests are needed. I don't know if I cover enough cases.

pancetta:
Well, tests pass, must be correct, no? 😉

PR #430, opened 2024-05-08
Housekeeping

jakob-fritz:
Some changes with reformatting and housekeeping


.github/workflows/ci_pipeline.yml
```diff
@@ -110,47 +97,41 @@ jobs:
           - Second_orderSDC
           - soft_failure
           - TOMS
-        python: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        # python: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python: ['3.10']
```
brownbaerchen, line 101: Same as above

.github/workflows/ci_pipeline.yml
```diff
@@ -110,47 +97,41 @@ jobs:
...
         exclude:
           - env: parallelSDC
             python: '3.11'
           - env: parallelSDC
             python: '3.12'
-
```
brownbaerchen, line 119: I should have asked this in the last PR. But pytest allows for conditional skips. I believe we could add to specific tests to be skipped for specific python versions. This would clean up the pipeline a little bit. But it has its own disadvantages. What do you think? Since I am on my phone I cannot look up if this really works.

etc/environment-postprocess.yml
```diff
@@ -1,3 +1,5 @@
+---
```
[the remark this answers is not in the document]
jakob-fritz: The reason I changed it is that it is now according to [specification](https://yaml.org/spec/1.2.2/#22-structures) and the linter found it.

It is often done otherwise and it seems odd.

As there is only a single document per file we could also leave them away. That is not strictly compliant with spec, but there is no reason for an error, as there is only a single document per file. This is way more important if you use a stream instead of a file.

.github/workflows/ci_pipeline.yml
jakob-fritz: We can also move the skipping of some jobs to the job-level instead of the  workflow level. I didn't know about this option. The downside is that it is not directly obvious that some tests are skipped, but (as an upside) more code is tested with more python-version.

.github/workflows/ci_pipeline.yml
```diff
@@ -10,76 +12,63 @@ jobs:
...
     runs-on: ubuntu-latest
-
     strategy:
       fail-fast: false
       matrix:
         env: ['base', 'fenics', 'mpi4py', 'petsc', 'pytorch']
-        python: ['3.8', '3.9', '3.10', '3.11', '3.12']
-
+        # python: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python: ['3.10']
```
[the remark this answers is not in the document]
jakob-fritz: Done in b460c3bfbe4709e718938daa18ea4d28386c3158

.github/workflows/ci_pipeline.yml
jakob-fritz: Done in b460c3bfbe4709e718938daa18ea4d28386c3158

etc/environment-postprocess.yml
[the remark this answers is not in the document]
jakob-fritz: Shall I remove them again to improve readability (to many people)? Or shall we leave as is now?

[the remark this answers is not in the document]
brownbaerchen: No, better leave it if this is the standard somehow.

lisawim:
We have test coverage reports again! Yippie-ki-yay! Thanks a lot for this good job!! :)

.github/workflows/ci_pipeline.yml
```diff
@@ -395,13 +356,13 @@ jobs:
 #
 #    steps:
 #      - name: Checkout
-#        uses: actions/checkout@v3
```
brownbaerchen, line 398: Do we need this entire job anymore? I guess this is made redundant by the actually working implementation of this. Should we delete the comments?

brownbaerchen (commented):
Apart from the minor thing, LGTM. Thanks a lot! Any objections, @pancetta?

.github/workflows/ci_pipeline.yml
jakob-fritz: From my point of view, I can also remove that job and the other one, that is commented out. I just didn't want to do that myself, as I was not sure if you (whoever) wanted to reactivate them in the future. But I can also happily remove them if that is fine with everyone.

jakob-fritz: Removed jobs in 48d7d4bbe34c5375ddfb981ee1b711c24a91f54c

jakob-fritz:
@pancetta shall this be merged, or do you have any objections or change requests?

brownbaerchen:
Merge away!

PR #431, opened 2024-05-10
Modification of ``DAE`` project

lisawim:
Since more files were added to the ``DAE`` project ``README.rst`` got an update by adding the new files to the list there. Further, I added a section how to implement a DAE problem in ``pySDC``.


brownbaerchen:
Are you sure the implementation details are set in stone at this point? You can consider adding the tutorial as documentation to the actual code of the class you implement here and then add the code to the readme. This is done in the other tutorials and makes sure the tutorials stay up to date when you change the code. I do appreciate your dedication to documentation!

lisawim:
Thanks a lot for your idea! I did catch what you suggested but I'm not quite sure yet how to realise that.

brownbaerchen:
Don't worry, it's not difficult. I did it recently in #417. You can use `.. literalinclude::` to include a file of code or any txt file or whatever into the Readme.

lisawim:
In order to follow the suggestion of @brownbaerchen I moved problem classes from simple_DAE.py into separate files so that each file only contains one problem class which is cleaner. First I wanted to use ``.. literalinclude::`` for only specific lines but this is quite non-flexible, since lines could change and then for the ``README`` they need to be changed as well (which could be forgotten). Creating files also requires adaption of the files in ``run`` and the test files. Further, I (tried to add) a ``.. literalinclude::`` in the README but it seems to not work yet. I guess I'm not sure which path I need to add. I thought the current one is correct, but this seems to be not the case.. Help is appreciated!

lisawim:
This probably works now..

docs/source/projects/doc_DAE_problematicF.rst
brownbaerchen, line 1: Why do you need these files? What is different from directly including the python files in the README?

pancetta: Once upon a time, there was a (good) reason for that. So good that I don't recall what it was.. It have have been a problem with including plots, but hey.

brownbaerchen: Please try without the extra files and see if you still need them. I got away without in #417 when doing really similar things.

lisawim:
Not from my side.

brownbaerchen:
Nope, let's go! :)

PR #432, opened 2024-05-13
Runge-Kutta sweeper for DAEs

lisawim:
This PR contains the Runge-Kutta sweeper base class for solving DAEs. Runge-Kutta methods for general DAEs $$0 = F(u, u', t)$$ read $$0 = F(u_0 + \Delta t \sum_{j=1}^m a_{i,j} U_j, U_m),$$ $$u_M = u_0 + \Delta t \sum_{j=1}^M b_j U_j$$ where a (non)-linear system needs to be solved at each node (in the same way as it needs to be done in the SDC-DAE sweepers). In literature, there are also *half-explicit* Runge-Kutta methods especially for semi-explicit DAEs, which is left for future work (i.e., I'll hand in the stuff some time).

The new Runge-Kutta sweeper for DAEs requires a ``du_exact`` method from the problem returning initial conditions for the derivative of the solution since initial conditions $u(t_0)$ and $u'(t_0)$ are needed. The base class thus has an attribute ``du_init`` where the value of the derivative at the end point is stored and passed to the new step. I don't know if I've chosen a smart solution for that, for the first time it works however. 

Butcher tableaux can be easily implemented in ``pySDC.implementations.sweeper_classes.Runge_Kutta.py``, where I added two new methods. The tests in ``test_Runge_Kutta_sweeper.py`` are adapted to include the two new methods in the tests. Of course, I also added tests for the Runge-Kutta-DAE sweeper for different cases of DAEs where the orders are tested. Note that for DAEs of higher index $k > 1$ order reduction is observed in the algebraic variable.

I'd really appreciate your review. Please let me know if there is something that I can improve.


pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
```diff
@@ -0,0 +1,181 @@
...
+    >>>     pass
+
+    More details can be found [here](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/implementations/sweeper_classes/Runge_Kutta.py).
+    """
+
+    def __init__(self, params):
+        super().__init__(params)
+        self.du_init = None
+
+    def predict(self):
```
brownbaerchen, line 82: You should raise `NotImplementedError` when someone tries to predict with something other than zero here.

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        Predictor to fill values at nodes before first sweep.
+        """
+
+        # get current level and problem
+        lvl = self.level
+        prob = lvl.prob
+
+        du_init = self.du_init[:] if self.du_init is not None else prob.du_exact(lvl.time)[:]
+        lvl.f[0] = prob.dtype_f(du_init)
```
brownbaerchen, line 92: I am a bit irked by the exact solution appearing in the sweeper. But I don't know how to fix it because I am still confused by DAEs. 

We should definitely make sure that the exact solution in the problem class is only called in the first time step. And I would like to see this explicitly in this function. Can we do that somehow? What if we put this in the `__init__` function of this sweeper? Is the problem class properly initialised at that time? 

Or maybe we should make this a parameter of the sweeper. Because you are not limited to the `u_exact` function of the problem for choosing initial conditions. And if you put in something else, this will be totally off. But then the problem is not initialised at the time that you pass the sweeper parameters to the controller during instantiation.

In short: I am not happy, but I don't know a good solution and I am willing to compromise.

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
```diff
@@ -0,0 +1,181 @@
...
+                ----------
+                du : dtype_u
+                    Unknowns of the system (derivative of solution u).
+
+                Returns
+                -------
+                sys : dtype_f
+                    System to be solved.
+                """
+                du_mesh = prob.dtype_f(du)
```
brownbaerchen, line 143: Why is this called `du_mesh` and not `local_du_approx` or something more descriptive?

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
```diff
@@ -0,0 +1,181 @@
...
+
+                local_u_approx = prob.dtype_u(u_approx)
+
+                # defines the "implicit" factor, note that for explicit RK the diagonal element is zero
```
brownbaerchen, line 147: This comment seems out of place. It is not clear what is defining the "implicit" factor.

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
```diff
@@ -0,0 +1,181 @@
...
+        assert lvl.status.unlocked
+        assert lvl.status.sweep <= 1, "RK schemes are direct solvers. Please perform only 1 iteration!"
+
+        M = self.coll.num_nodes
+        for m in range(M):
+            u_approx = prob.dtype_u(lvl.u[0])
+            for j in range(1, m + 1):
+                u_approx += lvl.dt * self.QI[m + 1, j] * lvl.f[j][:]
+
+            def F(du):
```
brownbaerchen, line 119: Is this function shared with the DAE sweeper for SDC? If yes, I suggest having it only once. You can make it a static method of the SDC DAE sweeper and import it here for instance.

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
```diff
@@ -0,0 +1,181 @@
...
+
+                local_u_approx = prob.dtype_u(u_approx)
+
+                # defines the "implicit" factor, note that for explicit RK the diagonal element is zero
+                local_u_approx += lvl.dt * self.QI[m + 1, m + 1] * du_mesh
+
+                sys = prob.eval_f(local_u_approx, du_mesh, lvl.time + lvl.dt * self.coll.nodes[m + 1])
+                return sys
+
+            finit = prob.dtype_f(lvl.f[m])
```
brownbaerchen, line 153: Why do you need to copy here? Can you not simply pass `lvl.f[m].flatten()` in the next line?

pySDC/projects/DAE/tests/test_RungeKuttaDAE.py
```diff
@@ -0,0 +1,305 @@
...
+            np.log(errorsDiff[i] / errorsDiff[i - 1]) / np.log(dt_list[i] / dt_list[i - 1])
+            for i in range(1, len(dt_list))
+        ]
+    )
+    orderAlg = np.mean(
+        [np.log(errorsAlg[i] / errorsAlg[i - 1]) / np.log(dt_list[i] / dt_list[i - 1]) for i in range(1, len(dt_list))]
+    )
+
+    assert np.isclose(
+        orderDiff, expectedOrderDiff[get_sweeper(sweeper_name).__name__], atol=1e0
```
brownbaerchen, line 155: `sweeper_name` should suffice here

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
lisawim: Yes, I agree!

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
lisawim: Having this function only ones avoids some lines. DRY principle I heard from a wise person sometime.. I had an idea setting then
```
def F(u, local_u_approx):
    # do something
``` 
because this function requires the local ``u_approx`` from ``update_nodes``. I also would then make ``u_approx`` make as an argument for the ``solve_system`` method (similar to ``rhs`` in the other sweepers in ``pySDC``). Realising this requires adaption on the other DAE-sweepers and in the parent problem class. Does it probably make sense to do this in another PR, since I know you like clean PR's?

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
lisawim: I'm also not really happy with that. But I haven't found a better solution for that.. Your ideas are quite good. I'd like to realise your first idea. However, as far as I know the level is not yet instantiated at a level when the sweeper is initialised, right? Therefore, calling ``self.du_exact(lvl.time)`` in ``__init__`` would not work?

brownbaerchen: Yes, you are right. How about adding sth like `self.fully_initialized = False` to `__init__` of sweeper. Then you can do in `predict`:
```
if self.fully_initialized:
    du_init = self.du_init[:]
else:
    du_init = prob.du_exact(lvl.time)[:]
    self.fully_initialized = True
```
or sth similar.

Still not sure if this is the best solution, though.

lisawim: I implemented such a statement where we initialize only in the first step (in the first call of this if-else-statement) with the exact solution of ``du``. I still haven't a better solution as well, but for the first time this would be okay and we can move brainstorming to anytime else.

pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
```diff
@@ -78,17 +79,23 @@ class RungeKuttaDAE(RungeKutta):
...
+        Predictor to fill values with zeros at nodes before first sweep.
         """
 
         # get current level and problem
         lvl = self.level
         prob = lvl.prob
 
-        du_init = self.du_init[:] if self.du_init is not None else prob.du_exact(lvl.time)[:]
+        if self.fully_initialized:
+            du_init = self.du_init[:]
```
brownbaerchen, line 94: Do we need the deep copy by using `[:]` here? We don't we just do
```
if not self fully_initialized:
    self.du_init = prob.du_exact(lvl.time)
    self.fully_initialized = True

lvl.f[0] = prob.dtype_f(self.du_init)
...
```

Issue #434, opened 2024-05-22
Inconsistent Allen-Cahn problems


brownbaerchen:
There are many versions of Allen-Cahn implemented in pySDC. Mostly, they solve the same equation, but differ only in dimensionality or the solver. However, there are some subtle differences that can lead to bugs. Comparing for instance the [2D FD](https://github.com/Parallel-in-Time/pySDC/blob/d89cac11ddef08c1c3010ccdd4f059e20c452f72/pySDC/implementations/problem_classes/AllenCahn_2D_FD.py) and [2D MPIFFT](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/implementations/problem_classes/AllenCahn_MPIFFT.py) implementations, we find the FD version has a grid [-0.5, 0.5]^2 and the MPIFFT version has a grid [0, 1]^2. Also, the FD version has phases $\pm 0.5$, with the other one having phases $0$ and $1$. I noticed this when trying to use a [hook for computing the radius](https://github.com/Parallel-in-Time/pySDC/blob/d89cac11ddef08c1c3010ccdd4f059e20c452f72/pySDC/projects/TOMS/AllenCahn_monitor.py) designed for the FD version on the FFT one.

Are there any reasons for these discrepancies? Do they matter in practice? If not, I suggest we streamline this at some point.

Also, we can unify the GPU and CPU versions of the FD implementation and change to the work counters there in the process.

pancetta:
So, there is no (good) reason for this discrepancy. We started somewhere and changed the way we did things over time. This is rather unfortunate and quite a few people stumbled across this problem. It would be good to homogenize this at some point, indeed.

PR #435, opened 2024-05-22
Removed nested explicit loop in Allen Cahn 2D FD exact solution

brownbaerchen:
Very minor performance improvement, a first baby step towards streamlining Allen-Cahn implementations as mentioned in #434.

### DIFF
--- a/pySDC/implementations/problem_classes/AllenCahn_2D_FD.py
+++ b/pySDC/implementations/problem_classes/AllenCahn_2D_FD.py
@@ -134,41 +134,6 @@ class allencahn_fullyimplicit(ptype):
         self.work_counters['rhs'] = WorkCounter()
         self.work_counters['linear'] = WorkCounter()
 
-    @staticmethod
-    def __get_A(N, dx):
-        """
-        Helper function to assemble FD matrix A in sparse format.
-
-        Parameters
-        ----------
-        N : list
-            Number of degrees of freedom.
-        dx : float
-            Distance between two spatial nodes.
-
-        Returns
-        -------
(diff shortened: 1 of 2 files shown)


lisawim:
So few changes, but the code looks quite more cleaner now. This is nice! Using the ``np.meshgrid`` is this so much faster than using these two for loops?

brownbaerchen:
> So few changes, but the code looks quite more cleaner now. This is nice! Using the `np.meshgrid` is this so much faster than using these two for loops?

In principle, explicit loops should be avoided in python in general. python is known for being slow compared to compiled languages. But this is no problem in practice, because we only use python to interface to compiled code with numpy etc, which uses shared memory parallelisation and vectorisation to speed up loops.
You can do that manually with explicit loops over lists when using [numba](https://numba.pydata.org), for jit compilation. But if you do neither, you get the very slow python speed. There are a lot of explicit loops in pySDC that are not compiled, but they are only small. For instance, we loop over the collocation nodes in the sweeper, which is just a handful of iterations. It doesn't make a large difference there and because of what goes on underneath, we can't use numpy in this case. And numba struggles with more complicated objects, as far as I know.

However, for the tests of this code, doesn't make a large difference. I guess the majority of the time is not spent in generating the initial conditions. However, it's bad practice and looks different from the mpi4py-fft version of Allen-Cahn. That's why I changed it, without really achieving too much :D 

Maybe you saw me add the `girl scout rule` tag. With this I want to encourage exactly the sort of PR that this is. When you take the time to understand some piece of code, often you spot some way to quickly improve it in small ways. Then you can take an extra 30 mins and make it easier to understand for the next person reading it.

PR #437, opened 2024-05-22
Tweaks for `FrozenClass`

brownbaerchen:
The idea behind `FrozenClass` is to allow to add and change only certain variables in order to prevent bugs from typos.
These are used in `status` objects, for instance. But we want to allow users to add status variables, without having a huge list of every variable anyone has ever used hardcoded in there.

In this PR, I fix a truly horribly BS solution that I implemented when I didn't know class attributes. I wrote a very complicated function that loops through all objects and bypasses the `__set_attr__` functions by using `__dict__`.
The new solution uses a class attribute for `FrozenClass` and derived classes, which carries a list of the allowed variables. This can easily be appended, of course, but it cannot be confused with simple assignment.
This should also give slightly better performance because the variables have to be added to the class only once, whereas they had to be added to each object before.

I did not do this too thoroughly. Basically, I refactored until the tests passed. I am sure there is some BS left in the convergence controllers related to this. But this I want to do when I have more time for refactoring. I just felt that the current version is too bad to leave as is.

### DIFF
--- a/pySDC/core/ConvergenceController.py
+++ b/pySDC/core/ConvergenceController.py
@@ -42,6 +42,7 @@ class ConvergenceController(object):
             params (dict): The params passed for this specific convergence controller
             description (dict): The description object used to instantiate the controller
         """
+        self.controller = controller
         self.params = Pars(self.setup(controller, params, description))
         params_ok, msg = self.check_parameters(controller, params, description)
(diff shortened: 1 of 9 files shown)


tlunet:
Maybe get rid of those `FrozenClass` when you have more time for refactoring :stuck_out_tongue: ...

PR #439, opened 2024-05-30
SDC-DAE MPI Sweepers

lisawim:
The current SDC-DAE sweepers now get MPI versions by adding ``fully_implicit_DAE_MPI`` and ``SemiImplicitDAEMPI`` to the SDC-DAE family! Hooray! For this, I implemented a base class for doing the MPI stuff called ``SweeperMPIDAE`` similar to the ``SweeperMPI`` base class. Then, for the SDC-DAE sweepers we have multiple inheritance again and it seems like it works, which is great.

To make sure the sweepers do work as they should I added a test that checks the order for both sweepers for different cases of DAEs (index one and index two case). I was struggling how to implement the test, because first I would like to write a test that checks if the MPI and non-MPI versions do match as [here](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/tests/test_sweepers/test_MPI_sweeper.py) and I still want to do this. However, I was not able to understand how the stuff using ``argparse`` does work and my tests fail all the time.. Since I don't want to invest too much time (which I would do anyways but I can hear someone calling from far away then..) so I decided to skip that for the first time (but have this still on my to do list in my mind) and wrote instead a test checking the order of accuracy. Another point that is kind of embarrassing for me: Running the file ``accuracy_check_MPI.py`` does work when I run it with ``mpiexec -n 3 python3 accuracy_check_MPI.py`` but fails when I use pytest for the test file, and it returns some "Cant open file [bla bla] - No such file or directory" so maybe I did a mistake using the correct path? Let's see if the test here is successful. Setting path's is still a big mystery for me although I thought I would understand it .. :D

Since this PR is a request of @Ouardghi I also added ``playground_MPI.py`` with some comments how to use and how to run the files. Moreover, the sweepers do also contain some documentation. I'm not sure if this is enough but if he has questions he should not hesitate to ask me!

I appreciate your review very much! Feel free to throw comments at me I'm always happy about that!

### DIFF
--- a/pySDC/projects/DAE/sweepers/SemiImplicitDAEMPI.py
+++ b/pySDC/projects/DAE/sweepers/SemiImplicitDAEMPI.py
@@ -1,0 +1,100 @@
+from mpi4py import MPI
+
+from pySDC.projects.DAE.sweepers.fully_implicit_DAE_MPI import SweeperDAEMPI
+from pySDC.projects.DAE.sweepers.SemiImplicitDAE import SemiImplicitDAE
+
+
+class SemiImplicitDAEMPI(SweeperDAEMPI, SemiImplicitDAE):
+    r"""
+    Custom sweeper class to implement the fully-implicit SDC parallelized across the nodes for solving fully-implicit DAE problems of the form
+
+    .. math::
+        u' = f(u, z, t),
+
+    .. math::
+        0 = g(u, z, t)
+
+    More detailed description can be found in ``SemiImplicitDAE.py``. To parallelize SDC across the method the idea is to use a diagonal :math:`\mathbf{Q}_\Delta` to have solves of the implicit system on each node that can be done in parallel since they are fully decoupled from values of previous nodes. First such diagonal :math:`\mathbf{Q}_\Delta` were developed in [1]_. Years later intensive theory about the topic was developed in [2]_. For the DAE case these ideas were basically transferred.
+
+    Note
+    ----
+    For more details of implementing a sweeper enabling parallelization across the method we refer to documentation in [generic_implicit_MPI.py](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/implementations/sweeper_classes/generic_implicit_MPI.py) and [fully_implicit_DAE_MPI.py](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/projects/DAE/sweepers/fully_implicit_DAE_MPI.py).
+
+    For parallelization across the method for semi-explicit DAEs, differential and algebraic parts have to be treated separately. In ``integrate()`` these parts needs to be collected separately, otherwise information would get lost.
+
+    Reference
+    ---------
+    .. [1] R. Speck. Parallelizing spectral deferred corrections across the method. Comput. Vis. Sci. 19, No. 3-4, 75-83 (2018).
+    .. [2] G. Čaklović,  T. Lunet, S. Götschel, D. Ruprecht. Improving Efficiency of Parallel Across the Method Spectral Deferred Corrections. Preprint, arXiv:2403.18641 [math.NA] (2024).
+    """
+
+    def integrate(self, last_only=False):
+        """
+        Integrates the gradient. Note that here only the differential part is integrated, i.e., the
+        integral over the algebraic part is zero. ``me`` serves as buffer, and root process ``m``
+        stores the result of integral at node :math:`\tau_m` there.
+
+        Parameters
+        ----------
+        last_only : bool, optional
+            Integrate only the last node for the residual or all of them.
+
+        Returns
+        -------
+        me : list of dtype_u
+            Containing the integral as values.
+        """
+
+        L = self.level
+        P = L.prob
+
+        me = P.dtype_u(P.init, val=0.0)
+        for m in [self.coll.num_nodes - 1] if last_only else range(self.coll.num_nodes):
+            recvBufDiff = me.diff[:] if m == self.rank else None
+            recvBufAlg = me.alg[:] if m == self.rank else None
+            integral = P.dtype_u(P.init, val=0.0)
+            integral.diff[:] = L.dt * self.coll.Qmat[m + 1, self.rank + 1] * L.f[self.rank + 1].diff[:]
+            self.comm.Reduce(integral.diff[:], recvBufDiff, root=m, op=MPI.SUM)
+            self.comm.Reduce(integral.alg[:], recvBufAlg, root=m, op=MPI.SUM)
+
+        return me
+
+    def update_nodes(self):
+        r"""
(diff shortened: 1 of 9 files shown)


brownbaerchen:
The pytest problems are probably because of our changes to the pipeline recently. Every project now has it's own environment file, yours is [here](https://github.com/Parallel-in-Time/pySDC/blob/e372a43c621e1bb8ef86c68e02393a3338c096d8/pySDC/projects/DAE/environment.yml). Since you weren't using MPI before, Jakob didn't add it. Please add
```
   - mpich
   - mpi4py>=3.0.0
```
to the dependencies for your project like I also had to do in #438.

pySDC/projects/DAE/sweepers/SemiImplicitDAEMPI.py
```diff
@@ -0,0 +1,100 @@
...
+        P = L.prob
+
+        me = P.dtype_u(P.init, val=0.0)
+        for m in [self.coll.num_nodes - 1] if last_only else range(self.coll.num_nodes):
+            recvBufDiff = me.diff[:] if m == self.rank else None
+            recvBufAlg = me.alg[:] if m == self.rank else None
+            integral = P.dtype_u(P.init, val=0.0)
+            integral.diff[:] = L.dt * self.coll.Qmat[m + 1, self.rank + 1] * L.f[self.rank + 1].diff[:]
+            self.comm.Reduce(integral.diff[:], recvBufDiff, root=m, op=MPI.SUM)
+            self.comm.Reduce(integral.alg[:], recvBufAlg, root=m, op=MPI.SUM)
```
brownbaerchen, line 58: Do you need the separate communications here? Can you not put `me` as receive buffer?

pySDC/projects/DAE/sweepers/fully_implicit_DAE_MPI.py
```diff
@@ -0,0 +1,229 @@
...
+
+    are inherited, which then overwritten by the child class.
+
+    Parameters
+    ----------
+    params : dict
+        Parameters passed to the sweeper.
+    """
+
+    def __init__(self, params):
```
brownbaerchen, line 35: don't need this

pySDC/projects/DAE/sweepers/fully_implicit_DAE_MPI.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # check if Mth node is equal to right point and do_coll_update is false, perform a simple copy
+        if self.coll.right_is_node and not self.params.do_coll_update:
+            # a copy is sufficient
+            root = self.comm.Get_size() - 1
+            if self.comm.rank == root:
+                L.uend = P.dtype_u(L.u[-1])
+            # broadcast diff parts and alg parts separately
+            self.comm.Bcast(L.uend.diff, root=root)
+            self.comm.Bcast(L.uend.alg, root=root)
```
brownbaerchen, line 136: Again, do you need to do this separately?

pySDC/projects/DAE/tests/test_MPI_sweepers.py
```diff
@@ -0,0 +1,31 @@
...
+    import os
+    import subprocess
+
+    # Set python path once
+    my_env = os.environ.copy()
+    my_env['PYTHONPATH'] = '../../..:.'
+    my_env['COVERAGE_PROCESS_START'] = 'pyproject.toml'
+    cwd = '.'
+
+    cmd = ('mpirun -np ' + str(num_procs) + ' python pySDC/projects/DAE/run/accuracy_check_MPI.py').split()
```
brownbaerchen, line 20: Sorry for the comment with the environment. Clearly, you did this correctly already ;)

There is a simple solution to the path stuff. You can import the file you need and then get the path with `.__path__`. I do that [here](https://github.com/Parallel-in-Time/pySDC/blob/e372a43c621e1bb8ef86c68e02393a3338c096d8/pySDC/projects/Resilience/tests/test_vdp.py#L18) for instance.
I am wondering, though. Why do the tests pass if this path is incorrect?

brownbaerchen (commented):
Looks great! I think the test that MPI and non-MPI produce the same result is quite valuable, actually. But I don't know if it's better than testing the order.
@pancetta, what do you think?

pySDC/projects/DAE/tests/test_MPI_sweepers.py
lisawim: I guess it is something local that does not work for me. Maybe something stupid.. I tried now your path solution. Then I get a ``ModuleNotFoundError: No module named 'pySDC'``

brownbaerchen: What python is used when you do `mpiexec`? Can you run `mpiexec which pytest` and check that it's the same as simply `which pytest`? Are you using conda/mamba? Do other scripts work? 
If some imports work, but not others, maybe you installed pySDC with pip but not in editable mode? Maybe try `mpirun pip install -e <path/to/pySDC>`.

lisawim: I made it already work ..somehow. This was interesting when I comment ``my_env['PYTHONPATH'] = '../../..:.'`` then it does work (locally). The tests then pass. So for pushing I added this again since I guess this is needed for the CI pipeline.

pySDC/projects/DAE/sweepers/SemiImplicitDAEMPI.py
lisawim: Yes, you are right. I did change that.

pySDC/projects/DAE/sweepers/fully_implicit_DAE_MPI.py
lisawim: No, I agree with you and did change that.

lisawim:
I totally agree with you @brownbaerchen! Thus I added also a test checking ``uend`` in MPI and non-MPI version.

pySDC/projects/DAE/tests/test_MPI_sweepers.py
brownbaerchen: I recommend to do `pip install -e` instead of adding pySDC to python path because then you don't need to change your python path in your bashrc. I assume this will also work locally then. Of course, you install to a specific python version. So this may not be what you want and you can safely ignore this comment.

brownbaerchen:
LGTM 🚀

pancetta:
Looks good indeed. Can you add this to the documentation of the project or should this happen in #431 ?

lisawim:
I can add this here and we can directly merge then if you would like to.

pancetta:
Thanks, @lisawim. Well done, we appreciate your quick response to @Ouardghi's request.

PR #441, opened 2024-06-17
Fix Gitlab-CI

jakob-fritz:
An update of the gitlab-instance made some changes necessary. Fixing bugs to make Gitlab-CI run again.

Sorry for doing this in a PR instead of my own repo, but there, the Gitlab-CI is not triggered (i.e. this cannot be tested in my own repo).

Finished (and ready for review and merge), when changing from draft to actual PR


jakob-fritz:
Mirroring and before_script runs as expected (as it is done on Login-Nodes), but CLUSTER is not yet available again. Therefore the Script-part of CI does not run.

pancetta (approved):

Issue #442, opened 2024-06-19
Make pySDC compatible with numpy 2.0


jakob-fritz:
There was an update to numpy 2.0, and it seem to have broken the code.

CI-Job with numpy 1.26.4: https://github.com/Parallel-in-Time/pySDC/actions/runs/9542384329/job/26297127121
CI-Job with numpy 2.0.0: https://github.com/Parallel-in-Time/pySDC/actions/runs/9577459034/job/26405728204

Some changes to the code seem to be needed to fix it.

pancetta:
Looks like `numpy.ogrid` now returns a tuple, see [v2.0](https://numpy.org/doc/stable/reference/generated/numpy.ogrid.html) vs. [v1.25](https://numpy.org/doc/1.25/reference/generated/numpy.ogrid.html). Thoughts?

brownbaerchen:
I suppose we just cast the tuples to list...

brownbaerchen:
I did some fixes. If the tests pass on my fork, I will do a PR.
However, there is [this](https://github.com/Parallel-in-Time/pySDC/actions/runs/9577459034/job/26405735156) issue of some incompatibilities of numpy with matplotlib in the parallel SDC project, I think. @pancetta, why do you limit the matplotlib version? We can get rid of this limit or limit the numpy version further, I guess.

tlunet:
Fixed with #445

brownbaerchen:
Well, just removing plotting removes plots from the [website](https://parallel-in-time.org/pySDC/projects/parallelSDC.html)...

pancetta:
Darn, you're right. Thoughts @tlunet ?

tlunet:
None ... to be honest, I secretly hoped that no one would catch that :sweat_smile:

My first guess :  it's either trial and error on a separate branch to solve the deadlock, or duplicate the test without MPI, and regenerate the figures with it.

pancetta:
If I'm not mistaken this is now [fixed](https://parallel-in-time.org/pySDC/projects/parallelSDC.html), right?

brownbaerchen:
Guess, so. I believe @jakob-fritz implemented checks in the CI if all plots are there, so we should know if some are missing (right?)

PR #443, opened 2024-06-19
Some fixes for the new `numpy` 2.0 version.

brownbaerchen:
See #442 for details. Still need to fix the parallel SDC project.

### DIFF
--- a/pySDC/implementations/problem_classes/AllenCahn_Temp_MPIFFT.py
+++ b/pySDC/implementations/problem_classes/AllenCahn_Temp_MPIFFT.py
@@ -123,7 +123,7 @@ class allencahn_temp_imex(ptype):
         L = np.array([self.L] * ndim, dtype=float)
 
         # get local mesh
-        X = np.ogrid[self.fft.local_slice(False)]
+        X = list(np.ogrid[self.fft.local_slice(False)])
         N = self.fft.global_shape()
         for i in range(len(N)):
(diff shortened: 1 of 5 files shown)


pancetta:
Weird.. it looks like the `parallelSDC` tests are stuck..

brownbaerchen:
Was that why you limited the matplotlib version? Is it plotting something weird?

pancetta:
I don't recall exactly what the problem was, but there was one. Maybe tests got stuck..? 🙄

pancetta:
I have no clue what's going on with those three tests. Can someone reproduce this locally?

brownbaerchen:
No works for me locally. Where are even the tests for python versions newer than 3.10? Did they pass? I didn't find them.

pancetta:
Huh, good question! That's more than weird.

tlunet:
I've added more adaptations to numpy2 in the [`qmat-switch`](https://github.com/tlunet/pySDC/tree/qmat-switch), the only problem left are for `parallelSDC` (test not passing).

In practice, removing the plotting function from the test make the test passe for 3 MPI threads, but not for 5 :
https://github.com/tlunet/pySDC/actions/runs/9625706930/job/26550873170#step:5:69

I suspect some additional multi-threading with the new versions of matplotlib (or numpy ...?) that make the mpirun execution reach a ressource limit that provoque the unending wait for the test

Issue #436, opened 2024-05-22
Moving quadrature coefficients generation in a separate package


tlunet:
Hi everyone,

I intend to move in a very near future all the functionalities in pySDC focused on node generation, $Q$ matrix, and $Q_\Delta$ coefficients in a separate repo of in `Parallel-in-Time`, create an associated pip package and make pySDC depend on it. 

Main reason is : more libraries that are not pySDC start to use the same tools to generate those coefficients (SWEET, Alex-Brown integration into Gusto, SDC implementation for Dedalus, etc ...), and each time there is a code duplication for that (I'm especially aware of that because I'm the one doing it :sweat_smile:).

Also, quadrature nodes, Q and QDelta coefficients are not necessarily something only related to pySDC => putting it in a separate repo would help its sharing amongst other SDC developers, and centralize all contributions (new coefficients generation, etc ...) into one same package.

This would require moving some code belonging to pySDC into this new package at some point, hence @pancetta @danielru let me know if you are okay with this. If I don't get any strong objection for it, I'll start to work on it in the next few days.

Cheers :innocent:

tlunet:
Potential names for this package

- `qmatrix`
- `kewmatrix`
- `kewpy`
- `qsdc`

brownbaerchen:
Hm. I didn't know your such a fan of botanical gardens :D How about leaning in to the quadrature aspect? Like `quadrat`? `quadrature` is already taken for a similar package. But you could work "spectral" into the name somehow?

lisawim:
I really like your intention to write a own package for that! :) Hm, what about simply ``spectralquadrature`` (although I quite enjoy ``kewpy``)? But this is rather a long name and probably not wanted. Or ``sqmatrix``? ``sqewpy``? :D

tlunet:
I'm also in favor of short names, not very sure about `quadrature` as it is a very generic name, and there is already the `quadpy` package that can be confused with. 

Didn't understood the joke about botanical gardens tho :sweat:

brownbaerchen:
I thought `kew...` was a reference to the famous [Kew Garden](https://en.wikipedia.org/wiki/Kew_Gardens). I guess I didn't get your joke :D I think `kewpy` is too easily confused with `cupy`.

pancetta:
[...]

PR #445, opened 2024-06-23
`qmat` switch

tlunet:
After implementing most of the core coefficients generation of `pySDC` into [`qmat`](https://github.com/Parallel-in-Time/qmat), here is the associated PR finalizing the switch mentioned in #436.

Main changes :

- `LagrangeApproximation` and `NodeGenerator` classes (and their associated tests) are now in `qmat`
- the base `CollBase` class is now mostly an interface to the `Collocation` class from `qmat`
- the base `Sweeper` class produces implicit and explicit QDelta matrices using `qmat` generators
- additional base `pip` dependency for `pySDC` set to `qmat>=0.1.8`
- docstring / documentation update to link to `qmat` documentation when needed

Additional changes :

- includes complete adaptation to numpy v2, based on the first changes done by @brownbaerchen in his fork. The problematic behavior for the `parallelSDC` project has been patch by removing testing (and coverage) of the plotting function for `preconditioner_playground_MPI`. If this PR is accepted, no need to merge #443 anymore as it was already merged in this PR.
- `Sweeper.get_qDelta_implicit` and `Sweeper.get_qDelta_explicit` only have one argument now : `qd_type`. The previous `coll` one was simply replaced by `self.coll`, as no `Sweeper` instance in `pySDC` code use another `Collocation` class as the base underlying collocation for the sweeper. This was a weird mix between OOP and Procedural Programming that does not make sense but rather the code more complex than it should be.
- complete renaming of the core sub-modules to follow a naming consistent with contribution guidelines, using the refactoring tools from VSCode. Only one issue had to be solved manually for a private class name based attribute in a `Hook` class (see https://github.com/tlunet/pySDC/commit/32f64bf032119f989c9302c3c077d1cb4fc07de8, I won't even try to understand the idea behind this sort of attribute ...)


pyproject.toml
```diff
@@ -30,8 +28,15 @@ dependencies = [
     'sympy>=1.0',
     'numba>=0.35',
     'dill>=0.2.6',
+    'qmat',
```
pancetta, line 31: I guess that should have `>=0.1.8` here, too?

pancetta:
Great work, thanks a lot. Taking the quadrature stuff out of pySDC is a very good idea and I appreciate your work.

pancetta:
Not sure why you're obscuring this PR with so much renaming, though.. this makes it much harder to see what actually happens here.

CONTRIBUTING.md
```diff
@@ -9,6 +9,11 @@ Others are specific conventions chosen for the pySDC library, that may follow Py
 Finally, while `pySDC` provides many base functionalities that implement classical flavors of SDC, it also allows problem-specific applications through Object-Oriented Programming (OOP) and the implementation of custom inherited classes.
 This follows a specific OOP framework, you can look at the page on [custom implementations](./docs/contrib/04_custom_implementations.md) for more details. Additional guideline are also given on how to [document the code](./docs/contrib/05_documenting_code.md) in `pySDC`.
 
+> 📣 Some core components (collocation matrix, $Q-\Delta$ coefficients, Lagrange barycentric interpolation) 
+> are currently implemented and developed in the [`qmat` companion package](https://github.com/Parallel-in-Time/qmat).
```
pancetta, line 13: I'd leave out the `currently`, because this sounds like this is a temporary thing.

pancetta:
Is there a reason why the CI pipeline is back to 40+minutes now? Or is that a GitHub glitch?  I see no obvious cause.

pyproject.toml
tlunet: Done in the next commit

tlunet:
> Not sure why you're obscuring this PR with so much renaming, though.. this makes it much harder to see what actually happens here.

Well, this was a big change anyway, and the naming of the core classes has been confusing me since a long time ... so that's mostly shooting two birds with one stone.

pancetta:
> > Not sure why you're obscuring this PR with so much renaming, though.. this makes it much harder to see what actually happens here.
> 
> Well, this was a big change anyway, and the naming of the core classes has been confusing me since a long time ... so that's mostly shooting two birds with one stone.

Understood. As a preference, I think [smaller PRs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/best-practices-for-pull-requests#write-small-prs) with a clear focus are better, though.

pancetta:
And no, this is not stopping me from merging this, of course!

CONTRIBUTING.md
tlunet: Done

tlunet:
> And no, this is not stopping me from merging this, of course!

I guessed :sweat_smile: ... but don't worry, I'll stick to smaller PR with one dedicated topic from now.

Also in my opinion, it would make sense to drop a new release after the PR is merged.

pancetta:
For sure and already on my todo list!

tlunet:
> Is there a reason why the CI pipeline is back to 40+minutes now? Or is that a GitHub glitch? I see no obvious cause.

Probably a GitHub glitch ... but project testing still take quite some time (especially the `parallelSDC`, `DAE` and `Resilience`). Where are we on testing projects only for one Python version ?

pancetta:
Unless this will significantly shorten the CI time, I'd rather have broad coverage.

tlunet:
> Unless this will significantly shorten the CI time

Probably not, since most are run in parallel ... just significantly less computation ressource used (mostly energy saving)

PR #447, opened 2024-06-24
Add pytest for tests with "benchmark" to Github-Actions

jakob-fritz:
Currently, the tests marked with "benchmark" are run only for benchmarking and only on HPC (via Gitlab). This can lead to codepaths that are not run in Github-Actions leading to passing CI in Github while failing with errors on Gitlab.

For earlier failing, these tests can (and maybe should) be included in Github-Actions.


jakob-fritz:
I'm not sure, if there are other tools that are better suited to collect, run and summarize those benchmarks. (Although jube comes to mind)
And/But I am also unsure if it results in too much overhead especially when combined with coverage, as this needs some additional logging.

pancetta:
To me, this link between gitlab and github has two use cases: benchmarking and HPC-based testing. For the latter, the tests will not run on Github actions. For the first, benchmarking should be done with JUBE anyway and this is independent of coverage etc..

PR #449, opened 2024-06-26
Fix for #448 and migration to scipy 1.14

### LINKED ISSUE #448: `L.uold` will be overwritten when using assignment operator somewhere else
During the usage of the convergence controller ``EstimateEmbeddedError`` that uses the ``StoreUOld`` controller I noticed that the increment is directly computed as zero after the first iteration which could not be correct.  Then it seems like using the assignment operator overwrites the ``L.uold`` value when
```
L.u[m + 1][:] = P.solve_system(rhs, alpha, L.u[m + 1], L.time + L.dt * self.coll.nodes[m])
```
is applied.

Using the assignment operator ``[:]`` should be the correct way to assign correctly to the values of the data type but this leads then instead to kind of trouble of overwriting. Considering the different memory adresses of ``L.uold`` and ``L.u`` have shown that they are different which is weird. @brownbaerchen suggested me to accept this bug for the first time, but we both don't know how large the impact of this bug is. 

Since I don't have any plan where this bug could come from, I reproduced this bug for someone else on a [branch in my fork](https://github.com/lisawim/pySDC/tree/issue_store_uold) and running the file [playground_error.py](https://github.com/lisawim/pySDC/blob/issue_store_uold/pySDC/projects/DAE/run/playground_error.py) shows the things explained above.

brownbaerchen:
Python 3.8 is not compatible with scipy 1.14 at this point, so we had to remove it from the tests.

Closes #448


pancetta:
How are we doing in terms of 3.12?

brownbaerchen:
How do you mean? On my fork all tests pass with python > 3.8.

pancetta:
Do we explicitly test with 3.12?

brownbaerchen:
Yes, we test with 3.9, 3.10, 3.11 and 3.12. I am also developing on 3.12 locally.

PR #452, opened 2024-06-28
FrozenClass performance fix

brownbaerchen:
In #437, I made the names the classes are allowed to have a class attribute of `FrozenClass`. Turns out there are a few issues.
Some convergence controllers added attributes in every step, leading to huge lists. This caused really poor performance. This PR includes a fix for that such that attributes are only ever added once, regardless of whether it is allowed or not.

However, since the allowed attributes are a class attribute of `FrozenClass` they are shared across all derived classes. And I didn't find a quick fix for that. It's not a huge problem. Basically, you add, for instance `error_embedded_estimate` to the status class of level and then you can set `error_embedded_estimate` for the status of the step as well. It's not nice. It doesn't really enforce what the frozen classes are about. I am sure there is a simple solution for this that I am just not able to see at the moment.

### DIFF
--- a/pySDC/helpers/pysdc_helper.py
+++ b/pySDC/helpers/pysdc_helper.py
@@ -1,3 +1,6 @@
+import logging
+
+
 class FrozenClass(object):
     """
     Helper class to freeze a class, i.e. to avoid adding more attributes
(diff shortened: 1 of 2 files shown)


brownbaerchen:
What's the status of this? Anyone want to review / merge?

PR #454, opened 2024-07-01
Change partition of JUWELS

jakob-fritz:
As the nodes of development partitions are not up-and-running yet.

Therefore, the CI-Jobs shall use the "regular" partitions. Can be reverted later, once the development partitions are back online


jakob-fritz:
Closed, as probably the previously used partitions will be available again tomorrow. Then, this change is not needed anymore.

PR #459, opened 2024-07-11
Maintaining website

jakob-fritz:
New PR instead of #458, not with a branch that was rebased earlier (instead of master merged into branch)

Housekeeping for the website of pySDC including:

- Some updates on RST-Files (e.g. change duplicate names of links)
- Added plots for parallelSDC, so that they are included in the website
- Corrected some paths to images
- Building the website will now fail if images are not found and not included, to highlight this issue early when reappearing

### DIFF
--- a/docs/contrib/02_continuous_integration.md
+++ b/docs/contrib/02_continuous_integration.md
@@ -150,7 +150,7 @@ This stage allows to checks how much of the `pySDC` code is tested by the previo
 - `pySDC/tutorial`
 
 This analysis is done in parallel to the test each time a pull is done on any branch (main repository or fork).
-You can look at the current coverage report for the [master branch](https://parallel-in-time.org/pySDC/coverage/index.html) or compare the results with [previous builds](https://app.codecov.io/gh/Parallel-in-Time/pySDC). Codecov will also comment on any pull request, indicating the change of coverage.
+You can look at the current [coverage report for the master branch](https://parallel-in-time.org/pySDC/coverage/index.html) or [compare the results with previous builds](https://app.codecov.io/gh/Parallel-in-Time/pySDC). Codecov will also comment on any pull request, indicating the change of coverage.
 
 During developments, you can also run the coverage tests locally, using :
 
--- a/docs/contrib/02_continuous_integration.md
+++ b/docs/contrib/02_continuous_integration.md
@@ -150,7 +150,7 @@ This stage allows to checks how much of the `pySDC` code is tested by the previo
 - `pySDC/tutorial`
 
 This analysis is done in parallel to the test each time a pull is done on any branch (main repository or fork).
-You can look at the current coverage report for the master branch [here](https://parallel-in-time.org/pySDC/coverage/index.html) or compare the results with previous builds [here](https://app.codecov.io/gh/Parallel-in-Time/pySDC). Codecov will also comment on any pull request, indicating the change of coverage.
+You can look at the current coverage report for the [master branch](https://parallel-in-time.org/pySDC/coverage/index.html) or compare the results with [previous builds](https://app.codecov.io/gh/Parallel-in-Time/pySDC). Codecov will also comment on any pull request, indicating the change of coverage.
 
 During developments, you can also run the coverage tests locally, using :
 
--- a/pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
+++ b/pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
@@ -16,11 +16,11 @@ class AdaptiveCollocation(ConvergenceController):
 
     Use this convergence controller by supplying parameters that the sweeper accepts as a list to the `params`.
     For instance, supplying
-    ```
-    params = {
-        'num_nodes': [2, 3],
-    }
-    ```
+
+        params = {
+            'num_nodes': [2, 3],
+        }
+
     will use collocation methods like you passed to the `sweeper_params` in the `description` object, but will change
     the number of nodes to 2 before the first iteration and to 3 as soon as the 2-node collocation problem is converged.
     This will override whatever you set for the number of nodes in the `sweeper_params`, but you still need to set
--- a/pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
+++ b/pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
@@ -16,9 +16,9 @@ class AdaptiveCollocation(ConvergenceController):
 
     Use this convergence controller by supplying parameters that the sweeper accepts as a list to the `params`.
     For instance, supplying
-    
+
     .. code-block:: python
-    
+
         params = {
             'num_nodes': [2, 3],
         }
--- a/pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
+++ b/pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
@@ -16,7 +16,9 @@ class AdaptiveCollocation(ConvergenceController):
 
     Use this convergence controller by supplying parameters that the sweeper accepts as a list to the `params`.
     For instance, supplying
-
+    
+    .. code-block:: python
+    
         params = {
             'num_nodes': [2, 3],
         }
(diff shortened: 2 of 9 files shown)


docs/contrib/02_continuous_integration.md
```diff
@@ -150,7 +150,7 @@ This stage allows to checks how much of the `pySDC` code is tested by the previo
 - `pySDC/tutorial`
 
 This analysis is done in parallel to the test each time a pull is done on any branch (main repository or fork).
-You can look at the current coverage report for the master branch [here](https://parallel-in-time.org/pySDC/coverage/index.html) or compare the results with previous builds [here](https://app.codecov.io/gh/Parallel-in-Time/pySDC). Codecov will also comment on any pull request, indicating the change of coverage.
+You can look at the current coverage report for the [master branch](https://parallel-in-time.org/pySDC/coverage/index.html) or compare the results with [previous builds](https://app.codecov.io/gh/Parallel-in-Time/pySDC). Codecov will also comment on any pull request, indicating the change of coverage.
```
brownbaerchen, line 153: This sounds like the links are leading to the master branch or the previous builds to me. The old version was less ambiguous.

pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
```diff
@@ -16,11 +16,11 @@ class AdaptiveCollocation(ConvergenceController):
...
-    ```
-    params = {
-        'num_nodes': [2, 3],
-    }
-    ```
+
+        params = {
+            'num_nodes': [2, 3],
+        }
+
```
brownbaerchen, line 23: The ticks indicate that this is code. Why did you remove them?

pySDC/projects/parallelSDC/preconditioner_playground_MPI.py
brownbaerchen, line 1: Does this fix the weird stalling issue that we had when upgrading to new versions of scipy etc?

pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
jakob-fritz: Because they were not rendered correctly (check [the website](https://parallel-in-time.org/pySDC/pySDC/implementations.convergence_controller_classes.adaptive_collocation.html#implementations.convergence_controller_classes.adaptive_collocation.AdaptiveCollocation) ). In local deployment it got rendered correctly with this new syntax

pySDC/projects/parallelSDC/preconditioner_playground_MPI.py
jakob-fritz: Yes, now the CI works again when plotting the images. This is why the whole thing took me so long...

pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
brownbaerchen: @lisawim, what do you usually do in this case? I believe you have quite some experience with writing documentation :D

pySDC/projects/parallelSDC/preconditioner_playground_MPI.py
jakob-fritz: And additionally some minor changes and corrections (e.g. using the with syntax to make use of auto-closing of file or updating error-messages)

pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
tlunet: Actually, code blocks in docstring should probably written like this : https://stackoverflow.com/questions/64451966/python-sphinx-how-to-embed-code-into-a-docstring

Or a more "python-like" approach could be used : https://gitlab.com/tlunet/casper/-/blob/master/casper/maths/analytic.py?ref_type=heads#L79 (not sure how it is rendered by sphinx, tho)

docs/contrib/02_continuous_integration.md
jakob-fritz: Just updated what the link-text says. Should be more specific and less ambiguous now.

pySDC/projects/parallelSDC/preconditioner_playground_MPI.py
brownbaerchen: Very nice! Thanks a lot for that! :)

tlunet: Lol ... not closing the file made the test stall then :sweat_smile: ...

jakob-fritz: Unfortunately not. I gave that a try rather early, but this did not solve the stalling. It may just result in some issue sometime or maybe increased memory or such (as the file is kept open).

Also it did not work out to "just" call the plotting-function as a function, as then un-pickling does not because of the namedTuple.

pySDC/implementations/convergence_controller_classes/adaptive_collocation.py
lisawim: What @tlunet said. :D

jakob-fritz: Thank you @tlunet for the links. I now included the first way in the file. I didn't know about those and "simply" indenting them already makes it keep formatting, but marking it as code is way better.

pySDC/projects/parallelSDC/preconditioner_playground_MPI.py
tlunet: Oh ... what solved the stalling issue then ? Not sure to see exactly what from your changes

jakob-fritz: Yes. I got the plotting out of the MPI-context. There I got plotting to work but only for a single graph; not for all four of them (still unsure why). This is why I added the command line arguments. With this, I can call the simulation with MPI and later the plotting outside of it. But both calls call the file as script and make it load the definition of the named-tuple. Otherwise loading the pkl-file is not possible.

tlunet: Hum ... maybe get rid of those pickle stuff for test could be an idea for https://github.com/orgs/Parallel-in-Time/projects/3, to avoid those kind of issues (which I would guess comes from pickling some MPI communicator or something like that). And use some lightweight text approach, like [this](https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/helpers/testing.py) for instance ... what do you think @brownbaerchen ?

brownbaerchen: Sounds to me like all tasks are opening the file at the same time and the file system will not allow it. I have had similar issues before. Only one task writes the file and that works fine. I guess you essentially open the file only on a single task now as well. Did you try to simply `if comm.rank == 0: <open...>`? I don't think there is a general issue with MPI and pickle.

jakob-fritz: Well, pickling the data is still done in the MPI-context. un-pickling also worked in the MPI-context. Even plotting one figure worked in the MPI-context. But plotting multiple figures did not. So it does not seem to be an issue with pickling and MPI.

brownbaerchen:
LGTM

pySDC/projects/parallelSDC/preconditioner_playground_MPI.py
jakob-fritz: @brownbaerchen , Yes I did. Only rank==0 writes the file and also rank==0 read it for plotting. Now also only one process reads it as only one process runs this code at that point in time anyway.
But I also assumed initially, that it is some issue with reading the file, which did not turn out to be the case, as the issue was in the actual plotting or writing files (did check which part of it).

brownbaerchen: Is it to do with writing the plotting file? Still mysterious...

lisawim:
Please correct me if I missed something in our conservation in this PR here but there are still some plots missed in the ``parallelSDC`` projects on the [pySDC website](https://parallel-in-time.org/pySDC/projects/parallelSDC.html). :(

brownbaerchen:
It seems to me like the website was not generated after merging because the post processing step is skipped. Is this because the GitLab action failed?

jakob-fritz:
Yes, the postprocessing step is executed only once both pipelines ran successfully. As now the gitlab-pipeline failed because the "devel" partition for the benchmark is not yet up-and-running.

I/We (Robert and I) decided for this (both pipelines are required), as otherwise e.g. the coverage is misleading, as it sometimes includes coverage from tests from HPC and sometimes it doesn't.

PR #460, opened 2024-07-23
Added `integrate` method to `RungeKuttaDAE`

lisawim:
Since `SwitchEstimator` requires values at the intermediate stages when applying Runge-Kutta methods, these values will now be computed in `update_nodes`. This required adding the `integrate` method. Here, we cannot inherit from the parent since it uses the `get_full_f` method (splitting of the right-hand side `f` which is currently not possible).

When adding that I noticed that `EDIRK4DAE` achieved order $3$ for a fully-implicit index- $2$ problem instead of $2$ as already set in the current test. I was wondering since order reduction could be expected for higher-index problems. However, when you compute the theoretically derived order conditions for index- $2$ problems in [[E. Hairer, C. Lubich, M. Roche, p. 63]](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://link.springer.com/book/10.1007/BFb0093947&ved=2ahUKEwi16r-NubyHAxXLR_EDHVYbBVMQFnoECBgQAQ&usg=AOvVaw3APQeH4x4n3u1nwATd3ZnN), then you get that the order conditions cannot be satisfied since the coefficient matrix in the Butcher tableau is singular. Thus, setting a wider range of time step sizes like `dt_list = np.logspace(-2.5, -1.0, num=7)` leads to a correct result again.

### DIFF
--- a/pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
+++ b/pySDC/projects/DAE/sweepers/RungeKuttaDAE.py
@@ -102,6 +102,31 @@ class RungeKuttaDAE(RungeKutta):
         lvl.status.unlocked = True
         lvl.status.updated = True
 
+    def integrate(self):
+        r"""
+        Returns the solution by integrating its gradient (fundamental theorem of calculus) at each collocation node.
+        ``level.f`` stores the gradient of solution ``level.u``.
+
+        Returns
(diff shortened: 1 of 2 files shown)


brownbaerchen:
LGTM

PR #464, opened 2024-07-28
Class `DenseOutput` to get dense output of solution

lisawim:
As already mentioned in the [pySDC v6 discussion](https://github.com/Parallel-in-Time/pySDC/discussions/446#discussioncomment-10125864) it would be nice to get the solution also at points not computed in the simulation by doing interpolation. At least I need that for computing a reference solution of the Van der Pol DAE problem that has a quite steep slope requiring a very small time step size $\Delta t\leq 10^{-9}$ to have a good resolution of this. However the computation takes forever.. Thus, a computation of adaptivity to speedup the computation might make a sense here!

To realize this I implemented hooks to log the solution at all nodes together with nodes in `LogSolution` and provided a test for that. After simulation you obtain 
```
u_dense = get_sorted(stats, type="u_dense", sortby="time")
nodes = get_sorted(stats, type="nodes", sortby="time")
```
and then the class `DenseOutput` is used:
```
sol = DenseOutput(nodes, u_dense)
```

Evaluating at requested times `t_eval` you get

```
u_eval = [sol(t) for t in t_eval]
```

The class `DenseOutput` realizes the handling of calling the numerical solution and interpolation at requested time points. For interpolation `qmat` package is used.

I was not sure where to store the `DenseOutput` class in `pySDC` so I decided to first store it in the `DAE` project together with the test.

I assume it is a _nice-to-have_ here also in `pySDC` even routines like `scipy.integrate.solve_ivp` also have that feature. For my issue: Since the Van der Pol DAE is a DAE of index $1$ `scipy.integrate.solve_ivp` can still be used for solving. However, due to the steep slope the routine fails at some point and provided the error message that `Required step size is less than spacing between numbers` where for this problem solving this very stiff example seem to not suitable anymore. Nice to know: I discovered that someone is working on a [routine](https://github.com/JonasBreuling/scipy_dae) of scipy to solve DAEs. I was very happy about that and played around with that. So I put Van der Pol into it and look! I got the same error message.. so sad.. Anyways. Using the `DenseOutput` this should speedup the things.

What do you think about that? As usual: Improvements, corrections, comments, etc. are really appreciated! Many thanks in advance! :)

### DIFF
--- a/pySDC/implementations/hooks/log_solution.py
+++ b/pySDC/implementations/hooks/log_solution.py
@@ -224,7 +224,7 @@ class LogToFile(Hooks):
 
         Returns
         -------
-        str : 
+        str :
             Path of file.
         """
         return f'{cls.path}/{cls.file_name}_{cls.format_index(index)}.pickle'
--- a/pySDC/implementations/hooks/log_solution.py
+++ b/pySDC/implementations/hooks/log_solution.py
@@ -6,19 +6,25 @@ import numpy as np
 
 class LogSolution(Hooks):
     """
-    Store the solution at the end of each step as "u".
+    Store the solution at the end of each step as "u". It is also possible to get the solution
+    on each node as "u_dense" and corresponding collocation nodes as "nodes" at the end
+    of the step.
     """
 
     def post_step(self, step, level_number):
         """
         Record solution at the end of the step
 
-        Args:
-            step (pySDC.Step.step): the current step
-            level_number (int): the current level number
+        Parameters
+        ----------
+        step : pySDC.core.step.Step
+            Current step.
+        level_number : int
+            Current level number.
 
-        Returns:
-            None
+        Returns
+        -------
+        None
         """
         super().post_step(step, level_number)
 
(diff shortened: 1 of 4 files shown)


brownbaerchen (changes requested):
Seems a bit more complicated than needed in places. One of my favourite design principles is [KISS](https://en.wikipedia.org/wiki/KISS_principle), which always reminds me that we are all just a bunch of idiots. ;)
To me, KISS applies mainly in refactoring. After I "finish" something to the point that it does what I want, I go back and check how much more it is doing that I need and how much less generic it is than possible. This is best done by unit testing which tests only a single feature rather then the whole thing. This often exposes what input is really needed for a function. If you think it's needed, you can add an integration test afterwards to test the whole thing is working. 
Everybody has their own style of coding. I learnt that writing the unit tests at the same time as I write the functions is what works best for me. As Robert's book recommendation The Pragmatic Programmer points out, the test is the first client of your code. Meaning that as you switch between writing code and test, you switch between being developer and user. Crucially, you should put yourself in the role of a generic user. Don't develop the test for exactly what you need, do it as generically as possible.

tlunet:
> Seems a bit more complicated than needed in places. One of my favourite design principles is [KISS](https://en.wikipedia.org/wiki/KISS_principle), which always reminds me that we are all just a bunch of idiots. ;) To me, KISS applies mainly in refactoring. After I "finish" something to the point that it does what I want, I go back and check how much more it is doing that I need and how much less generic it is than possible. This is best done by unit testing which tests only a single feature rather then the whole thing. This often exposes what input is really needed for a function. If you think it's needed, you can add an integration test afterwards to test the whole thing is working. Everybody has their own style of coding. I learnt that writing the unit tests at the same time as I write the functions is what works best for me. As Robert's book recommendation The Pragmatic Programmer points out, the test is the first client of your code. Meaning that as you switch between writing code and test, you switch between being developer and user. Crucially, you should put yourself in the role of a generic user. Don't develop the test for exactly what you need, do it as generically as possible.

Totally agree with you ... and I would suggest something similar to your messages @brownbaerchen :wink:

PR #466, opened 2024-07-30
Fix syncing ci

jakob-fritz:
Use a new version of Github2Lab-Action

This new version triggers a new Gitlab-Pipeline even if no new code is pushed. This makes it possible to rerun a Github-Action with rerunning Gitlab-CI. It also means that a scheduled pipeline now triggers a separate Gitlab-CI.

Furthermore start the Gitlab-Actions a minute apart to make sure the querying doesn't fail.

### DIFF
--- a/.github/workflows/gitlab_ci.yml
+++ b/.github/workflows/gitlab_ci.yml
@@ -106,7 +106,7 @@ jobs:
           echo "Merge the two parts of the Merge-Request to test the resulting version"
           git merge "${{ github.event.pull_request.head.sha }}"
       - name: Mirror and wait for Gitlab-CI
-        uses: jakob-fritz/github2lab_action@v0.7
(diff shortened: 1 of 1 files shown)


.github/workflows/gitlab_ci.yml
```diff
@@ -7,7 +7,7 @@ on:
   pull_request_target:
     types: [opened, synchronize, reopened, labeled]
   schedule:
-    - cron: '1 5 * * 1'
+    - cron: '2 5 * * 1'
```
brownbaerchen, line 10: Does this mean you want to run more or less often? Why?

jakob-fritz: Neither, I want to start it a minute later. By this, the other workflow is started a minute earlier than this one, meaning that querying the other workflow works as it is already started

brownbaerchen:
LGTM

PR #467, opened 2024-08-07
Made new plots incorporating first review of adaptivity paper

brownbaerchen:
I had to change a lot of configurations for this, but not much in terms of functionality. I know it's a lot of mess. I apologise if anyone actually tries to review it :D I should have refactored at the time, but now I don't have time..

I also updated a reference for the exact solution of NLS and improved efficiency of the stats helper slightly. Van der Pol now allows relative Newton tolerance. 
Also, my git hook applies black to any file that is staged in my commits. Therefore, when I merge master, it reformats some of the files in playgrounds or tests that are not forced to be correctly formatted at that time. This includes two files in the DAE playground.
Finally, I added a small test that solver and RHS evaluation in Quench are consistent.


pancetta:
Looks like codecov.io is dead again?

PR #470, opened 2024-09-02
Coverage-report now uses less restricted file-names

jakob-fritz:
The coverage-file created in Gitlab was named without the "3.10" suffix. This lead to the issue, that it was not taken into account, when creating the whole coverage report.

This issue now solves this, but changing the name of the created coverage-file. An alternativ would be to remove the need for the suffix "3.10" in another file (`.github/workflows/postprocess.yml` line 36).

Then, the coverage-report should include the coverage information from the cupy-tests.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -63,17 +63,16 @@ jobs:
           coverage run -m pytest --continue-on-collection-errors -v --durations=0 pySDC/tests -m ${{ matrix.env }}
       - name: Make coverage report
         run: |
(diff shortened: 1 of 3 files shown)


jakob-fritz:
Went with the second option (less restrictive file-names when the report is created).
Also, changed to python 3.12 for the report and artifacts, to reflect current python as standard.

The coverage-files from individual tests and data-dir now does not include the python-version anymore to make updating easier, as fewer places need changing. Now this is only determined by which jobs upload artifacts.

brownbaerchen (approved):
Looks great, thanks a lot!

PR #475, opened 2024-09-03
Migrate to ruff

### LINKED ISSUE #473: Migrate from Flakeheaven to ruff
Migrate because flakeheaven does not support Python 3.12 (see https://github.com/flakeheaven/flakeheaven/issues/173)

jakob-fritz:
As Flakeheaven does not support python 3.12, the idea was to migrate to ruff.

Most of the config was reused, but some rules differ. Therefore, some rules were added (See line 90ff in `pyproject.toml`) and some small changes in the code have been done to satisfy the new linter. The changes in the codes are grouped per commit; all changes with the same error are in the same commit.

The statistic (lines of code, number of paths, ...) are not available in ruff.

Closes #473

### DIFF
--- a/pySDC/implementations/hooks/log_solution.py
+++ b/pySDC/implementations/hooks/log_solution.py
@@ -99,10 +99,13 @@ class LogToFile(Hooks):
 
     path = None
     file_name = 'solution'
+
     def logging_condition(L):
         return True
+
     def process_solution(L):
         return {'t': L.time + L.dt, 'u': L.uend.view(np.ndarray)}
+
     def format_index(index):
         return f'{index:06d}'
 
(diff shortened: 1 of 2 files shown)


pySDC/projects/DAE/sweepers/semiImplicitDAE.py
```diff
@@ -26,7 +26,7 @@ class SemiImplicitDAE(FullyImplicitDAE):
         0 = g(\vec{U}_0 + \Delta t (\mathbf{Q} \otimes \mathbf{I}_{n_d}) \vec{U}, \vec{z}, \tau),
 
     where
-    
+
```
brownbaerchen, line 29: Why was this not fixed by black? Do black and ruff have different rules?

brownbaerchen (approved):
Thanks a lot! LGMT

pySDC/projects/DAE/sweepers/semiImplicitDAE.py
jakob-fritz: No, but black ignores the docstring. Ruff does not

PR #476, opened 2024-09-06
Implemented framework for spectral methods and Rayleigh Benard Convection with ultraspherical method

brownbaerchen:
This is quite a large PR and I do apologise. But I think it's finally time to merge some stuff back to the main repository.

I implemented a new spectral helper, which allows to build 2D rectangular geometries with spectral discretizations via direct products of 1D spectral bases. Supported 1D bases are FFTs, Chebychov and Ultraspherical.
Then I implemented a few problems with this, namely heat equation, Burger's equation and Rayleigh Benard Convection. The goal was to make it simple to implement new problems with this. It's a lot more manual than Dedalus. But with a rough understanding of the discretizations it should not be too hard to implement new problems.

This PR does not include any projects done with any of the new problems. I want to merge before because this is already a lot of changes and my branches keep diverging. Also, while I did make an effort to write clean code, I don't have the time to go back and refactor everything too rigorously. Especially the transforms are kind of a mess because I use FFTs to do DCTs and I have not found a clean way to do it, especially with padding. And honestly, I don't know if I will find a cleaner way.. If you spot any large BS or have some hints, I will try, though ;)

The main goal was to do Rayleigh Benard Convection. I will run some more testing, but I was able to confirm the numerical order of SDC with fixed step size when turbulence develops (see [here](https://github.com/user-attachments/files/16907212/RBC-order-130.pdf)) and I copied the spacial part from Dedalus rather closely.
Speaking of Dedalus, I cannot compete in terms of runtime. My code is slower by a factor of maybe 10. It's hard to compare exactly because of SDC and RK and so on. I will probably shave off a few transforms, but Dedalus is optimised in a way that I will not be able to compete with anytime soon. They cache matrix decompositions and things like that to make solving more efficient. Initial tests with porting to GPUs also didn't look promising as GPUs don't excel at solving sparse linear systems. I haven't given up yet, but no promises.
On the other hand, with this implementation it's much easier to see what's actually going on. The optimisation and generality of Dedalus makes some details hard to follow. So maybe this is still of interest to @chelseajohn. Let me know what you think.

If you are interested in the details of Chebychov or Ultraspherical methods, I recommend [this paper](https://doi.org/10.1137/120865458), which is a rather concise summary.

### DIFF
--- a/pySDC/helpers/spectral_helper.py
+++ b/pySDC/helpers/spectral_helper.py
@@ -1,0 +1,1575 @@
+import numpy as np
+import scipy
+from pySDC.implementations.datatype_classes.mesh import mesh
+
+
+class SpectralHelper1D:
+    """
+    Base class for 1D spectral discretizations.
+    """
+
+    fft_lib = scipy.fft
+    sparse_lib = scipy.sparse
+    linalg = scipy.sparse.linalg
+    xp = np
+
+    def __init__(self, N, x0=None, x1=None, useGPU=False, **kwargs):
+        self.N = N
+        self.x0 = x0
+        self.x1 = x1
+        self.L = x1 - x0
+        self.useGPU = useGPU
+
+        if useGPU:
+            self.setup_GPU()
+
+    @classmethod
+    def setup_GPU(cls):
+        """switch to GPU modules"""
+        import cupy as cp
+        import cupyx.scipy.sparse as sparse_lib
+        import cupyx.scipy.sparse.linalg as linalg
+        from pySDC.implementations.datatype_classes.cupy_mesh import cupy_mesh
+
+        cls.xp = cp
+        cls.sparse_lib = sparse_lib
+        cls.linalg = linalg
+
+    def get_Id(self):
+        return self.sparse_lib.eye(self.N)
+
+    def get_zero(self):
+        return 0 * self.get_Id()
+
+    def get_differentiation_matrix(self):
+        raise NotImplementedError()
+
+    def get_integration_matrix(self):
+        raise NotImplementedError()
+
+    def get_wavenumbers(self):
+        """
+        Get the grid in spectral space
+        """
+        raise NotImplementedError
+
+    def get_empty_operator_matrix(self, S, O):
+        """
+        Return a matrix of operators to be filled with the connections between the solution components.
+
+        Args:
+            S (int): Number of components in the solution
+            O (sparse matrix): Zero matrix used for initialization
+
+        Returns:
+            list of lists containing sparse zeros
+        """
+        return [[O for _ in range(S)] for _ in range(S)]
+
+    def get_basis_change_matrix(self, *args, **kwargs):
+        """
+        Some spectral discretization change the basis during differentiation. This method can be used to transfer between the various bases.
+        """
+        return self.sparse_lib.eye(self.N)
+
+    def get_BC(self, kind, **kwargs):
+        """
+        To facilitate boundary conditions (BCs) we use either a basis where all functions satisfy the BCs automatically,
+        as is the case in FFT basis for periodic BCs, or boundary bordering. In boundary bordering, specific lines in
+        the matrix are replaced by the boundary conditions as obtained by this method.
+
+        Args:
+            kind (str): The type of BC you want to implement
+
+        Returns:
+            self.xp.array: Boundary condition
+        """
+        raise NotImplementedError(f'No boundary conditions of {kind=!r} implemented!')
+
+    def get_filter_matrix(self, kmin=0, kmax=None):
+        """
+        Get a bandpass filter.
+
+        Args:
+            kmin (int): Lower limit of the bandpass filter
+            kmax (int): Upper limit of the bandpass filter
+
+        Returns:
+            sparse matrix
+        """
+        k = abs(self.get_wavenumbers())
+
+        kmax = max(k) if kmax is None else kmax
+
+        mask = self.xp.logical_or(k >= kmax, k < kmin)
+
+        F = self.get_Id().tolil()
+        F[:, mask] = 0
+        return F.tocsc()
+
+    def get_1dgrid(self):
+        """
+        Get the grid in physical space
+
+        Returns:
+            self.xp.array: Grid
+        """
+        raise NotImplementedError
+
+
+class ChebychovHelper(SpectralHelper1D):
+    """
+    The Chebychov base consists of special kinds of polynomials, with the main advantage that you can easily transform
+    between physical and spectral space by discrete cosine transform.
+    The differentiation in the Chebychov T base is dense, but can be preconditioned to yield a differentiation operator
+    that moves to Chebychov U basis during differentiation, which is sparse. When using this technique, problems need to
+    be formulated in first order formulation.
+
+    This implementation is largely based on the Dedalus paper (arXiv:1905.10388).
+    """
+
+    def __init__(self, *args, transform_type='fft', x0=-1, x1=1, **kwargs):
+        assert x0 == -1
+        assert x1 == 1
+        super().__init__(*args, x0=x0, x1=x1, **kwargs)
+        self.transform_type = transform_type
+
+        if self.transform_type == 'fft':
+            self.get_fft_utils()
+
+        self.cache = {}
+        self.norm = self.get_norm()
+
+    def get_1dgrid(self):
+        '''
+        Generates a 1D grid with Chebychov points. These are clustered at the boundary. You need this kind of grid to
+        use discrete cosine transformation (DCT) to get the Chebychov representation. If you want a different grid, you
+        need to do an affine transformation before any Chebychov business.
+
+        Returns:
+            numpy.ndarray: 1D grid
+        '''
+        return self.xp.cos(np.pi / self.N * (self.xp.arange(self.N) + 0.5))
+
+    def get_wavenumbers(self):
+        return self.xp.arange(self.N)
+
+    def get_conv(self, name, N=None):
+        '''
+        Get conversion matrix between different kinds of polynomials. The supported kinds are
+         - T: Chebychov polynomials of first kind
+         - U: Chebychov polynomials of second kind
+         - D: Dirichlet recombination.
+
+        You get the desired matrix by choosing a name as ``A2B``. I.e. ``T2U`` for the conversion matrix from T to U.
+        Once generates matrices are cached. So feel free to call the method as often as you like.
+
+        Args:
+         name (str): Conversion code, e.g. 'T2U'
+         N (int): Size of the matrix (optional)
+
+        Returns:
+            scipy.sparse: Sparse conversion matrix
+        '''
+        if name in self.cache.keys() and not N:
+            return self.cache[name]
+
+        N = N if N else self.N
+        sp = self.sparse_lib
+        xp = self.xp
+
+        def get_forward_conv(name):
+            if name == 'T2U':
+                mat = (sp.eye(N) - sp.diags(xp.ones(N - 2), offsets=+2)) / 2.0
+                mat[:, 0] *= 2
+            elif name == 'D2T':
+                mat = sp.eye(N) - sp.diags(xp.ones(N - 2), offsets=+2)
+            elif name[0] == name[-1]:
+                mat = self.sparse_lib.eye(self.N)
+            else:
+                raise NotImplementedError(f'Don\'t have conversion matrix {name!r}')
+            return mat
+
+        try:
+            mat = get_forward_conv(name)
+        except NotImplementedError as E:
+            try:
+                fwd = get_forward_conv(name[::-1])
+                import scipy.sparse as sp
+
+                if self.sparse_lib == sp:
+                    mat = self.sparse_lib.linalg.inv(fwd.tocsc())
+                else:
+                    mat = self.sparse_lib.csr_matrix(sp.linalg.inv(fwd.tocsc().get()))
+            except NotImplementedError:
+                raise NotImplementedError from E
+
+        self.cache[name] = mat
+        return mat
+
+    def get_basis_change_matrix(self, conv='T2T', **kwargs):
+        return self.get_conv(conv)
+
+    def get_integration_matrix(self, lbnd=0):
+        S = self.sparse_lib.diags(1 / (self.xp.arange(self.N - 1) + 1), offsets=-1) @ self.get_conv('T2U')
+        n = self.xp.arange(self.N)
+        if lbnd == 0:
+            S = S.tocsc()
+            S[0, 1::2] = (
+                (n / (2 * (self.xp.arange(self.N) + 1)))[1::2]
+                * (-1) ** (self.xp.arange(self.N // 2))
+                / (np.append([1], self.xp.arange(self.N // 2 - 1) + 1))
+            )
+        else:
+            raise NotImplementedError
+        return S
+
+    def get_differentiation_matrix(self, p=1):
+        '''
+        Keep in mind that the T2T differentiation matrix is dense.
+
+        Args:
+            p (int): Derivative you want to compute
+
+        Returns:
+            numpy.ndarray: Differentiation matrix
+        '''
+        D = self.xp.zeros((self.N, self.N))
+        for j in range(self.N):
+            for k in range(j):
+                D[k, j] = 2 * j * ((j - k) % 2)
+
+        D[0, :] /= 2
+        return self.sparse_lib.csr_matrix(self.xp.linalg.matrix_power(D, p))
+
+    def get_norm(self, N=None):
+        '''get normalization for converting Chebychev coefficients and DCT'''
+        N = self.N if N is None else N
+        norm = self.xp.ones(N) / N
+        norm[0] /= 2
+        return norm
+
+    def get_fft_shuffle(self, forward, N):
+        """
+        In order to more easily parallelize using distributed FFT libraries, we express the DCT via an FFT following
+        doi.org/10.1109/TASSP.1980.1163351. The idea is based on reshuffling the data to be periodic and rotating it
+        in the complex plane. This function returns a mask to do the shuffling.
+
+        Args:
+            forward (bool): Whether you want the shuffle for forward transform or backward transform
+            N (int): size of the grid
+
+        Returns:
+            self.xp.array: Use as mask
+        """
+        xp = self.xp
+        if forward:
+            return xp.append(xp.arange((N + 1) // 2) * 2, -xp.arange(N // 2) * 2 - 1 - N % 2)
+        else:
+            mask = xp.zeros(N, dtype=int)
+            mask[: N - N % 2 : 2] = xp.arange(N // 2)
+            mask[1::2] = N - xp.arange(N // 2) - 1
+            mask[-1] = N // 2
+            return mask
+
+    def get_fft_shift(self, forward, N):
+        """
+        As described in the docstring for `get_fft_shuffle`, we need to rotate in the complex plane in order to use FFT for DCT.
+
+        Args:
+            forward (bool): Whether you want the rotation for forward transform or backward transform
+            N (int): size of the grid
+
+        Returns:
+            self.xp.array: Rotation
+        """
+        k = self.get_wavenumbers()
+        norm = self.get_norm()
+        xp = self.xp
+        if forward:
+            return 2 * xp.exp(-1j * np.pi * k / (2 * N) + 0j * np.pi / 4) * norm
+        else:
+            shift = xp.exp(1j * np.pi * k / (2 * N))
+            shift[0] = 0.5
+            return shift / norm
+
+    def get_fft_utils(self):
+        """
+        Get the required utilities for using FFT to do DCT as described in the docstring for `get_fft_shuffle` and keep
+        them cached.
+        """
+        self.fft_utils = {
+            'fwd': {},
+            'bck': {},
+        }
+
+        # forwards transform
+        self.fft_utils['fwd']['shuffle'] = self.get_fft_shuffle(True, self.N)
+        self.fft_utils['fwd']['shift'] = self.get_fft_shift(True, self.N)
+
+        # backwards transform
+        self.fft_utils['bck']['shuffle'] = self.get_fft_shuffle(False, self.N)
+        self.fft_utils['bck']['shift'] = self.get_fft_shift(False, self.N)
+
+        return self.fft_utils
+
+    def transform(self, u, axis=-1, **kwargs):
+        if self.transform_type == 'dct':
+            return self.fft_lib.dct(u, axis=axis) * self.norm
+        elif self.transform_type == 'fft':
+            result = u.copy()
+
+            shuffle = [slice(0, s, 1) for s in u.shape]
+            shuffle[axis] = self.fft_utils['fwd']['shuffle']
+
+            v = u[(*shuffle,)]
+
+            V = self.fft_lib.fft(v, axis=axis, **kwargs)
+
+            expansion = [np.newaxis for _ in u.shape]
+            expansion[axis] = slice(0, u.shape[axis], 1)
+
+            V *= self.fft_utils['fwd']['shift'][(*expansion,)]
+
+            result.real[...] = V.real[...]
+            return result
+        else:
+            raise NotImplementedError
+
+    def itransform(self, u, axis=-1):
+        assert self.norm.shape[0] == u.shape[axis]
+
+        if self.transform_type == 'dct':
+            return self.fft_lib.idct(u / self.norm, axis=axis)
+        elif self.transform_type == 'fft':
+            result = u.copy()
+
+            expansion = [np.newaxis for _ in u.shape]
+            expansion[axis] = slice(0, u.shape[axis], 1)
+
+            v = self.fft_lib.ifft(u * self.fft_utils['bck']['shift'][(*expansion,)], axis=axis)
+
+            shuffle = [slice(0, s, 1) for s in u.shape]
+            shuffle[axis] = self.fft_utils['bck']['shuffle']
+            V = v[(*shuffle,)]
+
+            result.real[...] = V.real[...]
+            return result
+        else:
+            raise NotImplementedError
+
+    def get_BC(self, kind, **kwargs):
+        """
+        Get boundary condition row for boundary bordering.
+        """
+        if kind.lower() == 'integral':
+            return self.get_integ_BC_row(**kwargs)
+        elif kind.lower() == 'dirichlet':
+            return self.get_Dirichlet_BC_row(**kwargs)
+        else:
+            return super().get_BC(kind, **kwargs)
+
+    def get_integ_BC_row(self, **kwargs):
+        """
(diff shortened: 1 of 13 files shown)


danielru:
I don't have anything useful to say except I am very happy about having an **ultra-spherical** method in pySDC. So ultra.

pancetta:
Can we also have **ultra**-spectral deferred corrections??

pySDC/implementations/problem_classes/Burgers.py
```diff
@@ -0,0 +1,291 @@
...
+
+from pySDC.implementations.datatype_classes.mesh import mesh, imex_mesh
+from pySDC.implementations.problem_classes.generic_spectral import GenericSpectralLinear
+
+
+class Burgers1D(GenericSpectralLinear):
+
+    dtype_u = mesh
+    dtype_f = imex_mesh
+
```
tlunet, line 11: Could you add some class documentation pls :wink:

danielru: Make that ultra-documentation!

pancetta:
Maybe this is a stupid question, but why do you call the good ol' man "Chebychov"? I fail to find a reasonable reference for this way of writing his name. Google wants to refer me to "Chebychev", which to me is the far more common name?

brownbaerchen:
> Can we also have **ultra**-spectral deferred corrections??

Of course we can. I have been thinking about this only superficially and I don't think it's actually useful. Basically, it would be a different quadrature matrix and you would work in spectral space in time. But the advantage of the ultraspherial method is that it is sparse and hence more easy to solve directly. So maybe we should do ultraspherical Runge-Kutta instead.

brownbaerchen:
> Maybe this is a stupid question, but why do you call the good ol' man "Chebychov"? I fail to find a reasonable reference for this way of writing his name. Google wants to refer me to "Chebychev", which to me is the far more common name?

[Wikipedia](https://en.wikipedia.org/wiki/Pafnuty_Chebyshev#Transcription) convinced me that the "o" is more in line with the original Russian pronunciation. Notice also that the German spelling is "Tschebyschow". But I can change it to the more common spelling if you want ;)

pySDC/implementations/problem_classes/Burgers.py
```diff
@@ -5,6 +5,11 @@
 
 
 class Burgers1D(GenericSpectralLinear):
+    """
+    See https://en.wikipedia.org/wiki/Burgers'_equation
+    Discretization is done with a Chebychov method, which requires a first order derivative formulation.
+    Feel free to do a more efficient implementation using an ultraspherical method to avoid the first order business.
+    """
```
tlunet, line 12: Please don't cheat your way in : class documentation implies describing the parameters, so anyone using your class know what can be put in it ... Not sure the wikipedia link helps a lot here ...

pySDC/implementations/problem_classes/Burgers.py
```diff
@@ -125,9 +130,12 @@ def plot(self, u, t=None, fig=None, comp='u'):  # pragma: no cover
...
 class Burgers2D(GenericSpectralLinear):
+    """
+    See documentation of `Burgers1D`.
+    """
+
     dtype_u = mesh
     dtype_f = imex_mesh
-    xp = np
 
     def __init__(self, nx=64, nz=64, epsilon=0.1, fux=2, fuz=1, mode='T2U', **kwargs):
```
tlunet, line 140: why the `**kwargs` here ?

tlunet:
> > Maybe this is a stupid question, but why do you call the good ol' man "Chebychov"? I fail to find a reasonable reference for this way of writing his name. Google wants to refer me to "Chebychev", which to me is the far more common name?
> 
> [Wikipedia](https://en.wikipedia.org/wiki/Pafnuty_Chebyshev#Transcription) convinced me that the "o" is more in line with the original Russian pronunciation. Notice also that the German spelling is "Tschebyschow". But I can change it to the more common spelling if you want ;)

Well, I would prefer Tchebychev :wink: (according to Wikipedia ...)

pySDC/implementations/problem_classes/Burgers.py
tlunet: PS : see https://github.com/Parallel-in-Time/pySDC/blob/master/pySDC/implementations/problem_classes/odeScalar.py for an example.

Also, writing documentation is quite easy with Spyder (with auto generation and rendering ...). But maybe there is also a plugin to do that in Vim ...

pancetta:
Since we're using English here, I'd prefer the common "Chebyshev". As Wikipedia says: "Currently, the English transliteration Chebyshev has gained widespread acceptance". This is also about findability and consistency with the literature (not the one from Russia, that is).

pySDC/implementations/problem_classes/Burgers.py
brownbaerchen: Possible keyword arguments are `useGPU`, for instance.

tlunet:
> Since we're using English here, I'd prefer the common "Chebyshev". As Wikipedia says: "Currently, the English transliteration Chebyshev has gained widespread acceptance". This is also about findability and consistency with the literature (not the one from Russia, that is).

Funny how Wikipedia became a reference now, while 20 years ago it wasn't always considered as trustworthy. Cannot wait a few years when we will use "As ChatGPT says : ..." :sweat_smile:

pySDC/implementations/problem_classes/Burgers.py
tlunet: Why is it not explicit then ?

brownbaerchen: Because this it is passed to the parent class, so that the individual problems don't need to worry about this.

brownbaerchen: Can spyder not autocomplete keyword arguments for parent classes?

tlunet: Depends ... But I see now : yet another point in favor of writing problem class documentation :smiling_imp:

brownbaerchen:
It turns out that it's actually very easy to cache LU decompositions using both scipy and cupy. This is particularly good on GPUs because once solved systems can be efficiently solved again with no detour to CPU needed. The LU decomposition is computed on CPU also when using GPUs. I will work out some adaptive step size selection controller that tries to minimise LU decompositions. Notice that SDC is ideally suited to this because the number of stages is often lower for a given order than in DIRK. That means we can get high order with fewer expensive LU decompositions.

pySDC/implementations/problem_classes/generic_spectral.py
```diff
@@ -0,0 +1,307 @@
...
+
+        Note that in implicit Euler, the right hand side will be composed of the initial conditions. We don't want that in lines that don't depend on time. Therefore, we multiply the right hand side by the mass matrix. This means you can only do algebraic conditions that add up to zero. But you can easily overload this function with something more generic if needed.
+
+        We use a tau method to enforce boundary conditions in Chebychov methods. This means we replace a line in the system matrix by the polynomials evaluated at a boundary and put the value we want there in the rhs at the respective position. Since we have to do that in spectral space along only the axis we want to apply the boundary condition to, we transform back to real space after applying the mass matrix, and then transform only along one axis, apply the boundary conditions and transform back. Then we transform along all dimensions again. If you desire speed, you may wish to overload this function with something less generic that avoids a few transformations.
```
pancetta, line 180: Found on "chov"!

brownbaerchen: I removed the need for the comment as well as the comment by implementing more efficient boundary conditions here.

pancetta:
Are we good to go now?

tlunet:
> Are we good to go now?

Well, I would like to have a more closer look, but I first have to finish correcting my 210 exams ... can you let me until the end of the week ?

pySDC/implementations/problem_classes/Burgers.py
```diff
@@ -0,0 +1,315 @@
...
+       BCr (float): Value at right boundary
+       f (int): Frequency of the initial conditions
+       mode (str): 'T2U' or 'T2T'. Use 'T2U' to get sparse differentiation matrices
+    """
+
+    dtype_u = mesh
+    dtype_f = imex_mesh
+
+    def __init__(self, N=64, epsilon=0.1, BCl=1, BCr=-1, f=0, mode='T2U', **kwargs):
+        self._makeAttributeAndRegister(*locals().keys(), localVars=locals(), readOnly=True)
```
tlunet, line 26: Careful : `self` also belongs to `locals()` at this point and is added to the parameters of the object. If this is not wanted, maybe add a safety check in `_makeAttributeAndRegister` to ignore `self` in the names list ...

pySDC/implementations/problem_classes/Burgers.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        import matplotlib.pyplot as plt
+
+        plt.rcParams['figure.constrained_layout.use'] = True
+        self.fig, axs = plt.subplots()
+        return self.fig
+
+    def plot(self, u, t=None, fig=None, comp='u'):  # pragma: no cover
+        r"""
+        Plot the solution. Please supply a figure with the same structure as returned by ``self.get_fig``.
```
tlunet, line 113: the `Please supply a figure` .... should be in put in the documentation of `fig` below, explicitly mentioning that if not provided, the one provided by `self.get_fig` is used.

pySDC/implementations/problem_classes/HeatEquation_Chebychev.py
```diff
@@ -0,0 +1,350 @@
...
+from pySDC.implementations.datatype_classes.mesh import mesh
+from pySDC.implementations.problem_classes.generic_spectral import GenericSpectralLinear
+
+
+class Heat1DChebychev(GenericSpectralLinear):
+    dtype_u = mesh
+    dtype_f = mesh
+
+    def __init__(self, nvars=128, a=0, b=0, f=1, nu=1.0, mode='T2U', **kwargs):
+        self._makeAttributeAndRegister(*locals().keys(), localVars=locals(), readOnly=True)
```
tlunet, line 14: Same comment as before about `*locals()` and `self` ...

pySDC/implementations/problem_classes/HeatEquation_Chebychev.py
```diff
@@ -0,0 +1,350 @@
+import numpy as np
+from scipy import sparse as sp
+
+from pySDC.core.problem import Problem
+from pySDC.implementations.datatype_classes.mesh import mesh
+from pySDC.implementations.problem_classes.generic_spectral import GenericSpectralLinear
+
+
+class Heat1DChebychev(GenericSpectralLinear):
```
tlunet, line 9: Please put some class documentation (mentioning in particular that problem is solved on [0,1] since it's hard-coded ...

pySDC/implementations/problem_classes/HeatEquation_Chebychev.py
```diff
@@ -0,0 +1,350 @@
...
+        iu = self.index('u')
+        u = self.spectral.u_init
+
+        u[iu] = (
+            xp.sin(self.f * np.pi * self.x) * xp.exp(-self.nu * (self.f * np.pi) ** 2 * t)
+            + (self.b - self.a) / 2 * self.x
+            + (self.b + self.a) / 2
+        )
+
+        if noise > 0:
```
tlunet, line 161: Noise addition code seems duplicated across the Head1D classes ... would be worth putting it in one unique utility function/method somewhere ...

In the same time, would cost a lot to add a possibility to change the seed

pySDC/implementations/problem_classes/RayleighBenard.py
```diff
@@ -0,0 +1,521 @@
...
+from pySDC.implementations.problem_classes.generic_spectral import GenericSpectralLinear
+from pySDC.implementations.datatype_classes.mesh import mesh, imex_mesh
+from pySDC.core.convergence_controller import ConvergenceController
+from pySDC.core.hooks import Hooks
+from pySDC.implementations.convergence_controller_classes.check_convergence import CheckConvergence
+
+
+class RayleighBenard(GenericSpectralLinear):
+    """
+    Rayleigh-Benard Convection is a variation of incompressible Navier-Stokes. See, for instance https://doi.org/10.1007/s00791-020-00332-3.
```
tlunet, line 13: At this point, please write the exact equations you are solving here ... giving an external reference to describe what you are doing here is vicious

pySDC/implementations/problem_classes/RayleighBenard.py
```diff
@@ -0,0 +1,521 @@
...
+        self.cax = []
+        divider = make_axes_locatable(axs[0])
+        self.cax += [divider.append_axes('right', size='3%', pad=0.03)]
+        divider2 = make_axes_locatable(axs[1])
+        self.cax += [divider2.append_axes('right', size='3%', pad=0.03)]
+        return self.fig
+
+    def plot(self, u, t=None, fig=None, quantity='T'):  # pragma: no cover
+        r"""
+        Plot the solution. Please supply a figure with the same structure as returned by ``self.get_fig``.
```
tlunet, line 241: Similar commet as in a previous `plot` method ...

pySDC/implementations/problem_classes/generic_spectral.py
```diff
@@ -0,0 +1,357 @@
...
+        cls.dtype_u = cupy_mesh
+
+        GPU_versions = {
+            mesh: cupy_mesh,
+            imex_mesh: imex_cupy_mesh,
+        }
+
+        cls.dtype_f = GPU_versions[cls.dtype_f]
+
+    def __init__(
```
tlunet, line 30: Parameter documentation should be put with the class documentation, not in the `__init__` docstring. Also, the `setup_GPU` should be mentioned in the class documentation (with a short description of its usage)

tlunet: And furthermore, `*args` and `**kwargs` are not used in the constructor, please remove them (just open the possibility to pass a wrong named argument that won't be taken into account ...)

pySDC/implementations/problem_classes/generic_spectral.py
```diff
@@ -0,0 +1,357 @@
...
+        super().__init__(init=self.spectral.init_forward if spectral_space else self.spectral.init)
+
+        self.work_counters[solver_type] = WorkCounter()
+        self.work_counters['factorizations'] = WorkCounter()
+
+        self.setup_preconditioner(Dirichlet_recombination, left_preconditioner)
+
+        self.cached_factorizations = {}
+
+    def __getattr__(self, name):
```
tlunet, line 95: This is weird ... why not have a `helper` or `sh` attribute used in between instead of the `spectral` ? Would remove this (very) implicit aspect of the code

pySDC/implementations/problem_classes/generic_spectral.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        Setup a sparse linear operator by adding relationships. See documentation for ``GenericSpectralLinear.setup_L`` to learn more.
+
+        Args:
+            LHS (dict): Equations to be added to the operator
+
+        Returns:
+            sparse linear operator
+        """
+        operator = self.spectral.get_empty_operator_matrix()
```
tlunet, line 108: ... especially since you are using `self.spectral.get_[...]` here and not `self.get_[...]` which you could theoretically do ?

pySDC/implementations/problem_classes/generic_spectral.py
```diff
@@ -0,0 +1,357 @@
...
+
+        if Dirichlet_recombination and type(self.axes[-1]).__name__ in ['ChebychevHelper, Ultraspherical']:
+            _Pr = self.spectral.get_Dirichlet_recombination_matrix(axis=-1)
+        else:
+            _Pr = Id
+
+        Pr_lhs = {comp: {comp: _Pr} for comp in self.components}
+        self.Pr = self._setup_operator(Pr_lhs) @ self.Pl.T
+
+    def solve_system(self, rhs, dt, u0=None, *args, skip_itransform=False, **kwargs):
```
tlunet, line 178: You could probably write the equation the implicit Euler update with a M matrix using this method ... would make things more cleared than the not your have trying to describe it with words ...

pySDC/tests/test_helpers/test_spectral_helper.py
```diff
@@ -0,0 +1,662 @@
...
+    u2_pad = u_pad**2
+
+    assert np.allclose(u, u_expect)
+    assert np.allclose(u_pad, u_expect_pad)
+
+    assert np.allclose(u2_hat_expect, helper.transform(u2_pad, padding=_padding))
+    assert not np.allclose(u2_hat_expect, helper.transform(u2)), 'Test is too boring, no dealiasing needed'
+
+
+# @pytest.mark.base
```
tlunet, line 283: Why commented ?

pySDC/tests/test_helpers/test_spectral_helper.py
```diff
@@ -0,0 +1,662 @@
...
+    _A = helper.get_empty_operator_matrix()
+    helper.add_equation_lhs(_A, 'u', {'u': D - Id})
+    A = helper.convert_operator_matrix_to_operator(_A)
+    A = helper.put_BCs_in_matrix(A)
+
+    rhs = helper.put_BCs_in_rhs(np.zeros((1, N)))
+    rhs_hat = helper.transform(rhs, axes=(-1,))
+
+    sol_hat = sp.linalg.spsolve(A, rhs_hat.flatten())
+    # print(_sol_hat)
```
tlunet, line 496: Still needed ?

pySDC/tests/test_helpers/test_spectral_helper.py
```diff
@@ -0,0 +1,662 @@
...
+    Dxx = helper.get_differentiation_matrix(axes=(0,), p=2)
+
+    # generate operator
+    _A = helper.get_empty_operator_matrix()
+    helper.add_equation_lhs(_A, 'u', {'u': Dz - Dxx * 1e-1 - Dx})
+    A = helper.convert_operator_matrix_to_operator(_A)
+
+    # prepare system to solve
+    A = helper.put_BCs_in_matrix(A)
+    # rhs = helper.put_BCs_in_rhs(helper.u_init)
```
tlunet, line 566: Still needed ?

pySDC/tests/test_helpers/test_spectral_helper.py
```diff
@@ -0,0 +1,662 @@
...
+    elif args.test == 'diff':
+        test_differentiation_matrix2D(**vars(args))
+    elif args.test == 'int':
+        test_integration_matrix2D(**vars(args))
+    elif args.test == 'tau':
+        test_tau_method2D(**vars(args))
+    elif args.test == 'dealias':
+        _test_transform_dealias(**vars(args))
+    elif args.test is None:
+        # test_transform(8, 3, 'fft', 'cheby', (-1,))
```
tlunet, line 653: Are commented part still needed ?

brownbaerchen:
Thanks a lot for taking the time to look through all this, @tlunet! I agree with most of your comments, but before I start implementing them, I want to address some big picture things.

Personally, I am not a big fan of properties because `get_` functions explicitly tell you that a computation is involved.  Therefore, you know to call these functions only as often as needed. Even with `u_init`, I am not sure I like it because to me the result is not what I expect from the name. But we have decided to go with this and I am not going rogue here ;) What do you think, @pancetta? If both of you prefer properties over getters, I will reconsider. 

I don't think the individual helpers need the same interface and I want to keep it as modular as possible by use of `kwargs`. For instance, I believe it would be quite easy to add a finite difference base in this framework that needs additional arguments for the stencil. I am not 100% sure that mixing FD and spectral works as I imagine and I don't believe it is useful. But nevertheless, I don't want to inhibit such things now. That being said, I agree that I spammed `kwargs` too much and I will get rid of some of it ;)

Finally, I am not happy with the implementation of the transforms, as are you. The 1D bases have a reasonably clean implementation of non-distributed 1D transforms, but the `SpectralHelper` has the big mess of doing everything in parallel. Particularly because I want to transform over multiple axes at once for FFT bases, I don't see how to put it in the 1D bases, where I would like to have it contained as well. Plus, there is a lot of `mpi4py-fft` specific stuff in there, especially now that I access the transfer classes, which are supposed to be hidden to `mpi4py-fft` users. And the padding implementation is even more mess. Sadly, I will not find a better way now. Definitely, for sure, absolutely, maybe, next year, though...

pySDC/helpers/spectral_helper.py
```diff
@@ -0,0 +1,1692 @@
...
+            p (int): Order of the derivative
+
+        Returns:
+            sparse differentiation matrix
+        """
+        sp = self.sparse_lib
+        xp = self.xp
+        N = self.N
+        l = p
+        from scipy.special import factorial
```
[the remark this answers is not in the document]
brownbaerchen: I like to import things that are used only in special cases only when needing them. Plus, you setup the matrices once before you run anything, so this statement is not traversed often either. I prefer to leave this as is.

pySDC/helpers/spectral_helper.py
```diff
@@ -0,0 +1,1692 @@
...
+        return self.xp.meshgrid(*grids)
+
+    def get_grid(self):
+        """
+        Get grid in physical space
+        """
+        grids = [self.axes[i].get_1dgrid()[self.local_slice[i]] for i in range(len(self.axes))][::-1]
+        return self.xp.meshgrid(*grids)
+
+    def get_fft(self, axes=None, direction='object', padding=None, shape=None, comm=None):
```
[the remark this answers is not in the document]
brownbaerchen: It is not always the same result when we use padding, which we need for dealiasing.
When doing 3/2 dealiasing in RBC with Nx by Nz, we first do the inverse FFT along x, which results in a 3/2Nx by Nz shape, which we pad with zeros to 3/2Nx by 3/2Nz shape before the dct along the other axis, which we do by a then unpadded larger transform. So we need two FFT objects here. The reason we don't let mpi4py-fft do the padding in both directions is the weird counting of wavenumbers in FFT. We have to pad in a different place with the DCT (at the end) rather than the middle as in FFT. This did drive me mad.

tlunet:
> I don't think the individual helpers need the same interface and I want to keep it as modular as possible by use of `kwargs`. For instance, I believe it would be quite easy to add a finite difference base in this framework that needs additional arguments for the stencil. I am not 100% sure that mixing FD and spectral works as I imagine and I don't believe it is useful. But nevertheless, I don't want to inhibit such things now. That being said, I agree that I spammed `kwargs` too much and I will get rid of some of it ;)

Well, you don't need `**kwargs` to have implement more genericity later : just adding a new keyword argument in any function with a default value is enough, and don't break any backward compatibility. Using `**kwargs` mostly hide the effective parameters of your method that a user has to provide, and people have to look into the code to understand how to use a method (same as it is, sadly, very often in pySDC ...). 

And if you do need `**kwargs` (this can happen some times), then you have to document it very well. Because if you don't, that a pain to understand what are the parameters of a function (speaking with experience here, as I spend so many time just reading pySDC code to understant which arguments were required, and if now what was the default value ...)

pySDC/helpers/spectral_helper.py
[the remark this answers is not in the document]
tlunet: Well, most best practices recommend (with good reasons) importing always at the top of the file, except if we are in a special situation which does not seems to be the case here (cf, e.g : https://stackoverflow.com/questions/128478/should-import-statements-always-be-at-the-top-of-a-module). Furthermore, hiding a dependency in a function (or method) does not play in favor of making things explicit ... (and the empty line added by black does make the code a tat ugly). 

I would prefer not to leave this as it is :smirk:

tlunet:
> Finally, I am not happy with the implementation of the transforms, as are you. The 1D bases have a reasonably clean implementation of non-distributed 1D transforms, but the `SpectralHelper` has the big mess of doing everything in parallel. Particularly because I want to transform over multiple axes at once for FFT bases, I don't see how to put it in the 1D bases, where I would like to have it contained as well. Plus, there is a lot of `mpi4py-fft` specific stuff in there, especially now that I access the transfer classes, which are supposed to be hidden to `mpi4py-fft` users. And the padding implementation is even more mess. Sadly, I will not find a better way now. Definitely, for sure, absolutely, maybe, next year, though...

Seems to be a TODO that can be put somewhere then (describing in details the current issues with the code), and solved later when we have more feedback

tlunet:
> Personally, I am not a big fan of properties because `get_` functions explicitly tell you that a computation is involved. Therefore, you know to call these functions only as often as needed. Even with `u_init`, I am not sure I like it because to me the result is not what I expect from the name. But we have decided to go with this and I am not going rogue here ;) What do you think, @pancetta? If both of you prefer properties over getters, I will reconsider.

Properties are the python getters ... and they can be extended by setters for more functionalities : for instance, if you want to change the grid at some point (for, e.g, adaptive meshing), then you can simply do `self.grid = ...` and an internal mechanism can handle the change. Finally, properties can be documented to indicate expensive computation, or eventually cached to be computed only once (they use this kind of mechanism in dedalus for instance ...)

pySDC/helpers/spectral_helper.py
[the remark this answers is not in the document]
tlunet: I was speaking about the test if MPI or not ... furthermore, it's hard to understand why we need `self.comm` and the `comm` attribute later, and why `self.comm` can be `None` and then replaced by the `comm` argument. I don't see how it relates to padding btw

pySDC/helpers/spectral_helper.py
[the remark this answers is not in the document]
tlunet: Well, most best practices recommend (with good reasons) importing always at the top of the file, except if we are in a special situation which does not seems to be the case here (cf, e.g : https://stackoverflow.com/questions/128478/should-import-statements-always-be-at-the-top-of-a-module). Furthermore, hiding a dependency in a function (or method) does not play in favor of making things explicit ... (and the empty line added by black does make the code a tat ugly).

I would prefer not to leave this as it is 😏

pySDC/implementations/problem_classes/generic_spectral.py
brownbaerchen: It's a bit implicit, yes. But it provides a lot of convenience. I want to keep this. I wrote that this happens in the class documentation ;)

pySDC/implementations/problem_classes/generic_spectral.py
brownbaerchen: In this class I always use `self.spectral.[...]`, but if you look in the actual problem implementations, you will find I don't do it there, which makes it a lot more readable. The goal is to do a Dedalus-like thing where users don't need to worry so much. Of course, they need to know what they are doing here a lot more. I don't claim to be competitive with Dedalus. But I think this is fine.

pySDC/helpers/spectral_helper.py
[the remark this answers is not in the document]
brownbaerchen: I disagree, but I gave in. Choose your fights.. :D

tlunet:
Sorry for the delay on that, I guess I can be happy with all the changes done, thanks @brownbaerchen !

I'll merge it then ... :wink:

PR #478, opened 2024-09-09
Coverage-reports from Gitlab are now found and combined

jakob-fritz:
The coverage-reports from the Jobs in Gitlab-CI are now also transferred correctly and combined with the other reports.


brownbaerchen:
I didn't find the artefact with the coverage report, but I trust you that it finally works :D

PR #480, opened 2024-09-13
Bugfix in computing absolute value with cupy datatype

brownbaerchen:
Seems I was too eagerly deleting stuff in #469 ... sorry about that! It's a shame we can't get the GPU tests with multiple GPUs to work.

### DIFF
--- a/pySDC/implementations/datatype_classes/cupy_mesh.py
+++ b/pySDC/implementations/datatype_classes/cupy_mesh.py
@@ -1,5 +1,4 @@
 import cupy as cp
-from pySDC.core.errors import DataError
 
 try:
     from mpi4py import MPI
@@ -37,6 +36,7 @@ class cupy_mesh(cp.ndarray):
         ):
             obj = cp.ndarray.__new__(cls, init[0], dtype=init[2], **kwargs)
             obj.fill(val)
+            cls.comm = init[1]
         else:
             raise NotImplementedError(type(init))
         return obj


pancetta:
What's the matter with multi-GPU tests?

brownbaerchen:
> What's the matter with multi-GPU tests?

I don't remember exactly, maybe @jakob-fritz does. Maybe it was related to job submission?

jakob-fritz:
I don't recall, we tried multi-GPU at all. So I don't know if (and what) issues arise with multi-GPU tests. I know, that we tried the single-GPU test on HPC, but have no insight in any multi-GPU attempts.

pancetta:
Could you @brownbaerchen set up a multi-GPU test, then? Not here, but in another PR?

brownbaerchen:
> Could you @brownbaerchen set up a multi-GPU test, then? Not here, but in another PR?

We have such a test [here](https://github.com/Parallel-in-Time/pySDC/blob/f46209f26ca77309531c3b43dcb2eb34fa82f078/pySDC/tests/test_sweepers/test_MPI_sweeper.py#L178), but we skip it because we don't know how to run it.

pancetta:
Not sure I understand the problem..? Why is it difficult to run it if you are on JUWELS?

brownbaerchen:
Currently, we use [this script](https://github.com/Parallel-in-Time/pySDC/blob/f46209f26ca77309531c3b43dcb2eb34fa82f078/etc/juwels_cupy.sh) to launch pytest on JUWELS. Usually, we do MPI based tests by executing shell commands from within python scripts. But I am not sure that we can launch a larger job from within the job. I believe we tried posting `srun` commands, but I am not totally sure. Also, the first job that runs pytest will collect the coverage report, how does that work when the code is run by a different job on a different node? I am sure it's possible but we decided to postpone it. Maybe we can try again now that we are happy with the rest of the GitLab integration.

pancetta:
Well, we do have multiple `pytest` calls in the pipeline anyway, so I guess this should be possible within a single step? @jakob-fritz ideas are welcome.

PR #481, opened 2024-09-13
TL: first revision for the parallel SDC theory paper

tlunet:


pancetta:
Huh? You pushed `5.5.1` in #474 already..

pySDC/core/sweeper.py
```diff
@@ -263,3 +263,15 @@ def level(self, L):
     @property
     def rank(self):
         return 0
+
+    def updateVariableCoeffs(self, k):
+        """
+        Eventually update QDelta implicit coefficients if variable ...
```
brownbaerchen, line 269: Why `Eventually`? That sounds like it will come in the future

tlunet: Nah, it's just that most preconditioners keep the same coefficient for different `k`

brownbaerchen:
> Huh? You pushed `5.5.1` in #474 already..

That looks like a problem with the diff on GH. I wouldn't worry about it.

brownbaerchen:
Now is the perfect time to automate linting. I updated mine to ruff this morning myself in less than 2 mins using our [guide](https://github.com/Parallel-in-Time/pySDC/blob/master/docs/contrib/02_continuous_integration.md).

pySDC/core/sweeper.py
brownbaerchen: Ah, I see! This is a false friend, I guess. https://www.oed.com/search/dictionary/?scope=Entries&q=eventually

pancetta: `Potentially`?

tlunet: Well, this one was vicious ...

PR #491, opened 2024-09-26
Increased performance of tau methods

brownbaerchen:
We need to replace some lines in the matrix with the boundary conditions in tau methods. Then we also need to replace the respective values in the right hand side with the values at the boundary. This PR makes the second part more efficient. After profiling on GPUs, it turns out that the previous implementation was rather expensive. By caching a mask that can streamline this operation, solving systems with already factored matrices is up to 10x faster on GPUs and 2x faster on CPUs than before, depending on resolution.


pancetta:
Please merge `master` again

PR #492, opened 2024-09-26
More efficient `eval_f` in RBC on GPU

brownbaerchen:
After profiling the code on GPUs some more, I noticed some efficiency can be gained from reducing the number of matrix multiplications. That is all that I am doing here: Rather than computing derivatives by multiplying each component with the derivative matrix separately, I construct a large matrix containing the derivative matrix multiple times in order to compute all derivatives simultaneously. Also, I cache the $L$ matrix in the basis of Chebychov T polynomials, rather than computing it every time. This is not a huge difference, but saves a few ms in every call of the function on both CPU and GPU.


pancetta:
Please merge `master` again

PR #495, opened 2024-10-17
Resilience experiments with Rayleigh-Benard convection

brownbaerchen:
This PR includes a configuration for Rayleigh-Benard convection (RBC) on a 256x128 grid with Rayleigh number of $2^4$. Reference solutions are computed for t=20 and t=21. Individual runs for fault experiments then start at t=20 and go to t=21. There is a convergence controller that makes sure the target time is reached to accuracy of $10^{-9}$, which is sufficiently accurate to use the reference solution. I verified that I get the order of the collocation problem relative to such a reference solution, even if the residual stays at order 1 during the runs. The latest theory is that this is because the problem is perturbed by the method that enforces the boundary conditions. This perturbation is missing in the residual and I haven't found a good way to represent it there (yet). This matters when the perturbation is not small, i.e. the resolution is insufficient in z-direction. The configuration used here, however, is sufficiently resolved for this not to cause any issues.

### DIFF
--- a/pySDC/projects/Resilience/RBC.py
+++ b/pySDC/projects/Resilience/RBC.py
@@ -1,0 +1,165 @@
+# script to run a Rayleigh-Benard Convection problem
+from pySDC.implementations.problem_classes.generic_spectral import compute_residual_DAE
+from pySDC.implementations.problem_classes.RayleighBenard import RayleighBenard
+from pySDC.implementations.controller_classes.controller_nonMPI import controller_nonMPI
+from pySDC.projects.Resilience.hook import hook_collection, LogData
+from pySDC.projects.Resilience.strategies import merge_descriptions
+from pySDC.implementations.sweeper_classes.imex_1st_order import imex_1st_order
+from pySDC.core.convergence_controller import ConvergenceController
+
+from pySDC.core.errors import ConvergenceError
+
+
+class ReachTendExactly(ConvergenceController):
+
+    def setup(self, controller, params, description, **kwargs):
+        defaults = {
+            "control_order": +50,
+            "Tend": None,
+        }
+        return {**defaults, **super().setup(controller, params, description, **kwargs)}
+
+    def get_new_step_size(self, controller, step, **kwargs):
+        L = step.levels[0]
+        L.status.dt_new = min([L.params.dt, self.params.Tend - L.time - L.dt])
+
+
+def run_RBC(
+    custom_description=None,
+    num_procs=1,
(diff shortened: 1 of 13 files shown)


pancetta:
Hm, is anything of this tested somewhere?

brownbaerchen:
No, that's too expensive, I'm afraid. I had to choose a large resolution to get low residuals in the collocation problem and now these experiments take about 40s+ on my laptop each. Also, they need the reference solution. Computing that takes multiple hours, and checking it in is not great either.

brownbaerchen:
When doing experiments with work-precision plots, I noticed some issues with the reference solution. These are fixed now, there is now a test that the time that is supposed to be reached is actually reached very closely. The result is also more sensitivity to faults in RBC.
Also, I implemented a convergence controller that rounds the step size. The idea is that the cache of matrix factorisations gets a higher hit rate if there are fewer options for step sizes. I.e. if adaptivity wants a step size of 0.612412, but you have a factorisation for 0.6 in the cache, then it does not need to be recomputed if the step size is rounded to that value. This is also tested.
The parameters for the work-precision plot with RBC are not final, but since this includes some tests, I thought I add this already to this PR and change again in the future.

pySDC/projects/Resilience/RBC.py
```diff
@@ -0,0 +1,330 @@
...
+    return data
+
+
+RayleighBenard._u_exact = RayleighBenard.u_exact
+RayleighBenard.u_exact = u_exact
+
+PROBLEM_PARAMS = {'Rayleigh': 2e4, 'nx': 256, 'nz': 128}
+
+
+class ReachTendExactly(ConvergenceController):
```
pancetta, line 63: No documentation here?

pySDC/projects/Resilience/RBC.py
```diff
@@ -0,0 +1,330 @@
...
+    ax.loglog(dts, errors, label=f'{num_nodes} nodes', marker='x')
+    ax.loglog(
+        dts, [max_error * (me / dts[0]) ** (2 * num_nodes - 1) for me in dts], ls='--', label=f'order {2*num_nodes-1}'
+    )
+    ax.set_xlabel(r'$\Delta t$')
+    ax.set_ylabel(r'global error')
+    ax.legend(frameon=False)
+
+
+def test_order(t=14, dt=1e-1, steps=6):
```
pancetta, line 278: Sure you want to call this `test_xxx` ?

PR #509, opened 2024-12-20
Perform at least one SDC iteration

brownbaerchen:
Abdel and I were running into issues from time to time when the SDC residual of the prediction was low enough proclaim convergence. In this case, the code just does not do anything anymore but counting up the time and returning this solution.
I would consider this a bug.
The fix here is a bit weird because in single level SDC the sweep and iterations variables are interchangeable. Do you see any problem with doing it like this @pancetta?

### DIFF
--- a/pySDC/implementations/convergence_controller_classes/check_convergence.py
+++ b/pySDC/implementations/convergence_controller_classes/check_convergence.py
@@ -73,7 +73,7 @@ class CheckConvergence(ConvergenceController):
 
         # get residual and check against prescribed tolerance (plus check number of iterations)
         iter_converged = S.status.iter >= S.params.maxiter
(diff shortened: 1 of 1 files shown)


pancetta:
Well. Not the first time we're running into this, but I'm not sure this is actually a bug. It depends on how you count, but if the prediction is good enough, by iterate? Take a problem that is constant for some time, then the prediction is already fine.

brownbaerchen:
The problem is that the solution will remain constant forever in this case unless there is some time-dependent forcing.

pancetta:
True, if the prediction is just 0 or copy.

pancetta:
I think I'd rather agree on always doing a single iteration (at least) instead of distinguishing this here. So, I'd be ok with your first approach.

PR #512, opened 2025-01-17
FieldsIO implementation for 0D to 2D cartesian grid fields, with MPI

tlunet:
A small issue I encountered, in my attempt to use pySDC generated data to train a ML model for Rayleigh Benard convection, is that there was no efficient options (that I knew of) to write many (many many) field solutions when running many space-parallel simulations, hence a "tiny" technical problem.

So I quickly implemented a `FieldsIO` generic (abstract) class to write multiple field solution into a binary file, and specialized it to scalar, 1D and 2D fields written on Cartesian grids. All are compatible with Numpy arrays, and each class is as generic as possible, such that most problem class using one or more `mesh` class for their state variable can use it.

Also, I added the MPI-IO functionalities to the 1D and 2D grid classes, that can be used with any type of Cartesian block decomposition : user simply has to specify its own local bounds at start, and the classes automatically deals with the collective writing of the local fields.

Testing is done in sequential and parallel, and the parallel tests uses some basic `BlockDecomposition` class, that can be used up to 3D problems.

## Note

This was mostly inspired from previous works during my PhD. It may be moved in a near future to a more generic package (as `qmat`), since there is also an high-order interpolation algorithm (also compatible with MPI cartesian decomposition, developed during the end of my PhD) that I would like to make as a standalone package, and was needing this specific part.

## Bonus

CI uses now a "modern" python approach that installs `pySDC` in the test environment, so tests do not need some `sys.path` modification anymore, to ensure `pySDC` is in the path when running tests ...

### DIFF
--- a/pySDC/helpers/fieldsIO.py
+++ b/pySDC/helpers/fieldsIO.py
@@ -38,6 +38,13 @@ Notes
 -----
 🚀 :class:`Cart1D` and :class:`Cart2D` are compatible with a MPI-based cartesian decomposition.
 See :class:`pySDC.tests.test_helpers.test_fieldsIO.writeFields_MPI` for an illustrative example.
+
+Warning
+-------
+To use MPI collective writing, you need to call first the class methods :class:`Cart1D.initMPI` 
+or :class:`Cart2D.initMPI` from the associated class (cf their docstring).
+Also, their associated `setHeader` methods **must be given the global grids coordinates**,
+wether code is run in parallel or not. 
 """
 import os
 import numpy as np
@@ -82,6 +89,8 @@ class FieldsIO:
 
     tSize = T_DTYPE().itemsize
 
+    ALLOW_OVERWRITE = False
+
     def __init__(self, dtype, fileName):
         """
         Parameters
@@ -91,7 +100,7 @@ class FieldsIO:
         fileName : str
             File.
         """
-        assert dtype in DTYPES_AVAIL, f"dtype not available ({dtype})"
+        assert dtype in DTYPES_AVAIL, f"{dtype=} not available"
         self.dtype = dtype
         self.fileName = fileName
         self.initialized = False
@@ -170,6 +179,11 @@ class FieldsIO:
         assert self.header is not None, "header must be set before initializing FieldsIO"
         assert not self.initialized, "FieldsIO already initialized"
 
+        if not self.ALLOW_OVERWRITE:
+            assert not os.path.isfile(
+                self.fileName
+            ), "file already exists, use FieldsIO.ALLOW_OVERWRITE = True to allow overwriting"
+
         with open(self.fileName, "w+b") as f:
             self.hBase.tofile(f)
             for array in self.hInfos:
@@ -240,8 +254,8 @@ class FieldsIO:
         """Number of fields currently stored in the binary file"""
         return int((self.fileSize - self.hSize) // (self.tSize + self.fSize))
 
-    def check(self, idx):
-        """Utility method to check a positional index for a stored field"""
+    def formatIndex(self, idx):
+        """Utility method to format a fields index to a positional integer (negative starts from last field index, like python lists)"""
         nFields = self.nFields
         if idx < 0:
             idx = nFields + idx
@@ -262,7 +276,7 @@ class FieldsIO:
 
     def time(self, idx):
         """Time stored at a given field index"""
-        idx = self.check(idx)
+        idx = self.formatIndex(idx)
         offset = self.hSize + idx * (self.tSize + self.fSize)
         with open(self.fileName, "rb") as f:
             t = np.fromfile(f, dtype=T_DTYPE, count=1, offset=offset)[0]
@@ -285,7 +299,7 @@ class FieldsIO:
         field : np.ndarray
             Read fields in a numpy array.
         """
-        idx = self.check(idx)
+        idx = self.formatIndex(idx)
         offset = self.hSize + idx * (self.tSize + self.fSize)
         with open(self.fileName, "rb") as f:
             f.seek(offset)
@@ -494,7 +508,12 @@ class Cart1D(Scal0D):
     def initialize(self):
         """Initialize the binary file (write header) in MPI mode"""
         if self.MPI_ROOT:
-            super().initialize()
+            try:
+                super().initialize()
+            except AssertionError as e:
+                print(f"{type(e): {e}}")
+                self.comm.Abort()
+
         if self.MPI_ON:
             self.comm.Barrier()  # Important, should not be removed !
             self.initialized = True
@@ -502,7 +521,7 @@ class Cart1D(Scal0D):
     def addField(self, time, field):
         """
         Append one field solution at the end of the file with one given time,
-        eventually using MPI.
+        possibly using MPI.
 
         Parameters
         ----------
@@ -538,7 +557,7 @@ class Cart1D(Scal0D):
     def readField(self, idx):
         """
         Read one field stored in the binary file, corresponding to the given
-        time index, eventually in MPI mode.
+        time index, possibly in MPI mode.
 
         Parameters
         ----------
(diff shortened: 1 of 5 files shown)


pySDC/helpers/fieldsIO.py
```diff
@@ -0,0 +1,731 @@
...
+            return registered
+
+        return wrapper
+
+    def initialize(self):
+        """Initialize the file handler : create the file with header, removing any existing file with the same name"""
+        assert self.header is not None, "header must be set before initializing FieldsIO"
+        assert not self.initialized, "FieldsIO already initialized"
+
+        with open(self.fileName, "w+b") as f:
```
brownbaerchen, line 173: I think it would be handy to implement a check if the file already exists and add a parameter to this function to raise an error if the file already exists by default. This would protect against overwriting previous simulation results by accident.

pySDC/helpers/fieldsIO.py
```diff
@@ -0,0 +1,731 @@
...
+        with open(self.fileName, "ab") as f:
+            np.array(time, dtype=T_DTYPE).tofile(f)
+            field.tofile(f)
+
+    @property
+    def nFields(self):
+        """Number of fields currently stored in the binary file"""
+        return int((self.fileSize - self.hSize) // (self.tSize + self.fSize))
+
+    def check(self, idx):
```
brownbaerchen, line 243: This function is doing a bit more than its name suggests. How about `convertToPositiveIndex`?

pySDC/helpers/fieldsIO.py
```diff
@@ -0,0 +1,731 @@
...
+        if self.MPI_ROOT:
+            super().initialize()
+        if self.MPI_ON:
+            self.comm.Barrier()  # Important, should not be removed !
+            self.initialized = True
+
+    def addField(self, time, field):
+        """
+        Append one field solution at the end of the file with one given time,
+        eventually using MPI.
```
brownbaerchen, line 505: You are using "eventually" incorrectly again 🙈

pySDC/helpers/fieldsIO.py
```diff
@@ -0,0 +1,731 @@
...
+
+        for iVar in range(self.nVar):
+            offset = offset0 + (iVar * self.nX + self.iLocX) * self.itemSize
+            self.MPI_WRITE_AT(offset, field[iVar])
+        self.MPI_FILE_CLOSE()
+
+    def readField(self, idx):
+        """
+        Read one field stored in the binary file, corresponding to the given
+        time index, eventually in MPI mode.
```
brownbaerchen, line 541: eventually

pySDC/helpers/fieldsIO.py
```diff
@@ -0,0 +1,731 @@
...
+        nVar : int
+            Number of 1D variables stored.
+        gridX : np.1darray
+            The grid coordinates in x direction.
+        gridY : np.1darray
+            The grid coordinates in y direction.
+
+        Note
+        ----
+        When used in MPI decomposition mode, `gridX` and `gridY` **must** be the global grid.
```
brownbaerchen, line 601: Is there a way to enforce this by raising an error? This note seems easy to miss.. Also, would be nice to give only the local grid here. Getting the global grid would have been quite difficult for my large GPU Gray-Scott run..

pySDC/helpers/fieldsIO.py
```diff
@@ -0,0 +1,731 @@
...
+    @classmethod
+    def setupMPI(cls, comm: MPI.Intracomm, iLocX, nLocX, iLocY, nLocY):
+        super().setupMPI(comm, iLocX, nLocX)
+        cls.iLocY = iLocY
+        cls.nLocY = nLocY
+
+    def addField(self, time, field):
+        """
+        Append one field solution at the end of the file with one given time,
+        eventually using MPI.
```
brownbaerchen, line 654: You really like this word :D

brownbaerchen (commented):
Looks very nice! I really have just two main issues:
 - need to write global grid but local fields in MPI version of 2d cartesian grid
 - I would like protection against accidental overwrite of previous files

pySDC/helpers/fieldsIO.py
```diff
@@ -0,0 +1,731 @@
...
+>>> # Read the file using a the generic interface
+>>> fIn = FieldsIO.fromFile("file.pysdc")
+>>> times = fIn.times
+>>> assert len(times) == fIn.nFields
+>>> tEnd, uEnd = fIn.readField(-1)
+>>> assert tEnd == times[-1]
+
+Notes
+-----
+🚀 :class:`Cart1D` and :class:`Cart2D` are compatible with a MPI-based cartesian decomposition.
```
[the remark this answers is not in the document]
tlunet: Yes :smile:

pySDC/helpers/fieldsIO.py
```diff
@@ -0,0 +1,731 @@
...
+    def __init__(self, dtype, fileName):
+        """
+        Parameters
+        ----------
+        dtype : np.dtype
+            The data type of the fields values.
+        fileName : str
+            File.
+        """
+        assert dtype in DTYPES_AVAIL, f"dtype not available ({dtype})"
```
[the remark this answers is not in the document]
tlunet: Agree ... just forgot how to do that, and was too lazy to search for it :sweat_smile:

pySDC/helpers/fieldsIO.py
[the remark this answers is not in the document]
brownbaerchen: I vote to remove it. No idea what my vim will show :D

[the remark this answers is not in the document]
tlunet: Come on ! Emoji are fully integrated now in ASCII, even terminal shows it ...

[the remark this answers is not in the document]
brownbaerchen: Yeah ok, seems to work. Personally, I don't like to have any symbols in code that are not on my keyboard, but I suppose it's personal preference.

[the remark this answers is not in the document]
tlunet: You can simply use https://github.com/ikatyang/emoji-cheat-sheet/blob/master/README.md and copy paste :wink:

pySDC/helpers/fieldsIO.py
tlunet: At least there is one part of the code where I used it correctly

pySDC/helpers/fieldsIO.py
tlunet: This may a bit tricky ... both checking if the grid is global (and not local) and the possibility to give the local grid only requires additional all-to-all communications. And at then end, those are only 1D arrays that each processors can totally generate on their own, since normally each process start with the same problem parameters.

So in the attempt of making stuff simple, I would simply add some dedicated section in the module docstring, and blame any user for not reading it.

brownbaerchen: Can you explain why the fields can be written to disk in parallel, but the grid cannot?

tlunet: Grids are encoded in the header, that is only written by rank 0. Since it's only done once when creating the file, this makes things way easier. Also when writing collectively, each process needs to know the global size of the problem, which is directly determined from the global grid that every process stores.

brownbaerchen: I suppose we can merge as is then and eventually find a find a different solution with distributed writing of the grid once we need it.

tlunet: Did you got that the grids are simply the 1D coordinates in each directions, not the full N-dimensional grid ? Because if not, maybe the naming could be corrected :sweat_smile:

brownbaerchen: I did not get that! Ok, that is all fine then.

tlunet: Then I'll change the naming to `coord[...]`, which make be more explicit ... don't merge it yet !

brownbaerchen (approved):

tlunet:
All seems fine now, can be merge anytime

pancetta:
Go ahead, can’t do this today.On 18. Jan 2025, at 10:37, Thibaut Lunet ***@***.***> wrote:﻿
All seems fine now, can be merge anytime

PR #513, opened 2025-01-21
Firedrake coupling

brownbaerchen:
This PR includes things that are needed to couple pySDC and Firedrake and a small heat equation example. Seems I cannot add @jshipton as a reviewer, but you are very welcome to add comments anyways.

### What is needed:
 - data type that wraps Firedrake data and that pySDC can use
 - Communicator that wraps the Firedrake ensemble communicators and gives it an MPI communicator interface to use parallel implementations within pySDC

### Also included in this PR:
 - Small fixes for the generic implicit MPI and Runge-Kutta sweepers, which were more numpy-specific than we want
 - Basic forced 1D heat equation example, which is essentially the same as the Fenics test problem, but using Firedrake
 - Tutorial for running this serially, as well as MPI-parallel with diagonal SDC
 - Tests for the datatype that make sure the result is correct and we didn't accidentally do shallow copies
 - Tests for the heat equation that make sure the result is close to exact
 - Regression tests for the tutorial, serial and parallel
 - Another test environment using the Firedrake docker-hub image. As you can see there, the only dependence that pySDC has that is not part of Firedrake is qmat. So, while Firedrake is not easy to install, once you have it, adding pySDC is simple. The test environment is a bit complicated because it also installs Gusto, which we will use in the future

### Not included in this PR
Coupling between pySDC and Gusto. While this works in principle, it needs a lot of cleanup and will be a slightly larger PR. So we decided to merge the two couplings separately. We did not do any more elaborate equations here than heat equation because including elaborate versions of equations is one of the features of Gusto and that will come with the Gusto coupling.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -210,9 +210,10 @@ jobs:
         run: |
           . /home/firedrake/firedrake/bin/activate
 
-          mv data data_firedrake
+          cd ./pySDC
+          mv data ../data_firedrake
           coverage combine
-          mv .coverage coverage_firedrake.dat
+          mv .coverage ../coverage_firedrake.dat
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -169,6 +169,57 @@ jobs:
           path: |
             data_libpressio
             coverage_libpressio.dat
+
+  user_firedrake_tests:
+    runs-on: ubuntu-latest
+    container:
+      image: firedrakeproject/firedrake-vanilla:latest
+      options: --user root
+      volumes:
+        - ${{ github.workspace }}:/repositories
+    defaults:
+      run:
+        shell: bash -l {0}
+    steps:
+      - name: Checkout pySDC
+        uses: actions/checkout@v4
+        with: 
+          path: ./pySDC
+      - name: Checkout gusto
+        uses: actions/checkout@v4
+        with:
+          repository: firedrakeproject/gusto
+          path: ./gusto_repo
+      - name: Install pySDC
+        run: |
+          . /home/firedrake/firedrake/bin/activate
+          python -m pip install --no-deps -e /repositories/pySDC
+          python -m pip install qmat
+      - name: Install gusto
+        run: |
+          . /home/firedrake/firedrake/bin/activate
+          python -m pip install -e /repositories/gusto_repo
+      - name: run pytest
+        run: |
+          . /home/firedrake/firedrake/bin/activate
+          firedrake-clean
+          cd ./pySDC
+          coverage run -m pytest --continue-on-collection-errors -v --durations=0 /repositories/pySDC/pySDC/tests -m firedrake
+        timeout-minutes: 120
+      - name: Make coverage report
+        run: |
+          . /home/firedrake/firedrake/bin/activate
+
+          mv data data_firedrake
+          coverage combine
+          mv .coverage coverage_firedrake.dat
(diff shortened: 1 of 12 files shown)


pySDC/implementations/datatype_classes/firedrake_mesh.py
```diff
@@ -0,0 +1,114 @@
...
+
+        Args:
+            other (float): factor
+        Raises:
+            DataError: if other is not a float
+        Returns:
+            fenics_mesh: copy of original values scaled by factor
+        """
+
+        if type(other) in [int, float, complex]:
```
tlunet, line 63: Why don't you simply a try catch here ?

```python
try:
    me = firedrake_mesh(self)
    me.functionspace.assign(other * self.functionspace)
    return me
except:
    raise DataError("Type error: cannot multiply %s to %s" % (type(other), type(self)))
```

That way it would allows anything that `self.functionspace` allows, and be less pySDC dependent

tlunet: PS : that could also be done for the other operations (was implicit in the comment :sweat_smile:)

brownbaerchen: The other operations actually work only if both are in the pySDC datatype. And I want to keep it that way because otherwise you might leave the pySDC datatype unknowingly and everything stops working.

.github/workflows/ci_pipeline.yml
```diff
@@ -169,6 +169,58 @@ jobs:
...
+        - ${{ github.workspace }}:/repositories
+    defaults:
+      run:
+        shell: bash -l {0}
+    steps:
+      - name: Checkout pySDC
+        uses: actions/checkout@v4
+        with: 
+          path: ./pySDC
+      - name: Checkout gusto
```
pancetta, line 188: Do you need gusto in this PR?

pySDC/tests/test_tutorials/test_step_7.py
```diff
@@ -127,3 +127,27 @@ def test_D():
...
+
+
+@pytest.mark.firedrake
+def test_E():
+    from pySDC.tutorial.step_7.E_pySDC_with_Firedrake import runHeatFiredrake
+
+    runHeatFiredrake(useMPIsweeper=False)
+
+
+@pytest.mark.firedrake
```
pancetta, line 139: Don't you need to mark this as parallel test here?

pancetta:
Prettttttty awesome!

pySDC/tests/test_tutorials/test_step_7.py
brownbaerchen: Firedrake comes with MPI, so the environment necessarily has everything for parallel runs.

.github/workflows/ci_pipeline.yml
brownbaerchen: We don't need this here, but the pipeline was a bit of a mess to setup because I am checking out two repositories and then installing them. I thought there is not much of a point in untangling this and then reverting this for the follow-up Gusto PR. Unless you would want two separate test environments for Gusto and Firedrake? But I can remove Gusto here anyways if you want.

pancetta: Nope, all good.

pancetta:
Further thoughts, @jshipton or @tlunet ?

tlunet:
Looks fine by me. Having a first structure for a coupling with firedrake and gusto is already some nice input, eventually we can improve that later ...

brownbaerchen:
I would appreciate if we could merge this sooner rather than later. I cleaned up the Gusto coupling a lot today. I basically have to make sure the parameters in the tutorial are sensible and then I could open a PR for this from tomorrow.

PR #514, opened 2025-01-21
Fix for missing numpy data types on some platforms

brownbaerchen:
I had some trouble with round-off errors today and noticed that my numpy doesn't actually have quadruple precision. So `np.float128` will raise an error on my machine. `np.longdouble`, on the other hand, is just regular double on my machine. This is extremely weird behaviour that I neither like nor understand. I switched out the quadruple precision things for the word version so I can run the IO stuff with double precision at least.
Notice that the [numpy documentation](https://numpy.org/doc/2.1/reference/arrays.scalars.html#numpy.longdouble) says that `np.longdouble` is an alias for `np.float128` on Linux x86_64 machines. So this fix applies to Mac and Windows users.

### DIFF
--- a/pySDC/helpers/fieldsIO.py
+++ b/pySDC/helpers/fieldsIO.py
@@ -49,6 +49,7 @@ wether code is run in parallel or not.
 import os
 import numpy as np
 from typing import Type, TypeVar
+import logging
 
 T = TypeVar("T")
 
@@ -78,8 +79,6 @@ try:
         }
     )
 except AttributeError:
-    import logging
-
     logging.getLogger('FieldsIO').debug('Warning: Quadruple precision not available on this machine')
 try:
     DTYPES.update(
@@ -89,8 +88,6 @@ try:
         }
     )
 except AttributeError:
-    import logging
-
     logging.getLogger('FieldsIO').debug('Warning: Single precision not available on this machine')
 
 DTYPES_AVAIL = {val: key for key, val in DTYPES.items()}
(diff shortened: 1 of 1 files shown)


tlunet:
Interesting ... I'm always a bit confused on how numpy handle datatypes. In particular there is this thing :

```python
import numpy as np
u = np.zeros(10, dtype=np.float64)
u.dtype == np.float64  # => True
u.dtype is np.float64  # => False
```

The way data types are handled can probably still be improved, however I think replacing `np.float128` by `np.longdouble` will make the `FieldsIO` very platform dependent, and may lead to inconsistent file sizes (in particular if on some platform, `np.longdouble` are simply standart double precision numbers ...). The main advantage of keeping this size-based notation is that at least, it's explicit on how much bytes are used to store a field value.

I would rather see some import dependent definition, like this :

```python
DTYPES = {
    0: np.float64,  # double precision
    1: np.complex128,
}
try:
    DTYPES.update({
        2: np.float128,  # quadruple precision
        3: np.complex256,
    })
except AttributeError:
    # Some warning on double precision not available on this platform
# And the same for single precision ...
```

brownbaerchen:
You're right, that is a better solution! I went with the warning only on debug level and added a bit more elaborate message when you try to access a datatype that is not there.

pySDC/helpers/fieldsIO.py
```diff
@@ -69,11 +69,30 @@ class MPI:
...
 }
+try:
+    DTYPES.update(
+        {
+            2: np.float128,  # quadruple precision
+            3: np.complex256,
+        }
+    )
+except AttributeError:
+    import logging
```
tlunet, line 81: Can logging not be a module import ? Local import with black formating makes this very ugly ...

tlunet:
Looks all good, thanks !

pancetta:
Excellent solution, thanks!

PR #516, opened 2025-01-23
Added coupling between pySDC and Gusto

brownbaerchen:
This enables to use pySDC as a time discretisation method within Gusto, or to setup a geophysical problem in Gusto and then run it in pySDC.
The intended use is to take a script running a Gusto problem, setup the Gusto timestepper with the pySDC integrator and then either just use that in normal Gusto fashion, or extract the controller from the pySDC integrator and run in pySDC.

### Added in this PR
 - helper that inherits from the Gusto base time discretisation and sets up a pySDC controller that can then be used on it's own or called from Gusto during the time stepping 
 - generic Gusto problem class that accepts a Gusto equation or a Gusto residual and assembles nonlinear variational problems to be solved for evaluating the right hand side or doing an Euler step
 - IMEX version of generic gusto problem class
 - Tutorial on running Williamson 5 (flow over mountain) using pySDC within Gusto, includes plotting script taken almost unmodified from the Gusto repository. Script includes infrastructure for MPI parallel diagonal SDC.
 - Appended the tutorial README to include this and the Firedrake coupling.

### Tests
 - Test for using the generic Gusto problem class without upwinding and without the accompanying setup in the timestepper
 - Test that various RK methods and (IMEX) SDC give the same result between the Gusto and pySDC implementations
 - Test for adaptive step size selection within the coupling: First the problem is run adaptively using pySDC, then the same sequence of steps is solved in Gusto.
 - Test for the tutorial that after a few steps the results between Gusto and pySDC SDC implementations match

### Caveats
 - Gusto is not compatible with the most recent version of Firedrake. While it is being fixed, we have to limit the CI pipeline to use the last compatible version of Firedrake at the moment. Please let us know when we can revert to the most current version, @jshipton.
 - Gusto is not currently built for adaptive step size selection and the current implementation for this in the coupling is very hacky.

### DIFF
--- a/pySDC/helpers/pySDC_as_gusto_time_discretization.py
+++ b/pySDC/helpers/pySDC_as_gusto_time_discretization.py
@@ -43,11 +43,8 @@ class pySDC_integrator(TimeDiscretisation):
         controller_params,
         domain,
         field_name=None,
-        subcycling_options=None,
         solver_parameters=None,
-        limiter=None,
         options=None,
-        augmentation=None,
         t0=0,
         imex=False,
     ):
@@ -62,20 +59,12 @@ class pySDC_integrator(TimeDiscretisation):
                 mesh and the compatible function spaces.
             field_name (str, optional): name of the field to be evolved.
                 Defaults to None.
-            subcycling_options(:class:`SubcyclingOptions`, optional): an object
-                containing options for subcycling the time discretisation.
-                Defaults to None.
             solver_parameters (dict, optional): dictionary of parameters to
                 pass to the underlying solver. Defaults to None.
-            limiter (:class:`Limiter` object, optional): a limiter to apply to
-                the evolving field to enforce monotonicity. Defaults to None.
             options (:class:`AdvectionOptions`, optional): an object containing
                 options to either be passed to the spatial discretisation, or
                 to control the "wrapper" methods, such as Embedded DG or a
                 recovery method. Defaults to None.
-            augmentation (:class:`Augmentation`): allows the equation solved in
-                this time discretisation to be augmented, for instances with
-                extra terms of another auxiliary variable. Defaults to None.
         """
 
         self._residual = None
@@ -83,11 +72,8 @@ class pySDC_integrator(TimeDiscretisation):
         super().__init__(
             domain=domain,
             field_name=field_name,
-            subcycling_options=subcycling_options,
             solver_parameters=solver_parameters,
-            limiter=limiter,
             options=options,
-            augmentation=augmentation,
         )
 
         self.description = description
--- a/pySDC/helpers/pySDC_as_gusto_time_discretization.py
+++ b/pySDC/helpers/pySDC_as_gusto_time_discretization.py
@@ -1,0 +1,188 @@
+import firedrake as fd
+
+from gusto.time_discretisation.time_discretisation import TimeDiscretisation, wrapper_apply
+from gusto.core.labels import explicit
+
+from pySDC.implementations.controller_classes.controller_nonMPI import controller_nonMPI
+from pySDC.implementations.problem_classes.GenericGusto import GenericGusto, GenericGustoImex
+from pySDC.core.hooks import Hooks
+from pySDC.helpers.stats_helper import get_sorted
+
+
+class LogTime(Hooks):
+    """
+    Utility hook for knowing how far we got when using adaptive step size selection.
+    """
+
+    def post_step(self, step, level_number):
+        L = step.levels[level_number]
+        self.add_to_stats(
+            process=step.status.slot,
+            process_sweeper=L.sweep.rank,
+            time=L.time,
+            level=-1,
+            iter=-1,
+            sweep=-1,
+            type='_time',
+            value=L.time + L.dt,
+        )
+
+
+class pySDC_integrator(TimeDiscretisation):
+    """
+    This class can be entered into Gusto as a time discretization scheme and will solve steps using pySDC.
+    It will construct a pySDC controller which can be used by itself and will be used within the time step when called
+    from Gusto. Access the controller via `pySDC_integrator.controller`. This class also has `pySDC_integrator.stats`,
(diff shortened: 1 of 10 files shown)


pySDC/helpers/pySDC_as_gusto_time_discretization.py
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(
+        self,
+        equation,
+        description,
+        controller_params,
+        domain,
+        field_name=None,
+        subcycling_options=None,
+        solver_parameters=None,
+        limiter=None,
```
jshipton, line 48: I think that not all of these options work from pySDC, e.g. the limiter... probably the sub cycling options aren't used and probably not the augmentation either... would it be better to remove them from the interface and docs?

brownbaerchen: You're right! I removed those three parameters. I am not sure about the `options`. Seems to me like it might work when using Gusto for time integration because it adds a wrapper, but I don't actually know.

pySDC/implementations/problem_classes/GenericGusto.py
```diff
@@ -0,0 +1,264 @@
...
+    ):
+        """
+        Initialisation
+
+        Args:
+            equation (:class:`PrognosticEquation`): the model's equation.
+            apply_bcs (bool, optional): whether to apply the equation's boundary
+                conditions. Defaults to True.
+            solver_params (dict, optional): Solver parameters for the nonlinear variational problems
+            stop_at_divergence (bool): Whether to raise an error when the variational problems do not converge
```
jshipton, line 46:
```suggestion
            stop_at_divergence (bool, optional): Whether to raise an error when the variational problems do not converge. Defaults to False
```

pySDC/implementations/problem_classes/GenericGusto.py
```diff
@@ -0,0 +1,264 @@
...
+            LHS_cache_size (int): Size of the cache for solvers
```
jshipton, line 47:
```suggestion
            LHS_cache_size (int, optional): Size of the cache for solvers. Defaults to 12.
```

pySDC/implementations/problem_classes/GenericGusto.py
```diff
@@ -0,0 +1,264 @@
...
+            residual (Firedrake.form, optional): Overwrite the residual of the equation, e.g. after adding spatial methods.
```
jshipton, line 48:
```suggestion
            residual (Firedrake.form, optional): Overwrite the residual of the equation, e.g. after adding spatial methods. Defaults to None.
```

pySDC/implementations/problem_classes/GenericGusto.py
pancetta: Whoohoo, magic numbers! 😄

pySDC/implementations/problem_classes/GenericGusto.py
```diff
@@ -0,0 +1,264 @@
...
+        self.work_counters['ksp'].niter += self.solvers[factor].snes.getLinearSolveIterations()
+        self.work_counters['solver']()
+        return self.dtype_u(self.x_out)
+
+
+class GenericGustoImex(GenericGusto):
+    dtype_f = IMEX_firedrake_mesh
+    rhs_n_labels = 2
+
+    def evaluate_individual_term(self, u, label):
```
jshipton, line 186: slightly picky... but `label` might apply to more than one term so not sure "individual" is the right word

pySDC/implementations/problem_classes/GenericGusto.py
brownbaerchen: Well, 12 is a highly composite number and you ideally want something that is divisible by the number of collocation nodes and definitely something that is more than the number of collocation nodes. So it's not as bad as the preconditioners :D

pySDC/tutorial/step_7/F_pySDC_with_Gusto.py
```diff
@@ -0,0 +1,346 @@
+"""
+Example for running pySDC together with Gusto. This test runs a shallow water equation and may take a considerable
+amount of time. After you have run it, move on to step F_2, which includes a plotting script.
+
+This is Test Case 5 (flow over a mountain) of Williamson et al, 1992:
+``A standard test set for numerical approximations to the shallow water
+equations in spherical geometry'', JCP.
+
+This script is adapted from the Gusto examples.
```
jshipton, line 9:
```suggestion
This script is adapted from the Gusto example: https://github.com/firedrakeproject/gusto/blob/main/examples/shallow_water/williamson_5.py
```

pySDC/tutorial/step_7/F_2_plot_pySDC_with_Gusto_result.py
```diff
@@ -0,0 +1,205 @@
+def plot_williamson_5(
```
jshipton, line 1: I'm slightly concerned that this plotting script could end up getting out of date - but I'm not sure what we do about that! The ones in the Gusto case studies repository will be updated - can you link to that version or were there too many changes?

jshipton: I'm not exactly sure what I'm concerned about! Something may change somewhere...

brownbaerchen: If you think the dependencies of the plotting script are more likely to change than the script running the Williamson 5 test, then I can remove the plotting from pySDC and only put in the link to the Gusto one. 🤷

jshipton: Yeah, I see what you mean... maybe leave it as it is and see if it becomes a problem?

jshipton (approved):
This looks like a very thorough job to me @brownbaerchen ! I've checked all the Gusto residual manipulation and agree with what you've done, and it's all very comprehensively tested. Great work!

PR #525, opened 2025-01-29
Bumped to Stage2025 for HPC

jakob-fritz:
By this, also changed version of FFTW (as the previous one seems to  not be included in Stage 2025)


brownbaerchen:
I don't understand what's going on here. I was able to install mpi4py-fft in a fresh venv after loading the same modules as in this PR and then running
```
FFTW_LIBRARY_DIR="/p/software/juwels/stages/2025/software/FFTW/3.3.10-GCC-13.3.0/lib64" pip install --no-deps --force-reinstall -e .
```
However, when using the same command here, it complains that it does not find NumPy. Seems to me that the Python versions are getting mixed up somewhere. After loading the `Scipy-Stack` module, NumPy should be installed, which indeed it is because when installing pySDC, it is already there. Anybody knows what's happening?

tlunet:
> I don't understand what's going on here. I was able to install mpi4py-fft in a fresh venv after loading the same modules as in this PR and then running
> 
> ```
> FFTW_LIBRARY_DIR="/p/software/juwels/stages/2025/software/FFTW/3.3.10-GCC-13.3.0/lib64" pip install --no-deps --force-reinstall -e .
> ```
> 
> However, when using the same command here, it complains that it does not find NumPy. Seems to me that the Python versions are getting mixed up somewhere. After loading the `Scipy-Stack` module, NumPy should be installed, which indeed it is because when installing pySDC, it is already there. Anybody knows what's happening?

No idea, but I got similar problem before when loading Scipy-Stack. I believe there is another module that has numpy in it. But I would also strongly recommend not using any python install from JSC, but rather recompile python from source, create an environment from it, and install everything in there using pip.

pancetta:
The strange thing is that this started to fail with the new stage. All of this worked before, except for this weird `venv` problem (which we got rid off).

jakob-fritz:
I can have a brief look at it now. If I don't find anything, you can open a ticket at SC support.

jakob-fritz:
I could not really reproduce the issue, when using a venv in my own home directory (the shared one was not accessible to me, as it is owned by Robert). Therefore, I did some slight changes and let's see if it now works as expected.

brownbaerchen:
> I could not really reproduce the issue, when using a venv in my own home directory (the shared one was not accessible to me, as it is owned by Robert). Therefore, I did some slight changes and let's see if it now works as expected.

Same, I also found no issue when following the steps to install everything in a venv in my home directory.. So it looks like a combination of the new stages and the Gitlab pipeline that causes this issue..

brownbaerchen:
> Ok I cannot see, why it does not work with a venv on SCRATCH via Jacamar, but does work when you or I test it manually with a venv in the own home-directory. So, @brownbaerchen if you want to, go ahead and open a ticket, as I did not find where the approaches differ (apart from the location of the venv)

Well, thanks for trying anyways! I'll open a ticket next week. Really weird bug..

pancetta:
Any update here?

brownbaerchen:
No, haven't found the time to open the ticket yet.

brownbaerchen:
Well... I wanted to merge master before opening a ticket. But seems the pipeline is broken even for 2024 stage now. It complains now that the `module` command is not found on the juwels login node.. see e.g. [here](https://github.com/Parallel-in-Time/pySDC/actions/runs/13346193865/job/37276923662#step:5:144). What is going on with our pipeline these days? Why is there a new problem every day? :D

jakob-fritz:
I just tried to rerun the pipeline and now got a different error (No module named 'numpy' 🤦), which I don't unterstand. In any case, the general thing seems to work, as the last job on HPC passed about (an hour ago)[https://gitlab.jsc.fz-juelich.de/atml-ati/pysdc/-/pipelines/252614]. It was a scheduled one with the latest master. So it was not influenced by the changes of this MR, but anyway: It ran on the HPC-Systems 🎉

brownbaerchen:
Since the projects are kind of piling up right now, I will deal with this when I get back end of March. We can leave the PR open so we don't forget about this.

jakob-fritz:
As I updated my master-branch, I closed this PR by accident. The new branch has a new PR: #544

PR #527, opened 2025-02-04
Minor CI fixes

brownbaerchen:
The features from Gusto that we use here work again with the most recent Firedrake, so I switched the pipeline to use the latest Firedrake docker container. (@jshipton)
Also, black had an update and required some reformatting...

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -173,7 +173,7 @@ jobs:
   user_firedrake_tests:
     runs-on: ubuntu-latest
     container:
-      image: firedrakeproject/firedrake-vanilla:2025-01
+      image: firedrakeproject/firedrake-vanilla:latest
       options: --user root
       volumes:
         - ${{ github.workspace }}:/repositories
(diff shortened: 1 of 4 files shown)


.github/workflows/ci_pipeline.yml
```diff
@@ -173,7 +173,7 @@ jobs:
   user_firedrake_tests:
     runs-on: ubuntu-latest
     container:
-      image: firedrakeproject/firedrake-vanilla:2025-01
+      image: firedrakeproject/firedrake-vanilla:latest
```
pancetta, line 176: That is dangerous. You don't know what the next stage will bring.. can you pin the version to something?

brownbaerchen: un-pinning the version is precisely the point of this PR. The Gusto pipeline runs on the latest Firedrake, so if we pin Firedrake but not Gusto, we might get into trouble. If we pin both, we make sure that the current state runs now, but people may not be able to install the current Firedrake outside of containers in the future.
I suggest we test on the latest Firedrake and revert to a specific version if something fails that we cannot fix easily.
What is your take @jshipton, you have more experience on being dependent on Firedrake ;)

jshipton: I agree with @brownbaerchen - that seems to be the sanest option

PR #528, opened 2025-02-12
Firedrake ensemble fix

brownbaerchen:
After [some changes to Gusto](https://github.com/firedrakeproject/gusto/commit/81dedbde34e5e599239a329d50178e5b5e6a0ebb), some [strange deadlocks have started to appear in the pySDC tests](https://github.com/Parallel-in-Time/pySDC/actions/runs/13234132362/job/37084519438).

In the Gusto changes, a debug output showing the max/min values of some fields was added, which I presume requires communication across space. Removing this debug information is sufficient to get these tests to pass again.

In order to make sure the space-parallelism is set up correctly, I changed the MPI tests to always use four tasks and distribute them to do space-only, time-only, or space-time parallelism. All of these tests pass without the debug output.

This is definitely some fishy business. Do you know if the debug output works in Gusto without pySDC, @jshipton @tommbendall? If I add `return None` after [this line](https://github.com/firedrakeproject/gusto/blob/3737472c3c36cbf65b71f591dc812614016c5ac7/gusto/timestepping/timestepper.py#L170), the pySDC tests pass with debug level output in Gusto, so I am pretty confident it is related to the remainder of this function.

Btw, I had some issues in freeing communicators at the end of the script on my laptop, so I added the `Free` function to the Firedrake ensemble communicator wrapper. This should not make a large difference in practice.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -205,7 +205,7 @@ jobs:
           firedrake-clean
           cd ./pySDC
           coverage run -m pytest --continue-on-collection-errors -v --durations=0 /repositories/pySDC/pySDC/tests -m firedrake
-        timeout-minutes: 120
+        timeout-minutes: 45
       - name: Make coverage report
         run: |
           . /home/firedrake/firedrake/bin/activate
--- a/pySDC/helpers/firedrake_ensemble_communicator.py
+++ b/pySDC/helpers/firedrake_ensemble_communicator.py
@@ -67,6 +67,9 @@ class FiredrakeEnsembleCommunicator:
             return self.ensemble.ensemble_comm.Isend(buf=buf, dest=dest, tag=tag)
         return self.ensemble.isend(buf, dest, tag=tag)[0]
 
+    def Free(self):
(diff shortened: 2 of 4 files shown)


tommbendall:
Oh scratch what I said about pytest and logging -- I see that's exactly what you've done to get your CI passing.

brownbaerchen:
> I think the tests are hanging because the `field_data.min()` and `field_data.max()` calls that have been added involve parallel communication. Maybe these don't work in the presence of an Ensemble communicator?

I also suspect that this is causing the issue. However, the ensemble communicators just generate a set of space communicators, which should be useable within Gusto and which "don't know" that they are generated from the ensemble, afaik. I am not saying that this is a bug in Gusto, could be the way I use it. But it is definitely a bug somewhere.

> 
> Our test suite does include some parallel tests and there hasn't been an issue here.

Confusingly, the tests with debug level pass with one or two tasks, just not with four, making it even more fishy..


I guess for now we can just ignore this random bug where we don't even know if it's in pySDC, Gusto or Firedrake.. Just thought you should be aware of more weirdness :D

Issue #534, opened 2025-02-20
Add a generci hook class to output data using the `FieldsIO` helpers + demo + documentation.


tlunet:
Following discussion in #533. Here is a first TODO list :

- [ ] Description of the base native data IO structure in the documentation
- [ ] Hook class to use the FieldsID
    - [ ] base serial implementation
    - [ ] MPI setup
    - [ ] demo and documentation

tlunet:
But then, one critical question : if we make the `FieldsIO` classes the base IO feature of pySDC, shouldn't we make them compatible with python 3.9 and 3.10 ?

tlunet:
Adaptation to python 3.9 and 3.10 + full collective IO done with #538

PR #538, opened 2025-03-31
Performance fix for FieldsIO

brownbaerchen:
Turns out that collective IO is much faster than individual operations when using FieldsIO within pySDC runs on HPC. This PR changes from individual to collective IO if possible and raises a warning if not. To be specific, you can generate a decomposition with uneven amount of data on the tasks, which prevents collective IO and probably leads to load balancing issues in other places as well, but you will get a warning telling you to change resolution / number of tasks.
There are a few other little things in this PR such as adding some missing pytest markers.

### DIFF
--- a/pySDC/helpers/fieldsIO.py
+++ b/pySDC/helpers/fieldsIO.py
@@ -45,7 +45,7 @@ See :class:`pySDC.helpers.fieldsIO.writeFields_MPI` for an illustrative example.
 Warning
 -------
 To use MPI collective writing, you need to call first the class methods :class:`Rectilinear.initMPI` (cf their docstring).
-Also, `Rectilinear.setHeader` **must be given the global grids coordinates**, wether the code is run in parallel or not.
+Also, `Rectilinear.setHeader` **must be given the global grids coordinates**, whether the code is run in parallel or not.
 
 > ⚠️ Also : this module can only be imported with **Python 3.11 or higher** !
 """
@@ -495,13 +495,13 @@ class Rectilinear(Scalar):
     # MPI-parallel implementation
     # -------------------------------------------------------------------------
     comm: MPI.Intracomm = None
-    balanced_distribution = None
+    _num_collective_IO = None
 
     @classmethod
     def setupMPI(cls, comm: MPI.Intracomm, iLoc, nLoc):
         """
         Setup the MPI mode for the files IO, considering a decomposition
-        of the 1D grid into contiuous subintervals.
+        of the 1D grid into contiguous subintervals.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


tlunet:
Also, maybe we can make already the small modifications to make it work for python <= 3.10 ...

brownbaerchen:
I implemented the requests by @tlunet in the last commit:
 - With unbalanced distributions as much as possible of the IO is now using the fast collective operations, no more warning about unbalanced distributions
 - Changed the unpacking syntax to be compatible with older python versions, which now also test FieldsIO

pySDC/helpers/fieldsIO.py
```diff
@@ -624,9 +651,11 @@ def addField(self, time, field):
             self.MPI_WRITE(np.array(time, dtype=T_DTYPE))
         offset0 += self.tSize
 
+        _num_writes = 0
         for (iVar, *iBeg) in itertools.product(range(self.nVar), *[range(n) for n in self.nLoc[:-1]]):
```
tlunet, line 655: That's not exactly what I suggested ... all IO operation must be collective with _all ! It's just that we have to ensure that all processes are calling exactly the same number of time MPI_WRITE_AT or MPI_READ_AT. It means that you have to add after the first for loop :

```python
for _ in range(self.num_collective_IO - _num_writes):
    self.MPI_WRITE_AT(0, field[:0])
```
That should avoid the deadlock

pySDC/helpers/fieldsIO.py
```diff
@@ -515,6 +517,21 @@ def setupMPI(cls, comm: MPI.Intracomm, iLoc, nLoc):
         cls.nLoc = nLoc
         cls.mpiFile = None
 
+    @property
```
tlunet, line 520: Probably better to call the allreduce in `setupMPI` and store `numCollectiveIO` in one attribute at this point

pySDC/helpers/fieldsIO.py
brownbaerchen: Why is that better? I expect little performance hit by non-collective IO here because there are few non-collective operations relative to total operations.

pySDC/helpers/fieldsIO.py
brownbaerchen: Well, if you want that, then you need to pass the instance or the number of variables in `setupMPI`. I went this way in order not to change the interface.

pySDC/helpers/fieldsIO.py
tlunet: Still counts ... and you can have a majority of non-collective operation with some decomposition. For instance, a 2D problem of size [10,32] decomposed in the first direction in 6 sub-domains => 4 processes have 2 points, 2 have 1 point. Because you need to write contiguous chunk of data in the file, for the first loop all are doing collective write for the first points, but for the second points 4 processes are doing a non-collective operation.

It's just better to always have all process doing collective write, with those who don't write anything doing a no-op, and the MPI implementation is hopefully well done enough to avoid losing anything on that.

pySDC/helpers/fieldsIO.py
tlunet: Oh yes, I forgot that ... then put `cls._num_collective_IO` to None in `setupMPI`, to avoid any bug if two different decomposition are used one after the other

pySDC/helpers/fieldsIO.py
```diff
@@ -54,6 +54,7 @@
 from typing import Type, TypeVar
 import logging
 import itertools
+import warnings
```
tlunet, line 57: Unused now, right ?

pySDC/helpers/fieldsIO.py
brownbaerchen: Actually, I found some inconsistent behaviour when reading and writing nothing on some tasks on my laptop. Namely, if I print something, it goes through, but if I don't, it deadlocks. I don't understand this behaviour. I suggest we leave this with some non-collective IO operations and if you want, you can fix it when you return from vacation.

tlunet: I found were the problem was : `self.num_collective_IO` must be called at the beginning of `readField` or `addField`. Because the first time the property is evaluated, it calls a `allreduce` but then the process having the maximum number of collective IO are still launching a `write_at_all` or a `read_at_all`, which creates the deadlock.

Now, the `allreduce` is done before any other collective operation, the tests did pass on my laptop. Let's see what happen with the github tests ...

tlunet:
Seems like all the relevant tests passed. Feel free to merge it when you want, I'll update #534

Issue #539, opened 2025-04-01
Streamline MPI tests


brownbaerchen:
So far, we have done MPI-parallel testing in a really cumbersome way: pytest would run tests in serial, which launch a parallel execution using `subprocess`. This is slow, messy to write and gives useless error messages.
There are MPI extensions for pytest. None of them are perfect, but some of them are significantly better than what we do now. I suggest we add one of them to the pipeline and transition to using them in the future, although I am not 100% which is the best at this time.

# The Firedrake solution
The Firedrake people develop [mpi-pytest](https://github.com/firedrakeproject/mpi-pytest). This works very simply:
```
@pytest.mark.parallel(nprocs=[1, 2, 3])  # run in parallel on 1, 2 and 3 processes
def test_my_code_on_variable_nprocs():
    ...
```
The downside is that it doesn't use the Python `assert` statement, but a separate function. On the plus side, this communicates the result of the assertion between tasks, but it requires a different usage from normal. Normal assert is still available, but may lead to deadlocks.

# The DLR solution
The DLR people develop [pytest-isolate-mpi](https://pytest-isolate-mpi.readthedocs.io/en/stable/index.html). This seems a bit more mature, but also more cumbersome. A test is written with
```
@pytest.mark.mpi(ranks=[1, 2, 3])
def test_number_of_processes_matches_ranks(mpi_ranks, comm):
    ...
```
Note the `mpi_ranks` argument, which every function now has to have. `comm` is a pytest fixture here that is optional as an argument, I believe. I am not actually sure how deadlocks following asserts are handled here. There is no special assertion pytest-isolate-mpi. But there is easy handling of timeouts to deal with deadlocks and the reports contain information about on which tasks the test has failed.

# Choosing an option
Feel free to add more options or opinions in the comments. I am not clear on all pros and cons and I am not fully decided which I prefer. Both of the above options are pip installable and neither requires exotic dependencies.

pancetta:
Fixed by #694 and #695

PR #542, opened 2025-04-03
Implemented interface to FieldsIO in core problem class and respective hook

brownbaerchen:
I added three functions to the core problem class, which I hope cover most cases:
 - `setUpFieldsIO`: Mostly for setting up space decomposition. Here, you can call all the class methods of `FieldsIO` and derived classes
 - `getOutputFile`: This function is supposed to instantiate a `FieldsIO` object fully set up, only taking a file name as an argument. This needs to be done individually for the problem classes to set the header and data structure correctly
 - `processSolutionForOutput`: Do stuff to the solution before writing to the file. For instance, when using spectral methods, you may want to transform from spectral space to real space and throw away the imaginary part to save disc space.

Once these functions are set up for the problem class you want, you can simply add the hook `LogToFile` (I renamed the previous hook of this name to `LogToPickleFile`).
You configure this class via class attributes for file name, simulation time increment between writes and whether or not you allow overwriting of previous files.
This hook also supports restarting runs. If the path to the file exists and you are starting from a time greater than zero, it will append the file with the new solutions.

I added the setup of `FieldsIO` for `generic_spectral` and `TestEquation0D` classes. I have used this to run the 3D RBC examples that I will merge at some point.

Finally, I wrote a bunch of tests for this. Note that I am using `pytest-isolate-mpi` as discussed in #539 here.

This PR works on #534. It is addressing functionality parts rather than documentation, which should be added later as well.

### DIFF
--- a/pySDC/tests/test_hooks/test_log_to_file.py
+++ b/pySDC/tests/test_hooks/test_log_to_file.py
@@ -1,12 +1,22 @@
 import pytest
 
 
-def run(hook, Tend=0):
-    from pySDC.implementations.problem_classes.TestEquation_0D import testequation0d
-    from pySDC.implementations.sweeper_classes.generic_implicit import generic_implicit
+def run(hook, Tend=0, ODE=True, t0=0):
     from pySDC.implementations.controller_classes.controller_nonMPI import controller_nonMPI
+    from pySDC.helpers.fieldsIO import FieldsIO
 
-    level_params = {'dt': 1.0e-1}
+    if ODE:
+        from pySDC.implementations.problem_classes.TestEquation_0D import testequation0d as problem_class
+        from pySDC.implementations.sweeper_classes.generic_implicit import generic_implicit as sweeper_class
+
+        problem_params = {'u0': 1.0}
+    else:
+        from pySDC.implementations.problem_classes.RayleighBenard import RayleighBenard as problem_class
+        from pySDC.implementations.sweeper_classes.imex_1st_order import imex_1st_order as sweeper_class
+
+        problem_params = {'nx': 16, 'nz': 8, 'spectral_space': False}
+
+    level_params = {'dt': 1.0e-2}
 
     sweeper_params = {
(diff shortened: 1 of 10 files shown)


brownbaerchen:
Does anyone have any opinions on this? :)

pySDC/tests/test_hooks/test_log_to_file.py
```diff
@@ -15,71 +25,145 @@ def run(hook, Tend=0):
...
+    assert np.allclose([me[0] for me in u_restart], [me[0] for me in u_continuous]), 'Times don\'t match'
+    for u1, u2 in zip(u_restart, u_continuous):
+        assert np.allclose(u1[1], u2[1]), 'solution does not match'
 
 
-if __name__ == '__main__':
-    test_logging()
+@pytest.mark.mpi4py
+@pytest.mark.mpi(ranks=[1, 4])
+def test_loggingMPI(tmpdir, comm, mpi_ranks):
```
tlunet, line 167: Why is `mpi_ranks` not used ?

tlunet: Oh, I imagine that's for the mpi test, right ? Maybe adding a bit of doc about it in the contribution guide could be good ...

brownbaerchen: Indeed, all three arguments for this function are fixtures. `mpi_ranks` is mandatory for this version of MPI compatible pytest, see #539. If you remove it, the tests will fail and tell you to put it back in. But I can add a small comment that this is needed for the test. Although we need to take care not to copy paste this everywhere in case we don't need it anymore..
I do want to highlight how useful the `tmpdir` fixture is, though. It's perhaps not often that you need it. But if you do, it's quite handy to ensure you get a path that didn't exist before and that will not exist after the tests.

tlunet:
I'm fine with merging it. especially since @brownbaerchen started to use more the camelCase convention 😍

PR #548, opened 2025-04-24
Fix for website

brownbaerchen:
You really shouldn't let me do releases 🙈 Sorry about that!


pancetta:
Didn't catch that either.. OMG

tlunet:
What, `vim` didn't show you that :hushed: ?

PR #549, opened 2025-04-28
Make plot helper compatible with Python 3.13

brownbaerchen:
Python `distutils` has been [deprecated since version 3.10 and removed in version 3.12](https://docs.python.org/3/library/distutils.html). In 3.12 I could still use it but after switching to 3.13, the plot helper was broken for me. 
Apparently,`distutil` has been replaced by `shutil`. `shutil.which` returns what you get when you use `which` in the shell you run Python in, or `None` if there is no executable of the name you are looking for. Therefore, this should give the same functionality as `distutil.spawn.find_executable` as it is used in the plot helper.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -204,6 +204,7 @@ jobs:
           . venv-pySDC/bin/activate
           pip install -e /repositories/pySDC
           pip install qmat
+          pip install pytest-isolate-mpi
           # test installation
           python -c "import pySDC; print(f'pySDC module: {pySDC}')"
       - name: Install gusto
--- a/pySDC/tests/test_benchmarks/test_collocation.py
+++ b/pySDC/tests/test_benchmarks/test_collocation.py
@@ -25,6 +25,7 @@ def test_benchmark_collocation(benchmark):
     benchmark(wrapper)
 
 
+@pytest.mark.base
 @pytest.mark.parametrize("node_type", node_types)
 @pytest.mark.parametrize("quad_type", quad_types)
 def test_canintegratepolynomials(node_type, quad_type):
@@ -61,6 +62,7 @@ def test_canintegratepolynomials(node_type, quad_type):
         )
 
 
+@pytest.mark.base
 @pytest.mark.parametrize("node_type", node_types)
 @pytest.mark.parametrize("quad_type", quad_types)
 def test_relateQandSmat(node_type, quad_type):
@@ -82,6 +84,7 @@ def test_relateQandSmat(node_type, quad_type):
             )
 
 
+@pytest.mark.base
 @pytest.mark.parametrize("node_type", node_types)
 @pytest.mark.parametrize("quad_type", quad_types)
 def test_partialquadraturewithQ(node_type, quad_type):
@@ -104,6 +107,7 @@ def test_partialquadraturewithQ(node_type, quad_type):
             )
 
 
+@pytest.mark.base
 @pytest.mark.parametrize("node_type", node_types)
 @pytest.mark.parametrize("quad_type", quad_types)
 def test_partialquadraturewithS(node_type, quad_type):
--- a/pySDC/helpers/plot_helper.py
+++ b/pySDC/helpers/plot_helper.py
@@ -1,6 +1,6 @@
 import matplotlib as mpl
 import matplotlib.pyplot as plt
-from distutils.spawn import find_executable
+import shutil
 
 default_mpl_params = mpl.rcParams.copy()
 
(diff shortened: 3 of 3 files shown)


pancetta:
Any updates here?

brownbaerchen:
This action also fails on the master branch of my fork. Not sure why it doesn't here. I am pretty sure that it is unrelated to the changes in this PR, but I'll try and change the way the test is launched later and see if it helps.

pySDC/tests/test_benchmarks/test_collocation.py
```diff
@@ -25,6 +25,7 @@ def wrapper():
     benchmark(wrapper)
 
 
+@pytest.mark.base
```
brownbaerchen, line 28: @tlunet, do you agree with this markers? I don't think we need to run these tests in all environments. They are throwing some numpy deprecation warnings at the moment, which we will need to deal with someday.

.github/workflows/ci_pipeline.yml
```diff
@@ -204,6 +204,7 @@ jobs:
           . venv-pySDC/bin/activate
           pip install -e /repositories/pySDC
           pip install qmat
+          pip install pytest-isolate-mpi
```
tlunet, line 207: Should be added as a test dependency, directly in the pyproject.toml file, I beleive (see https://github.com/Parallel-in-Time/qmat/blob/4919f197bac230a27f8ccea4188645cb53fefd0a/pyproject.toml#L34) ... also, qmat should already be installed when installing pySDC

pySDC/tests/test_benchmarks/test_collocation.py
tlunet: They are necessary. You just need to replace in the assertion the == between arrays by some np.allclose (as the warning is suggesting ?)

tlunet: That's a polynomial deprecation warrning ?

brownbaerchen: I didn't pay much attention to the warning, tbh. For now, I just want to run these tests only in the base environment and not in every environment.

.github/workflows/ci_pipeline.yml
tlunet: Hence the idea of optional environment defined in the `pyproject.toml` file. You could create a `gusto` one or a `gusto-test`, etc ... and simply install using :

```bash
pip install -e ./repositories/pySDC[gusto]
```

which would install the base dependencies, + the optional ones (cf [here](https://github.com/Parallel-in-Time/qmat/blob/4919f197bac230a27f8ccea4188645cb53fefd0a/.github/workflows/ci_pipeline.yml#L25)). It makes the pipeline way cleaner, and centralize the dependencies in the `pyproject.toml`, which allows a better control for local or CI tests ...

pySDC/tests/test_benchmarks/test_collocation.py
tlunet: Oh, then that's fine indeed ... sorry I didn't realized that was an addition :sweat_smile:

.github/workflows/ci_pipeline.yml
brownbaerchen: We can think of adding MPI dependencies there, but I don't really see the point, tbh. I prefer this way where the setup of the container and what goes into the container is all together. Outside of the testing pipeline people are not going to use the container and will not need the testing thing.

Also, turns out the test is flaky somehow and adding pytest-isolate-mpi did not at all fix it. This is pretty annoying :D

tlunet: Then, my guess is that you can remove this and the `pip install qmat` :stuck_out_tongue:

tlunet:
Is the firedrake test the problematic one, of the Allen_Cahn ones ?

brownbaerchen:
> Is the firedrake test the problematic one, of the Allen_Cahn ones ?

The Allen_Cahn ones are just a one time whatever problem, I suppose. This one Firedrake test keeps failing.

brownbaerchen:
> Does firedrake imports mpi4py or some MPI binding already on his side ? Got some problem with fenix once, where importing mpi4py before dolfin was producing an error in tests ...

There are other tests using firedrake and mpi that run just fine. The failing one uses Firedrake ensemble communicators for space-time parallelism.

brownbaerchen:
> Updates here?

I have not managed to fix the firedrake tests yet. This plotting stuff is not related, but I will try to fix the firedrake stuff, maybe this week, hopefully by the end of next week. No need to merge this before.

PR #550, opened 2025-05-09
FieldsIO performance fix

tlunet:
Remove the for loop with `mpi_write_at_all` to use only one write with a sub-array view. In particular, it avoids the need of the nCollectiveIO stuff, and makes the code way simpler.

> Always beneficial to know MPI features to avoid re-implementing those in a more clumsy way :sweat_smile:


tlunet:
Why are the firedrake test randomly failing ?

PR #551, opened 2025-05-16
TL: mini fix for edge cases

tlunet:
There is a rare situation when the `MPI_SETUP` method (setup the subarray view for MPI_READ_ALL) is not called when using MPI : when the `Rectilinear` object is instantiated first, and the MPI communicator and local slices are given by `setupMPI` after. This requires to read the header again, or else it raises and error.

While this is not the recommended usage for `Rectilinear` (`setupMPI` should be called before instantiating any class), this modification allows it while not changing the overall behavior or performance of the class.


tlunet:
Firedrake tests failing as usual, but this can be merged when you wish

PR #553, opened 2025-06-10
Add Python 3.13 to core test pipeline

brownbaerchen:
According to the [status of Python versions](https://devguide.python.org/versions/), Python 3.13 is supported at least to some extend. When I last installed a fresh pySDC environment without explicitly specifying a Python version, I even ended up with 3.13. Therefore, I suggest we better test it to accommodate future users.
However, the pipeline is already too large as is, with some project tests taking a very long time. Therefore, I did not add 3.13 to those tests. In fact, I would be open to stop doing the project tests with all supported Python versions. IMHO, it would suffice to use the latest version (only 3.13 in this case) or to allow projects to specify a supported Python version.

### DIFF
--- a/.github/workflows/ci_pipeline.yml
+++ b/.github/workflows/ci_pipeline.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           environment-file: ${{ env.YML }}
           create-args: >-
-              python=3.12
+              python=3.13
       - name: Code reformatting with black
         run: |
           black pySDC  --check --diff --color
@@ -70,7 +70,7 @@ jobs:
           mv .coverage coverage_${{ matrix.env }}.dat
       - name: Uploading artifacts
         uses: actions/upload-artifact@v4
-        if: matrix.python == '3.12'
+        if: matrix.python == '3.13'
         with:
           name: test-artifacts-cpu-${{ matrix.env }}
           path: |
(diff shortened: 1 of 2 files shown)


tlunet:
+1 on the project tests for one Python version only.

Also, using the latest version for project test should simplify maintaining on the long term IMO => once a year, we upgrade the latest version, and eventually correct the project that need adaptation. Just need a proper clarification on when we upgrade every year.

pancetta:
Fine by me. Projects should have a way to specify their (set of) Python versions that should be tested. pySDC's core should be tested against as much as we can, though.

pancetta:
You mean like, `3.8`? If this project cannot be updated to support a newer version and if porting will be to tedious, we have to move it to `deprecated`. Not sure what else we can do?

pancetta:
But then, what's the harm? if the core of pySDC still works with that version? The more the better, no?

brownbaerchen:
It's straightforward to set the Python version individually by simply adding ` - python=3.13` to the environment files.

However, why not treat the Python version like any other dependency? If we don't specify a version, micromamba will select the latest reasonable version. Then we either support the latest that users get, or we react once the pipeline fails for some reason and restrict to an older version.

In any case, I suggest we don't remove 3.9 from the core tests because it is still supported.

pancetta:
Fine by me. Wild, but fine.

brownbaerchen:
Let's get this merged!

PR #554, opened 2025-06-17
Added caching decorator

brownbaerchen:
This is one feature that I added during the large refactor and one that can be merged independently. It is not particularly useful as used in this PR, but is used extensively in the refactored version.

The caching decorator returns a cached result if the function arguments match a previous call and can be used like this:
```
num_calls = 0

@cache
def increment(x):
    num_calls += 1
    return x + 1

increment(0)  # returns 1, num_calls = 1
increment(1)  # returns 2, num_calls = 2
increment(0)  # returns 1, num_calls = 2
```

### DIFF
--- a/pySDC/helpers/spectral_helper.py
+++ b/pySDC/helpers/spectral_helper.py
@@ -8,6 +8,8 @@ from functools import wraps
 def cache(func):
     """
     Decorator for caching return values of functions.
+    This is very similar to `functools.cache`, but without the memory leaks (see
+    https://docs.astral.sh/ruff/rules/cached-instance-method/).
 
     Example:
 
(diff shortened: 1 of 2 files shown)


pancetta:
Did those leaks show up for you? Or why did you start implementing this on your own in the first place?

brownbaerchen:
> Did those leaks show up for you? Or why did you start implementing this on your own in the first place?

Ruff complains about this. The [link](https://docs.astral.sh/ruff/rules/cached-instance-method/) leads to the ruff documentation of things they check against. So then I told ChatGPT to write sth, as one does, and it was quickly implemented.

I added the link to the ruff documentation about why I don't just use `functools.cache` to the documentation and added a test for this decorator.

PR #555, opened 2025-06-17
Added test against memory leaks with cache decorator

brownbaerchen:
Slightly more elaborate version of the test suggested by @tlunet in #554.


tlunet:
Neat !

PR #556, opened 2025-06-24
Switch to temporary paths in FieldsIO tests

brownbaerchen:
By use of the pytest `tmpdir` fixture (10/10 would recommend), the data created by the tests is automatically cleaned up afterwords.


pancetta:
This is not impacting the website built? This uses the figures created by `pytest`.

PR #557, opened 2025-06-25
Added workflow for triggering the Codecov bot in PRs

brownbaerchen:
I do miss the Codecov bot posting about coverage in our PRs. I want to bring it back, but I don't know a lot about it, so maybe this will not actually do the trick.. I cannot test it before since this workflow will only be triggered once it is in the master branch.

What I added here is a separate workflow that runs when the GitHub CI is done and during PRs. All it does in that case is upload the coverage report to Codecov, which I assume will then trigger the bot.

Problem: Coverage will almost certainly be lower than in the master branch because the GPU tests are not included in this report. This should not lead to fluctuations in the coverage report, though, because the report is generated for a non-master branch and therefore does not show up in the tracking.

I suggest we bring back the bot first and when we managed to do that, we configure it to not fail when coverage has decreased. It will still show which parts of the PR are tested and which aren't, which is very useful information.

### DIFF
--- a/.github/workflows/trigger_coverage_bot.yml
+++ b/.github/workflows/trigger_coverage_bot.yml
@@ -1,0 +1,48 @@
+---
+
+name: Trigger coverage bot
+
+on:
+  workflow_run:
+    workflows: ["CI pipeline for pySDC"]
+    types: ["completed"]
+  
+jobs:
+
+  Upload to Codecov:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


brownbaerchen:
What do you mean? Clicking on the codecov badge in the main README will lead you to the [report for the commit from 34 minutes ago](https://app.codecov.io/gh/Parallel-in-Time/pySDC/commit/3044940fb7028cc443a3953a5eb377cdefd28507)

jakob-fritz:
Ah ok. It lead me to the report from a week ago, when I tried it just before I wrote my comment. But maybe it was just an issue I had.

PR #561, opened 2025-06-30
Updated spectral helper to new scipy version

brownbaerchen:
There was a version change in scipy which broke some tests. This PR fixes the tests in the base and mpi4py environments. 
However, there is still one test failing in the `parallelSDC_reloaded` project. Do you have a minute to fix that one, @tlunet? If you could make it faster while you are at it, that would also be great.. ;)

### DIFF
--- a/pySDC/helpers/spectral_helper.py
+++ b/pySDC/helpers/spectral_helper.py
@@ -292,7 +292,7 @@ class ChebychevHelper(SpectralHelper1D):
 
         def get_forward_conv(name):
             if name == 'T2U':
-                mat = (sp.eye(N) - sp.diags(xp.ones(N - 2), offsets=+2)) / 2.0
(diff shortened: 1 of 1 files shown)


tlunet:
To solve the test problem, simply change this part of the reference data here : https://github.com/Parallel-in-Time/pySDC/blob/ed6cdecd6b38e54809bed38624cddda8a19a51ca/pySDC/projects/parallelSDC_reloaded/scripts/_dataRef.json#L22

to :

```json
"fig05_allenCahn_conv_0_VDHS": [Infinity,Infinity,
        0.5345708075891601,
        0.20140479812575454,
        0.05620974256627712,
        0.008404588766557962,
        0.0016989846252961774,
        0.00016807940141687374,
        0.0001875379340551794
    ],
```

This is just some tolerance issue for a non-converging case.

PR #564, opened 2025-07-03
Tutorial for Edinburgh

brownbaerchen:
These notebooks explain what pySDC is about, how to add a problem, and how to couple to other libraries.
They contain a lot of explanations, so people can go through them by themselves in principle, but the plan is for me to do talking with nobody else running anything during the tutorials. At least not from the tutorials..
It may be too much and I may skip through parts of the second notebook since the focus is on the coupling. But I guess it's better to leave this here and skip in the moment.

Feel free to complain about the examples, missing stuff, irrelevant stuff, ...


brownbaerchen:
The monodomain test is broken for some odd reason which is unrelated to this PR..
Do you have time to fix it @grosilho? Otherwise, I will see what I can do at some point..

pancetta:
Ready to go?

brownbaerchen:
Yes. Would be good to have this in the main repository before Monday so I can tell people where to find it.

PR #565, opened 2025-07-04
Reduced memory footprint of spectral discretizations via block-diagonal sparse operators

brownbaerchen:
The spectral discretizations require a few block-diagonal matrices among all degrees of freedom. Constructing them as such actually reduces the memory requirements quite a bit compared to constructing general block matrices with only the diagonal blocks.
This is independent of the larger refactor, which is why I did a separate PR.
I added a test that the resulting operators are the same and that indeed memory is reduced.


pancetta:
How large are these matrices in general?

PR #568, opened 2025-07-11
Fix for spatial preconditioners in generic spectral problem class

brownbaerchen:
This PR has a few small things:
 - There was a bug in the Dirichlet recombination (right preconditioner), which was never done
 - Initial guesses are now correctly treated for the right preconditioner
 - A few small tests including that the system matrices are actually banded once preconditioned


tlunet:
Looks good to me, ok to merge as soon as the test pass

PR #576, opened 2025-07-21
Use sc_venv_template for tests on JUWELS

brownbaerchen:
second try.. This time without git submodules.

### DIFF
--- a/etc/venv_juwels/activate.sh
+++ b/etc/venv_juwels/activate.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# See https://stackoverflow.com/a/28336473
+SOURCE_PATH="${BASH_SOURCE[0]:-${(%):-%x}}"
+
+RELATIVE_PATH="$(dirname "$SOURCE_PATH")"
+ABSOLUTE_PATH="$(realpath "${RELATIVE_PATH}")"
+
+[[ "$0" != "${SOURCE_PATH}" ]] && echo "The activation script must be sourced, otherwise the virtual environment will not work." || ( echo "Vars script must be sourced." && exit 1) ;
+
+source "${ABSOLUTE_PATH}"/config.sh
+source "${ABSOLUTE_PATH}"/modules.sh
+
+export PYTHONPATH="$(echo "${ENV_DIR}"/lib/python*/site-packages):${PYTHONPATH}"
+
+source "${ENV_DIR}"/bin/activate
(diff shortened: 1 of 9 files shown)


etc/venv_juwels/setup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+SOURCE_PATH="${BASH_SOURCE[0]:-${(%):-%x}}"
```
jakob-fritz, line 3: Maybe adding a brief explanation (e.g. "Getting the current path of this file") is helpful, as the line is not self-explanatory to bash-newbies ;-)

etc/venv_juwels/activate.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# See https://stackoverflow.com/a/28336473
```
jakob-fritz, line 3: That's a good link! Maybe add it to the other occurrences as well and additionally add a brief note here on what this does (to my understanding storing the path of the current file in the var `SOURCE_PATH`)

etc/venv_juwels/config.sh
```diff
@@ -0,0 +1,14 @@
+SOURCE_PATH="${BASH_SOURCE[0]:-${(%):-%x}}"
```
jakob-fritz, line 1: See other occurrences of this line

pySDC/tests/test_helpers/test_spectral_helper.py
```diff
@@ -519,7 +519,7 @@ def test_tau_method(bc, N, bc_val, kind='Dirichlet', useGPU=False):
 
     sol_poly = np.polynomial.Chebyshev(sol_hat)
     d_sol_poly = sol_poly.deriv(1)
-    x = xp.linspace(-1, 1, 100)
+    x = np.linspace(-1, 1, 100)
```
jakob-fritz, line 522: I know too little context to judge if pinning this to numpy instead of the more general xp (which also points to np, by default if I read [spectral_helper.py line 913](https://github.com/Parallel-in-Time/pySDC/blob/ed741a690a848d55dc0adb79f319c67df9e09ef1/pySDC/helpers/spectral_helper.py#L913C5-L913C12) correctly) makes sense of not. But I assume you know what you are doing ;-)

jakob-fritz (commented):
I like the restructuring. However, I found one line (in multiple occasions) not really easy to understand and would have appreciated a comment on what it does.
This is marked.

Furthermore, I don't know the reason for the change in `test_spectral_helper.py` (also marked with a comment), but assume you know what you did there and why you did it.

etc/venv_juwels/setup.sh
brownbaerchen: I fully agree. This is part of the template and I myself have just not questioned it :D

pySDC/tests/test_helpers/test_spectral_helper.py
brownbaerchen: Cupy doesn't have the polynomial stuff that I use in numpy, so in this test I copy everything back to CPU after the computation and then do the comparison on CPU.
I have no clue why this test was passing before with `xp`. But `np` is definitely the way to go.

PR #577, opened 2025-07-22
Sweepers require level during instantiation

### LINKED ISSUE #46: Can instantiate sweeper with problem
At the moment, it is possible to instantiate an object of type `imex_1st_order` by providing only the parameters `collocation_class` and `num_nodes`.

However, when calling e.g. the `integrate` function, an error is thrown because no problem or level is attached. This should be caught during instantiation, but I am not sure what is the proper way to that.

brownbaerchen:
As @danielru has correctly noted 10 years ago in #46, the sweepers can be instantiated, but not used without a level.
Since I now have some capacity for tending the garden, I thought I'd change that. 
I also think it makes it easier get started with developing a new sweeper since its now more obvious that sweeper and level interact.

Note that it is possible to pass `None` as the level in the SDC sweepers. The sweepers have setters for the level that make sure you pass a proper level, but they are not used during instantiation. In some tests, sweepers are instantiated to check the attached collocation problem, so we should keep this.

Closes #46

### DIFF
--- a/pySDC/core/level.py
+++ b/pySDC/core/level.py
@@ -67,13 +67,13 @@ class Level(FrozenClass):
             level_index (int): custom name for this level
         """
 
-        # instantiate sweeper, problem and hooks
-        self.__sweep = sweeper_class(sweeper_params, self)
-        self.__prob = problem_class(**problem_params)
-
         # set level parameters and status
         self.params = _Pars(level_params)
         self.status = _Status()
+
+        # instantiate sweeper, problem and hooks
+        self.__sweep = sweeper_class(sweeper_params, self)
+        self.__prob = problem_class(**problem_params)
 
         # set name
         self.level_index = level_index
--- a/pySDC/core/level.py
+++ b/pySDC/core/level.py
@@ -68,7 +68,7 @@ class Level(FrozenClass):
         """
 
         # instantiate sweeper, problem and hooks
-        self.__sweep = sweeper_class(sweeper_params)
+        self.__sweep = sweeper_class(sweeper_params, self)
         self.__prob = problem_class(**problem_params)
 
         # set level parameters and status
@@ -86,9 +86,6 @@ class Level(FrozenClass):
         self.fold = [None] * (self.sweep.coll.num_nodes + 1)
 
         self.tau = [None] * self.sweep.coll.num_nodes
-
-        # pass this level to the sweeper for easy access
-        self.sweep.level = self
 
         self.__tag = None
 
--- a/pySDC/projects/parallelSDC_reloaded/nilpotency.py
+++ b/pySDC/projects/parallelSDC_reloaded/nilpotency.py
@@ -3,7 +3,7 @@
 """
 Created on Fri Dec 22 17:17:14 2023
 
-Evaluate the nilpotency of diagonal preconditionners MIN-SR-S and MIN-SR-NS
+Evaluate the nilpotency of diagonal preconditioners MIN-SR-S and MIN-SR-NS
 with increasing number of nodes.
 """
 import numpy as np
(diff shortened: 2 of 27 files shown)


pySDC/tests/test_sweepers/test_preconditioners.py
tlunet, line 1: Those tests are exactly those done in `qmat`, is it still necessary to keep them here ?

pySDC/projects/parallelSDC_reloaded/nilpotency.py
```diff
@@ -36,7 +36,7 @@ def nilpotencyNS(d, Q):
 nil_MIN_SR_NS = []
 nNodes = range(2, 20)
 for m in nNodes:
-    s = Sweeper({"num_nodes": m, "quad_type": quadType, "node_type": nodeType})
+    s = Sweeper({"num_nodes": m, "quad_type": quadType, "node_type": nodeType}, None)
```
tlunet, line 39: Wait ... why can't the base `Sweeper` class have `None` as default value for the `level` argument ?

brownbaerchen: Well, the point of this PR is precisely to force passing of a level during instantiation of the sweeper.

Currently, sweepers are only instantiated without a level in a couple of places to setup the corresponding collocation problem.
Outside of testing a sweeper, you either want to have a level with your sweeper to do timestepping or you want to use qmat to generate the collocation problem. I don't see a situation where instantiating a sweeper without a level is useful outside of testing that the sweeper sets up the collocation problem correctly.

tlunet: Why couldn't we use the Sweeper without any level to simply do some SDC sweeps ? If the level is just the object storing the all solution values at each nodes, then it could be generated by default if the user don't specify it.

To fully speak my mind, I don't like adding positional constructor parameters in general (without default value) when they aren't necessary to define the behavior of the object. It's not very user friendly, and in this case this level parameter seems more related to ML stuff than the sweeper.

tlunet: Ok thanks ! It's clearer now (should have read daniel's issue first, I guess ... sorry :sweat_smile:).

But I'm still puzzled on this : 
- `Level` needs a `Problem` right, but need also to know the number of nodes from the sweeper (e.g for the RK sweeper)
- `Sweeper` will be instantiated with the number of nodes, but also need `Level` object

Feels like a circular dependency, right ? Doesn't help the user side ...

pySDC/core/level.py
```diff
@@ -67,14 +67,14 @@ def __init__(self, problem_class, problem_params, sweeper_class, sweeper_params,
...
-        # instantiate sweeper, problem and hooks
-        self.__sweep = sweeper_class(sweeper_params)
-        self.__prob = problem_class(**problem_params)
-
         # set level parameters and status
         self.params = _Pars(level_params)
         self.status = _Status()
 
+        # instantiate sweeper, problem and hooks
+        self.__sweep = sweeper_class(sweeper_params, self)
```
tlunet, line 75: Yep ... this is confusing :sweat_smile:

pySDC/projects/parallelSDC_reloaded/nilpotency.py
brownbaerchen: Yes, I agree that this is a bit confusing. However, the user does not typically interact with the level. I have never seen any but the core level class being used. So the idea is that users now know more explicitly that the sweeper relies on the level but need not concern themselves with how the level class is implemented.

tlunet: Ok, then let's keep it that way for the PR, it's fine enough. But we'll have to find a better solution, eventually ...

pancetta:
So, are we going to keep it, merge it, or throw it away?

PR #580, opened 2025-08-06
Allow other name for FLEX preconditioner

brownbaerchen:
Thibaut and I noticed earlier that in `qmat` you can use underscores or dashes in the names of preconditioners, but in `pySDC` only dashes lead to update of the preconditioner coefficients between iterations. Can lead to some difficult to find bugs because there is no indication that something has gone wrong.

This is a simple fix so that `MIN-SR-FLEX` and `MIN_SR_FLEX` now both work the same.

### DIFF
--- a/pySDC/core/sweeper.py
+++ b/pySDC/core/sweeper.py
@@ -270,5 +270,5 @@ class Sweeper(object):
         k : int
             Index of the sweep (0 for initial sweep, 1 for the first one, ...).
         """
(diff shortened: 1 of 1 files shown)


tlunet:
Suggestion : instead of having only one `generator` attribute, put instead a `genQI` and `genQE` attribute, that are used separately for the implicit and explicit QDelta matrices. And the `setupGenerator` must be quickly adapted to allow that

tlunet:
Yes :smile:

PR #581, opened 2025-08-08
Added spectrum computation to 3D RBC

brownbaerchen:
In turbulent flow simulations people often show a sort of energy spectrum to demonstrate that the spatial resolution is adequate. The main feature of this PR is a method for computing precisely this spectrum. @tlunet, please check if you find this reasonable.
Also, I fixed a bug in the Nusselt number computation. The formula employed there previously was valid for the special case of Ra=1, which is a case nobody cares about.
Now I can generate plots like this for verification:
Figure_1

# Some background on the spectrum

People are often worried about aliasing. As an example, consider the grid $x= 0, \pi/2, \pi, 3\pi / 2$. Evaluating the wave $\exp(i 2 x)$ on this grid, gives $1, -1, 1, -1$. Now consider $\exp(i6 \pi)$. Evaluating this on the grid gives the same values $1, -1, 1, -1$ because of the periodicity of the grid. The wave with wavenumber 6 is aliased to the wave with wavenumber 2 on this grid with only four modes. 
The fastest wavenumber supported on this grid is 3. This is easy to see because wavenumber 4 gets aliased to wavenumber 0 as both waves are always evaluated as constant 1.

Aliasing is not much of a problem as long as the wavenumbers greater than supported by the resolution never appear. However, they do tend to appear in nonlinear problems. In the above example, the fastest supported wave is wavenumber 3 and we just saw that it's square, which has wavenumber 6 is aliased to wavenumber 2.
When we are dealing with convection, we have terms such as $u\partial_x u$, which introduce waves of twice the speed, which may get aliased.

Now spectral methods are famous for exponential decay of coefficients with wavenumber. This is only true for special cases, but as a rule of thumb, coefficients decay quickly if the problem is well resolved.
If indeed coefficients decay quickly enough, the aliasing will not matter so much because the spurious contributions in low wavenumbers due to aliasing are small compared to the magnitude of the non-spurious slow waves.
This is why turbulence people show these spectrum plots: A lot of energy in the fastest supported wave, often called "2h wave" is associated with instability due to a lot of aliasing, but if there is little energy in the 2h wave, it's fine.

Actually, we get rid of the instability caused by aliasing by doing dealiasing.
The idea is simple: Get a larger grid that supports the waves so they don't get aliased. But larger grid is also more expensive and you may run into a vicious circle of increasing the resolution. So you only increase the size of the grid in physical space.
Before the inverse Fourier transform from frequency space to grid space, you padd the solution with zeros in wave numbers that would be aliased. Then you compute the square of the waves on the grid, compute the Fourier transform to go back to frequency space and throw away all the waves of wave numbers higher than the original resolution.
Orzag figured out that you need to padd by 50% to get rid of aliasing.

So aliasing is not really a problem for us, but we should still show these plots so reviewers are happy and if the coefficients are large that is still a sign of poor resolution.
Don't mind all the above text. Thanks for rubber ducking!

### DIFF
--- a/pySDC/implementations/problem_classes/RayleighBenard3D.py
+++ b/pySDC/implementations/problem_classes/RayleighBenard3D.py
@@ -28,9 +28,9 @@ class RayleighBenard3D(GenericSpectralLinear):
 
     The domain, vertical boundary conditions and pressure gauge are
 
-        Omega = [0, 8) x (-1, 1)
+        Omega = [0, Lx) x [0, Ly] x (0, Lz)
         T(z=+1) = 0
-        T(z=-1) = 2
+        T(z=-1) = Lz
         u(z=+-1) = v(z=+-1) = 0
         integral over p = 0
 
@@ -53,16 +53,16 @@ class RayleighBenard3D(GenericSpectralLinear):
     def __init__(
         self,
         Prandtl=1,
-        Rayleigh=2e6,
-        nx=256,
-        ny=256,
-        nz=64,
+        Rayleigh=1e6,
+        nx=64,
+        ny=64,
+        nz=32,
         BCs=None,
         dealiasing=1.5,
         comm=None,
         Lz=1,
-        Lx=1,
-        Ly=1,
+        Lx=4,
+        Ly=4,
         useGPU=False,
         **kwargs,
     ):
@@ -79,7 +79,6 @@ class RayleighBenard3D(GenericSpectralLinear):
             comm (mpi4py.Intracomm): Space communicator
             Lx (float): Horizontal length of the domain
         """
-        # TODO: documentation
         BCs = {} if BCs is None else BCs
         BCs = {
             'T_top': 0,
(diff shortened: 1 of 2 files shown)


pySDC/implementations/problem_classes/RayleighBenard3D.py
```diff
@@ -354,3 +353,55 @@ def compute_Nusselt_numbers(self, u):
             't': Nusselt_t,
             'b': Nusselt_b,
         }
+
+    def get_frequency_spectrum(self, u):
```
tlunet, line 357: Could you add just a bit of documentation here ? One can understand from the function how it works, but it would be better to allow that directly from the docstring

tlunet:
Seems fine, although I think that the spectrum computation should be put in a separate helper module. I do understand why you need to link it to the problem to have the transform available, but then a separate function should be able to compute it from the "pseudo-spectral" 3D field. Would allow to use the same computation for other problems or even other codes (like Dedalus with SDC, etc ...)

tlunet:
Looks good, thanks !

PR #584, opened 2025-08-20
More genericity for QDelta coefficients

tlunet:
This PR makes use of the new update in `qmat`, that allows to generate any QDelta coefficients using always the same `qGen` argument, without any required code modification for new QDelta types. 

It also re-activates the caching of the `genQE` and `genQI` attributes, which do not need to be re-built again when variable coefficients are updated between sweeps : it uses the internal cache of `qmat` that can store different QDelta matrices if they are different. Furthermore, it allows variable coefficients for explicit sweeps.

Finally, coefficient update are now done in the controller classes, more precisely in the `it_fine` method, rather than in the `update_node` of the sweeper class. It means that there is no need to modify the `update_node` of each sweeper implementation to allow the use of FLEX preconditionners (for instance). In case some implementation are still calling the `updateVariableCoeffs` method, it simply becomes a no-op.

In addition, a few cleaning / refactoring / type hints that appears to be necessary during the changes mentionned above.

### DIFF
--- a/pySDC/core/sweeper.py
+++ b/pySDC/core/sweeper.py
@@ -1,11 +1,17 @@
 import logging
 import numpy as np
-from qmat import QDELTA_GENERATORS
+from qmat.qdelta import QDeltaGenerator, QDELTA_GENERATORS
 
 from pySDC.core.errors import ParameterError
 from pySDC.core.level import Level
 from pySDC.core.collocation import CollBase
 from pySDC.helpers.pysdc_helper import FrozenClass
+
+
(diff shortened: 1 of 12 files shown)


pySDC/core/sweeper.py
```diff
@@ -251,6 +247,8 @@ def level(self, L):
         Args:
             L (pySDC.Level.level): current level
         """
+        from pySDC.core.level import Level
```
pancetta, line 250: Why is this here and not on top?

tlunet: Because it creates a circular dependency that produce an error on import

PR #589, opened 2025-10-21
Fix linter

brownbaerchen:
Ruff now checks for [B905](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/), i.e. it wants to use `strict=True` in `zip`. I guess I understand why this would be a sensible thing to do, but we violate this in so many places in the code that I prefer to just not check for it.
If you disagree, I can go through all instances of `zip` and add the keyword, of course. Please checkout the above link to the ruff website before deciding.


pancetta:
Uh.. sounds like a good thing to have `strict=true`, doesn't it?

PR #591, opened 2025-10-22
Weights for integrating across entire interval in spectral methods

brownbaerchen:
There are very simple relationships for the integrals across the entire intervals in some spectral methods. For instance, in Fourier methods, only the constant mode has nonzero integral since the integrals over the entire domain of sine and cosine modes vanish. For Chebychev methods, the relationship is less obvious but similarly easy to implement. 

This PR includes tests with randomly generated solutions. For Chebychev methods, a reference value is computed using qmat in physical space. For Fourier methods, the reference value is simply derived from the coefficient of the constant mode.


tlunet:
Neat !

PR #652, opened 2026-06-10
Fixes for running spectral discretizations on GPU

brownbaerchen:
Some issues came up when writing data to disk on GPUs. Previously, I had only run benchmarks without any writing of data. This PR addresses these issues.
Also, the vkfft backend is not working anymore for some reason. I simply reverted to cuFFT for the time being.

### DIFF
--- a/pySDC/implementations/problem_classes/generic_spectral.py
+++ b/pySDC/implementations/problem_classes/generic_spectral.py
@@ -429,6 +429,9 @@ class GenericSpectralLinear(Problem):
         self.setUpFieldsIO()
 
         coords = [me.get_1dgrid() for me in self.spectral.axes]
+        if self.spectral.useGPU:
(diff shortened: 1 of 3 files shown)


pySDC/implementations/problem_classes/generic_spectral.py
```diff
@@ -438,9 +441,14 @@ def getOutputFile(self, fileName):
...
-            return np.array(self.itransform(u).real)
+            u = self.itransform(u).real
         else:
-            return np.array(u.real)
+            u = u.real
+
+        if self.spectral.useGPU:
+            u = u.get()
+
+        return np.array(u)
```
tlunet: I the copy intentional here ? If not, maybe use `np.asarray`, and if yes maybe use the `copy` method from the array instead.

PR #653, opened 2026-06-11
Fix for non-contiguous arrays

brownbaerchen:
Turns out skipping the copy in #652 can lead to non-contiguous arrays that need to be cast to contiguous before writing. If the array is already contiguous, this is zero-copy.


tlunet:
Was that the reason of the "unstable RK443" ? :sweat_smile:

brownbaerchen:
> Was that the reason of the "unstable RK443" ? 😅

No without this the code just crashes on CPU with more than one task. Luckily I found the issue with RK443 and fixed it in #654... So sadly RK443 is not unconditionally unstable after all :D

PR #654, opened 2026-06-11
Fix IMEX RK on GPU

brownbaerchen:
Ok, sometimes I see why people are frustrated with Python. This deep-copy / shallow copy bug that made RK443 blow up on GPUs was quite hard to find...


pancetta:
Good catch. These bugs suck, big time.

tlunet:
Classic : programming in python looks easy, but properly doing it is not :sweat_smile:

pancetta:
I'm glad we now have AI for th.. oh, wait..

PR #679, opened 2026-09-07
Let the iteration estimator run with one step per rank

pancetta:
Last of three. #677 added the conformance suite and marked this asymmetry, #678 removed the broken half, and this closes it: the iteration estimator now works in both transports and stops the block at the same iteration either way.

## What was wrong

The estimator existed twice and worked once.

- `controller_MPI` had an inline copy that was never switched on anywhere, had no tests, and deadlocked when used. Removed in #678.
- The `CheckIterationEstimatorNonMPI` convergence controller worked, and tutorial step_8 C exercises it — but only with the whole block in one process. It took its running maximum from a buffer on itself, accumulated as the block was walked, which is a thing a rank cannot do: an object here is shared by a whole block in one case and private to a rank in the other.

## What it looks like now

One implementation, `CheckIterationEstimator`. The only quantity that is not step-local is the largest change anywhere in the block:

| transport | reduction |
|---|---|
| one step per rank | `MPI_Allreduce` with `MPI_MAX` |
| block in one process | maximum over the steps |

Everything else — `diff_local`, `diff_old`, `diff_first` — moved onto the steps via `add_status_variable_to_step`, where it belongs and where both transports reach it.

## The part that took three goes, because it is not obvious

My first version used `MPI_Scan`, a prefix maximum, which is the faithful translation of the old code. The estimator's inputs then matched **exactly** — I instrumented both sides and `diff_local` and `diff_old` agreed to every digit on every iteration — and it still stopped at 8 iterations against 6.

Checking convergence with the block in one process is a pass over the steps in order. A verdict reached at the last step arrives too late for the steps already visited in that pass, so the block runs one more iteration than the same problem on one step per rank: **a collective reaches every rank at once, a loop does not.** No amount of matching the arithmetic fixes that; the shape of the reduction has to change.

Taking the maximum over the whole block instead means every step holds the same number, reaches the same verdict in the same pass, and nothing has to be propagated between steps. It is taken on the *first* step visited, which is also the only moment it can be taken: `StoreUOld` overwrites `uold` step by step as the block is walked, so by the last step the earlier ones can no longer say how far they moved.

(An intermediate attempt reduced `force_done` inside `CheckConvergence` instead. It got to 7 against 6 and timed `test_step_8::test_A` out, because it ran `all(T.status.done ...)` inside the per-step loop where the not-yet-visited steps still hold last iteration's values. The controller already does that reduction correctly, afterwards. Reverted — and with the whole-block maximum it is not needed at all.)

## Requirement

Under MPI the estimator stops the whole block at once, so every rank has to take part in every iteration for the maximum to be defined. `check_parameters` asks for `all_to_done = True` with an explanation rather than assuming it. Nothing changes for the single-process case.

## Compatibility

`CheckIterationEstimatorNonMPI` stays as an alias. Tutorial step_8 C passes with its expected iteration counts unchanged, so no documented numbers move.

## Testing

**263 passed** across `test_controllers`, `test_convergence_controllers`, `test_tutorials` and `tests_core` (base + mpi4py); `ruff` and `black` clean.

Two new axes in the conformance suite, both folded into its existing single `mpirun` launch:

- `test_iteration_estimator_agrees` — the estimator stops at the same iteration and reaches the same answer in both transports. This is the parity the previous two PRs did not have, and it compares iteration counts rather than only the answer, since stopping early on an estimate is exactly where the two could disagree with nothing looking wrong.
- `test_iteration_estimator_actually_stops_early` — because an estimator that never fires would make both transports agree trivially.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pySDC/implementations/convergence_controller_classes/check_iteration_estimator.py
+++ b/pySDC/implementations/convergence_controller_classes/check_iteration_estimator.py
@@ -1,26 +1,58 @@
 import numpy as np
-from pySDC.core.convergence_controller import ConvergenceController, Status
+
+from pySDC.core.convergence_controller import ConvergenceController
 from pySDC.implementations.convergence_controller_classes.store_uold import StoreUOld
 
 
-class CheckIterationEstimatorNonMPI(ConvergenceController):
-    def __init__(self, controller, params, description, **kwargs):
+class CheckIterationEstimator(ConvergenceController):
+    r"""
+    Stop iterating once an estimate says the desired tolerance has been reached.
+
+    The estimate is the standard contraction argument. With :math:`d_k` the largest change between
+    the last two iterates,
+
+    .. math::
+        \tilde{L} = \min\left(\frac{d_k}{d_{k-1}}, 0.9\right), \quad
+        \alpha = \frac{d_1}{1 - \tilde{L}}, \quad
+        K = 1.05 \frac{\log(\text{errtol} / \alpha)}{\log \tilde{L}},
+
+    and once :math:`\lceil K \rceil` is no larger than the current iteration the block stops.
+
+    :math:`d_k` is the largest change anywhere in the block, so every step reaches the same verdict at
+    the same iteration. That maximum is the only thing here that is not step-local, and it is all the
+    two transports do differently: an ``MPI_Allreduce`` with ``MPI_MAX`` when there is one step per
+    rank, and a maximum over the steps when the block is in one process.
+
+    Taking it over the whole block rather than over the steps up to the one asking is what makes the
+    two agree. Checking convergence is a pass over the steps in order, so a verdict reached at the
+    last step arrives too late for the steps already visited, and the block would run an extra
+    iteration that the same problem on one step per rank does not -- a collective reaches every rank
+    at once, a loop does not. The maximum is therefore taken on the first step visited, which is also
+    the only moment it can be taken: `StoreUOld` overwrites ``uold`` step by step as the block is
+    walked, so by the last step the earlier ones can no longer say how far they moved.
+
+    Everything else lives on the steps rather than on this object, which is what lets one class serve
+    both transports -- an object here is shared by a whole block in one case and private to a rank in
+    the other.
+    """
+
+    def setup(self, controller, params, description, **kwargs):
         """
-        Initialization routine
+        Define default parameters here.
 
         Args:
             controller (pySDC.Controller): The controller
             params (dict): Parameters for the convergence controller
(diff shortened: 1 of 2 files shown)


pancetta:
Closing unmerged: this breaks pipelining, which is not a trade worth making for this feature.

## Why

The version pushed here requires `all_to_done = True` under MPI. That is synchronous global convergence — converged ranks keep iterating until the slowest finishes — so it removes the drain that PFASST and MSSDC exist to provide. Shipping an estimator that only works with the pipelining switched off is a footgun regardless of how the docstring is worded.

## What the right design looks like, for whoever picks this up

The quantity the estimator needs is a **prefix** maximum, not a global one: error propagates forward in time, so step `k` needs the largest change over steps `0..k` and nothing after it. That is mathematically motivated *and* forward-only, which is exactly the pipelined data flow — early steps see less, estimate less, and stop sooner. Each step then decides for itself and no verdict has to travel backwards through the block, which is also what makes the two transports agree without synchronising.

I implemented that. It passes parity and a pipelining guard with the block in one process, and **deadlocks under MPI** for a structural reason worth writing down:

```python
def it_check(self, comm, num_procs):
    self.send_full(comm=comm, level=0)
    if self.S.status.force_done:
        return None          # three of these, all BEFORE the convergence controllers run
```

Once the estimator sets `force_done` on a rank, that rank short-circuits out of `it_check` on every later entry and never runs the estimator again — so it stops forwarding its running maximum and its successor's receive strands. The estimator's own verdict severs the channel its successors depend on. Note `S.status.prev_done` cannot be used to guard the receive either: it is updated later in the same `it_check`, so at estimator time it is one iteration stale.

A working version therefore needs either a receive that tolerates a vanished predecessor without consulting `prev_done`, or the forward send hoisted above the `force_done` guards — plus a stress test across drop-out orders. That is its own PR, and only worth doing if the estimator is actually wanted under MPI. Nothing in this repository has ever used it.

## Status after closing

Unaffected and already merged:

- #678 removed the inline MPI implementation, which was dead and deadlocked when switched on, and made `use_iteration_estimator` raise in both controllers instead of being silently ignored in one. That fix matters *more* now, not less: having decided not to support the estimator under MPI, we are not leaving a broken implementation in the tree pretending otherwise.
- #677's conformance suite is what caught this. It surfaced the regression as behavioural failures (8 iterations against 6, then 7 against 6) that no unit test on the estimator would have seen.

The estimator remains available with the whole block in one process, via `CheckIterationEstimatorNonMPI`, as tutorial step_8 C uses it. The gap is documented rather than silent.

PR #687, opened 2026-09-12
Deferred-correction ("delta-form") SDC sweepers

pancetta:
A standard SDC sweep

```
u^{k+1}_m = u_0 + tau_m + dt (Q f^k)_m + dt sum_j QI_mj (f^{k+1}_j - f^k_j)
```

is algebraically identical to, writing `delta_m = u^{k+1}_m - u^k_m` and the collocation residual
`eps_m = u_0 + tau_m + dt (Q f^k)_m - u^k_m`,

```
delta_m = eps_m + dt sum_j QI_mj [f(u^k_j + delta_j) - f(u^k_j)],    u^{k+1}_m = u^k_m + delta_m
```

Written this way every sweep is **iterative refinement**: a high-precision residual, a correction
solve, a high-precision update. No Jacobian appears, so an IMEX splitting survives unchanged, and
nothing outside the sweeper changes — no core and no controller modifications.

## Why

The quantity handed to the node-local solver is a *correction*. Its magnitude tends to zero as the
sweeps converge, so a solver working at precision `eps` introduces an error of order `eps*|delta|`
rather than `eps*|u|`, and therefore does not cap the attainable accuracy. Reducing the precision of
a solve that returns the state stalls the iteration at that precision; reducing one that returns a
correction does not.

This PR is the sweeper only. The reformulated *hierarchy* that lets a whole coarse level drop below
backend precision, and the measurements, follow separately.

## What is here

`delta_implicit` and `delta_imex_1st_order` are drop-in replacements for `generic_implicit` and
`imex_1st_order`; `delta_implicit_MPI` for `generic_implicit_MPI`.

Three routes to the correction equation, selected automatically:

| route | when | cost |
|---|---|---|
| `solve_system_delta` | the problem provides it | the only route for a nonlinear implicit operator |
| `linear_implicit=True` | linear or affine operator | none — stock `solve_system`, once the affine part is removed |
| substitution fallback | otherwise | always correct, but the solver sees an `O(1)` unknown |

`correction_precision` stores `eps`, `delta` and `Df` in a reduced-precision datatype, scaled by the
residual's own magnitude — block floating point, so the format's mantissa is used however small the
correction gets. Without the scaling `float16` stalls at 5.9e-05, which is its smallest normal.

Problem classes gain two **optional** methods, documented on `Problem` and implemented here for the
finite-difference and FEniCS heat equations:

- `eval_f_increment(base, delta, t)` — `f(base+delta) - f(base)` expanded analytically. Without it
  the sweeper subtracts two stored right-hand sides, a cancellation carrying the operator norm.
- `solve_system_delta(rhs, factor, base, f_base, t)` — for a nonlinear implicit operator.

Neither is required. Both fallbacks are correct and neither fails loudly, which is why they are
documented on `Problem` rather than left to be discovered.

## Tests

31 in `test_delta_form.py`, mostly that the rewrite changes no answer: against `generic_implicit`
for `IE`/`LU`/`MIN-SR-S` and for a zero-diagonal (Picard) preconditioner, against `imex_1st_order`,
and under MLSDC and PFASST on a stock `BaseTransfer` hierarchy. Plus the three correction-equation
routes, and the correction storage and its scaling.

4 in `test_delta_form_MPI.py`: the node-parallel sweep matches the serial one. The reduced-precision
scale has to be reduced across the node communicator — each rank holds one node and so sees only
part of the residual the divisor comes from. The `float32` parametrisation is what says so: remove
the `allreduce` and it fails, while the full-precision case correctly does not notice.

Verified on numpy 1.26.4 and 2.4.6. The diff is additive apart from one `"""` → `r"""`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pySDC/core/level.py
+++ b/pySDC/core/level.py
@@ -56,6 +56,7 @@ class Level(FrozenClass):
         f (list of dtype_f): RHS values at the nodes
         fold (list of dtype_f): copy of RHS values for saving data during restriction
         tau (list of dtype_u): FAS correction, allocated via step class if necessary
+        u0_reference: copy of ``u[0]`` taken when a residual was handed down to this level
     """
 
     def __init__(
@@ -102,6 +103,17 @@ class Level(FrozenClass):
 
         self.tau: List[Optional[Any]] = [None] * self.sweep.coll.num_nodes
 
+        # Set by a transfer that hands this level a residual instead of letting it rebuild one, and
+        # read by the sweeper that carries that residual. PFASST overwrites `u[0]` with the
+        # predecessor's end value *after* the restriction, and the residual depends on `u[0]`
+        # additively, so a level carrying one has to be told what `u[0]` was when it arrived. It
+        # lives here, next to `tau`, because it is a copy of this level's own data and has to be
+        # dropped when the level is reset -- a stale one belongs to the previous step.
+        #
+        # Temporary. Putting the step-to-step exchange itself in delta form, so the predecessor
+        # sends an increment rather than a state, removes the need for it entirely.
+        self.u0_reference: Optional[Any] = None
+
         self.__tag: Optional[Any] = None
 
         # freeze class, no further attributes allowed from this point
@@ -129,6 +141,7 @@ class Level(FrozenClass):
         self.f = [None] * (self.sweep.coll.num_nodes + 1)
         self.fold = [None] * (self.sweep.coll.num_nodes + 1)
         self.tau = [None] * self.sweep.coll.num_nodes
+        self.u0_reference = None
 
     @property
     def sweep(self) -> Sweeper:
--- a/pySDC/core/problem.py
+++ b/pySDC/core/problem.py
@@ -41,7 +41,7 @@ class WorkCounter(object):
 
 
 class Problem(RegisterParams):
-    """
+    r"""
     Prototype class for problems, just defines the attributes essential to get started.
 
     Parameters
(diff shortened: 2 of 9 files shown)


pancetta:
Pushed `9a98f16c3`, acting on the review point that a sweeper should not know about multi-level things. It was right on both counts, and the reason I had for it was wrong.

**There is no `_ml` sweeper any more.** There used to be a second class, `delta_implicit_ml`, selected when the run had more than one level. It is **bit-identical** to `delta_implicit` on a single level (diff exactly `0.000e+00` over 3 steps × 5 sweeps) because every one of its overrides falls through to `super()` when no residual was handed down. So the fork bought nothing and cost two things: a suffix nobody can read, and a sweeper you have to pick by level count — which stock pySDC never asks of you, since `generic_implicit` runs on one level or four.

The hierarchy support folds into `DeltaFormMixin`. `update_nodes` becomes the template — sync, sweep, advance, accumulate — and each concrete sweeper supplies `_sweep_nodes`. **What a level does is decided by the transfer**, which is where that choice already lives in pySDC: `BaseTransfer` hands nothing down and every level rebuilds its own residual, so on a stock hierarchy this reproduces MLSDC and PFASST exactly.

**`u0_reference` moves to `Level`.** My note said `Level` being a `FrozenClass` forced per-level state onto the sweeper. That is false — `FrozenClass` is a registry with `add_attr`, and `ConvergenceController.add_status_variable_to_level` already uses it. `u0_reference` is a copy of the level's own `u[0]`, so it belongs next to `tau`. It had already leaked: tracing a 3-step run, it survived `reset_level()` holding the previous step's value on a live object.

It is marked temporary in the source. It exists only because PFASST overwrites `u[0]` with the predecessor's end value *after* the restriction, and the inherited residual depends on `u₀` additively. Putting the step-to-step exchange itself in delta form removes the need for it. Measured, 4 steps of two-level heat: `u[0]` moves after restriction 12/16 times, by up to 6.2e-01, and ignoring it costs 4.0e-06 — while serial MLSDC never moves it at all.

`eps_in` and `delta_acc` stay on the sweeper: they are the sweep's working state, not a copy of the level's data.

**New test.** `test_a_handed_down_residual_is_used_and_advanced` stubs the hand-down, so the sweeper's half of the contract is not dormant-and-untested until the transfer lands: it must use `eps_in` in place of the residual it would have computed, advance it to still agree with a residual rebuilt from scratch, and bank the corrections.

Two things the fold broke and the tests caught, worth recording: the rounding transfer still wrote the reference to the sweeper, so the level kept the pre-rounding value and the sync injected a shift of exactly the rounding error (3.4e-11 at `float32`); and inlining the correction recorder into `_solve_correction` missed its two early returns, so a problem with `solve_system_delta` recorded nothing.

3352 base tests pass, and the delta-form files on numpy 1.26.4 and 2.4.6.

